PR #1, opened 2022-01-23
Detector

rodja:
Introduction of Detector (from learning-loop.ai).


rosys/actors/detector.py
```diff
@@ -0,0 +1,82 @@
...
+            self.is_connected = True
+        except:
+            self.log.exception('connection failed; trying again')
+            self.is_connected = None
+            await self.sleep(reconnect_delay)
+
+    async def step(self):
+        if self.is_connected is None:
+            await self.connect()
+
```
falkoschindler, line 43: What if `self.connected` is false? There used to be an early exit here.

rosys/actors/detector.py
```diff
@@ -0,0 +1,82 @@
...
+        if self.world.upload_queue:
+            task_logger.create_task(self.upload(self.world.upload_queue.pop(0)), name='upload_frame')
```
falkoschindler, line 45: There used to be a throttling using `minimal_minutes_between_uploads`. Is it not needed anymore?

PR #2, opened 2022-01-26
Calibration

falkoschindler:


rosys/world/image.py
```diff
@@ -0,0 +1,32 @@
...
+    height: int
+
+
+@dataclass
+class Image():
+    camera_id: str
+    size: ImageSize
+    time: float  # World time of recording
+    data: Optional[bytes] = field(default=None, init=False)
+    detections: Optional[list[Detection]] = field(default=None, init=False)
```
falkoschindler, line 21: Why do we need `field`?

falkoschindler: Needed for inheritance. Already obsolete.

PR #4, opened 2022-02-06
Lizard stats

rodja:
Adds lizard statistics (responsiveness, update and processing duration)

### DIFF
--- a/rosys/actors/lizard.py
+++ b/rosys/actors/lizard.py
@@ -13,15 +13,21 @@ class Lizard(Actor):
         self.last_step = None
         event.register(event.Id.PAUSE_AUTOMATIONS, self._handle_pause)
         self.responsiveness_stats: list = []
+        self.update_stats: list = []
+        self.processing_stats: list = []
 
     async def step(self):
+        t = self.world.time
         await self.ensure_responsiveness()
         await self.hardware.update()
+        self.update_stats.append(self.world.time - t)
         t = self.world.time
         await event.call(event.Id.NEW_MACHINE_DATA)
         dt = self.world.time - t
         if dt > 0.02:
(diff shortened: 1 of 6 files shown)


rosys/actors/lizard.py
```diff
@@ -11,15 +12,22 @@ def __init__(self, hardware: Hardware):
...
         event.register(event.Id.PAUSE_AUTOMATIONS, self._handle_pause)
+        self.responsiveness_stats: list = []
+        self.update_stats: list = []
+        self.processing_stats: list = []
 
     async def step(self):
+        t = self.world.time
         await self.ensure_responsiveness()
         await self.hardware.update()
+        self.update_stats.append(self.world.time - t)
```
falkoschindler, line 23: possibly measures `await self.pause_automations(...)`...

rosys/actors/lizard.py
```diff
@@ -33,3 +41,12 @@ async def ensure_responsiveness(self):
         elif dt > 0.1:
             self.log.warning(f'esp serial communication is slow ({dt:.2f} s since last step)')
         self.last_step = self.world.time
+        self.responsiveness_stats.append(dt * 100)
+
+    def limit_stat_values(self):
```
falkoschindler, line 46: can be avoided using [`collections.deque`](https://docs.python.org/2/library/collections.html#collections.deque)

rosys/actors/lizard.py
```diff
@@ -33,3 +41,12 @@ async def ensure_responsiveness(self):
         elif dt > 0.1:
             self.log.warning(f'esp serial communication is slow ({dt:.2f} s since last step)')
         self.last_step = self.world.time
+        self.responsiveness_stats.append(dt * 100)
```
falkoschindler, line 44: Why `* 100`?

PR #8, opened 2022-03-08
Detector improvements

falkoschindler:


falkoschindler:
Tests are red due to an existing problem with the pytest infrastructure.

PR #10, opened 2022-03-25
backup full robot (except transient infos like velocity)

rodja:


rosys/persistence.py
```diff
@@ -9,15 +10,18 @@ class Persistence:
     def __init__(self, world: World, filepath: str = '~/.rosys/world.json'):
         self.world = world
         self.filepath = os.path.expanduser(filepath)
+        os.makedirs(os.path.dirname(self.filepath), exist_ok=True)
 
     def dump(self) -> dict:
         return {
+            'robot': self.world.robot.dict(exclude={'velocity', 'position', 'shape', 'parameters', 'prediction', 'detection', 'simulation', 'odometry', 'current_velocity', 'last_movement', 'hardware_time', 'clock_offset', 'carrot'}),
```
falkoschindler, line 17: - Why excluding the `shape`?
- What is `position`?

rosys/persistence.py
```diff
@@ -28,7 +32,7 @@ def backup(self):
             json.dump(self.dump(), f)
 
     def parse_world(self, obj: dict):
-        return World.parse_obj(obj)
+        return self.world.parse_obj(obj)  # NOTE we use the provided world for parsing to load all attributes
```
falkoschindler, line 35: I guess the purpose of this change is that `parse_world` will return whatever type `self.world` is.
As far as I understand the instance of `self.world` is omitted, because `parse_obj` is a class method.
`parse_world` is missing a return type. Since it should return "what `self.world` is", we can use a [TypeVar](https://docs.python.org/3/library/typing.html#typing.TypeVar): Introduce `W = TypeVar('W')` and use it as `self.world: W` and `def parse_world(...) -> W:`.

PR #15, opened 2022-09-07
Allow profiling inside rosys.vision and rosys.analyze packages

rodja:
moved profiling to rosys.debugging.


falkoschindler:
Why do we need a new debugging package? Why can't we use profile from analysis?

PR #23, opened 2023-09-22
made step_size of Carrot.move() independent from the Spline length

AndyPermaRobotics:
this optimizes the behavior of move() for very long splines and avoids that the Driver aborts at 99% because of the former stepsize of 1%


falkoschindler (approved):
Thanks a lot for this pull request, @AndyPermaRobotics!
I just did some minor changes and added a pytest to cover the case of driving long splines. It indeed succeeds when using the new step size. 👍

PR #25, opened 2023-09-29
Unify camera providers

falkoschindler:
This PR tries to unify different camera provider classes:

- [x] The `active` flag has a different meaning. We will rename it to `is_connected` and let it represent the device state, not whether the user wants it to be active.
- [x] Maybe we'll introduce `enable()`/`disable()` methods and an `is_enabled` property to allow the user to enable/disable the capturing process.
- [x] `MultiCameraProvider` and `CombiningCameraProvider` are doing the same. We can remove the latter.
- [x] The `CAMERA_ADDED` event should be emitted more consistently.


falkoschindler:
We came up with the following class hierarchy for structuring cameras and camera providers in our projects:

RoSys
-----

CameraProvider
UsbCameraProvider
RtspCameraProvider
SimulatedCameraProvider

Camera
UsbCamera -> UsbDevice
RtspCamera -> RtspDevice
SimulatedCamera

CalibratedCamera [mixin]


Project 1
---------

FolderCameraProvider
FolderCamera

GalaxyCameraProvider

CameraScanner
IpCamera


Project 2
---------

WifiCamera(CalibratedCamera)
WifiCameraProvider
WifiCameraProviderHardware
WifiCameraProviderSimulation
WifiCam

NiklasNeugebauer:
Upon further review and if I understand asyncio correctly, the locking mechanism will not be necessary. In fact, the old problem should also be fixable without a lock by moving the check for the existence of the device inside `set_parameters`.

PR #44, opened 2023-10-16
added lock to prevent crash on parameter update

angelom93:
Quick overview:

Problem: We were experiencing crashes when `get_images` was executed from within `set_parameters` on Jetson Orin Nano.

Solution: I've added thread synchronization to prevent concurrent execution of these functions, eliminating the crashes.

Testing: I've tested the changes on a Jetson Orin Nano, and it's running smoothly without crashes.

### DIFF
--- a/rosys/vision/usb_camera_provider_hardware.py
+++ b/rosys/vision/usb_camera_provider_hardware.py
@@ -1,3 +1,4 @@
+import asyncio
 import io
 import logging
 import re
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Looks good to me! 👍🏻
I just requested a review from @NiklasNeugebauer, since he's also working on this module.

NiklasNeugebauer:
Thanks for adding me!
I have looked through it, but could not quite figure out the root of the issue.
What is interfering here?

Can you share the specific error you are getting?

NiklasNeugebauer:
After a discussion, the issue is the potential disconnection of cameras in `capture_images` while looping over devices in `set_parameters`. The lock makes sense and will be moved into the camera classes in the 'active_cameras' branch

PR #46, opened 2023-10-19
Use `rosys.on_startup()` instead of `app.on_startup()`

falkoschindler:
This PR follows up on discussion #45 and propagates the consequent use of RoSys' own lifecycle registration functions.

I had to update the mkdocs config a little bit. We'll have to see if the doc generation and deployment still works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Modules can depend on other modules which is mostly implemented by passing them 
 
 ### Lifecycle Hooks and Loops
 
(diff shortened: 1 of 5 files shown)


rodja (approved):
Great! What do you think, @AndyPermaRobotics?

AndyPermaRobotics:
> Great! What do you think, @AndyPermaRobotics?

Looks great!
I appreciate that @falkoschindler also added the 'why' to the documentation.

PR #52, opened 2023-11-19
allow driving backwards

rodja:
This PR introduces the `backward` bool property for the `drive_to` method.


falkoschindler (changes requested):
I think this PR ignores the `minimum_turning_radius`. If the minimum radius is set, `drive_to` starts with driving on a circle before continuing on a spline. This circular part doesn't respect the `backward` parameter yet.

falkoschindler (approved):
There was still a 180-degree error when driving backwards. Now it should work as expected. I tested with a slightly modified "obstacles" example.

PR #53, opened 2023-11-19
Introduce odometer.reset

rodja:
I noticed that there are quite some things to be aware of if you want to reset the odometer to the initial position. This PR provides a handy reset method to take care of the details.


falkoschindler (approved):
Can you name an application for such a reset method?
Apart from that the code looks good. 👍

rodja:
> Can you name an application for such a reset method?

Sure: for the Agritechnica we demonstrated a our [Field Friend robot](https://feldfreund.de) going forward and weeding a line of crops. If the line ends it goes back to 0,0 and resets the odometer to ensure we can start the whole process anew.

PR #58, opened 2023-12-07
Enhanced image upload

denniswittich:
This PR allows to provide detections to the upload function of the 'Hardware Detector'. The detections are modified to match the format of the detector node. With this PR it is also possible to  provide a list of tags which are forwarded to the detector node. The default behavior is not changed.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -30,111 +30,99 @@ files = [
 
 [[package]]
 name = "aiohttp"
-version = "3.8.6"
+version = "3.9.1"
 description = "Async http client/server framework (asyncio)"
 optional = false
-python-versions = ">=3.6"
+python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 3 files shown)


rosys/vision/detector_hardware.py
```diff
@@ -89,10 +89,24 @@ async def upload_priority_images():
             rosys.background_tasks.create(upload_priority_images(), name='upload_priority_images')
             self.uploads.priority_queue.clear()
 
-    async def upload(self, image: Image) -> None:
+    async def upload(self, image: Image, tags: list[str] = []) -> None:
```
falkoschindler, line 92: What do you think about making `tags` a keyword-only argument?

falkoschindler (approved):
Looks good to me. 👍
Now I understand the trouble of converting box coordinates to integers. But I think it's a reasonable solution for now.

rosys/vision/detector_hardware.py
denniswittich: I like that idea. Please check again

PR #61, opened 2023-12-11
improve error message for multiple detection types

rodja:
This PR includes the available types in the exception.


denniswittich (approved):
Approved. Note that the keys are UUIDs if no meaningful name is given to the detector.

PR #67, opened 2024-01-08
use echarts instead of highcharts

codingpaula:
Use echarts instead of highcharts on the KPI page since echarts comes with nicegui and doesn't require an additional package to be installed.


falkoschindler (approved):
Very nice! 👌

PR #71, opened 2024-01-12
Module to interface with http MJPEG camera streams

NiklasNeugebauer:
This branch implements `MjpegCamera`, `MjpegDevice` and `MjpegCameraProvider`.
In particular, it is meant to interface with network cameras over http and read the provided MJPEG stream.

Some open questions:

### Gstreamer
Would using gstreamer as with the rtsp camera improve performance?

### How should we differentiate different camera modules?
This module is similar to the `RtspCamera` as cameras are also discovered over arp-scan.
A shared module could use a mechanism to detect and select the appropriate protocol at runtime or let the user decide.

### Naming
Since USB cameras can also provide an MJPEG stream, would `HttpCamera` be more appropiate?
I think that would be more clear from a user perspective.

### Other protocols
The camera I am testing this module on also supports AVC encoding. Should this be supported by this module, too?


rodja:
@NiklasNeugebauer I'll answer the questions you stated in the description below (Btw I think these kinds of questions are not well suited to be placed in the description).

> Would using gstreamer as with the rtsp camera improve performance?

I don't think so. Maybe if you also need to transform the image (size, crop) etc. Otherwise the raw jpeg stream is directly cut into jpeg data blobs.

> How should we differentiate different camera modules? This module is similar to the RtspCamera as cameras are also discovered over arp-scan. A shared module could use a mechanism to detect and select the appropriate protocol at runtime or let the user decide.

I think they work quite differently (one using gstreamer to create jpegs and this one only needing to cut the jpegs from a stream). Maybe extracting the "device discovery via arp-scan" could be extracted into a separate module. But I would leave that to another pull request.

> Since USB cameras can also provide an MJPEG stream, would HttpCamera be more appropiate? I think that would be more clear from a user perspective.

But there could be other http cameras like our Wifi Cam (which has a totally different interface and purpose). Maybe `HttpMjpegCamera`?

> The camera I am testing this module on also supports AVC encoding. Should this be supported by this module, too?

I would rather keep it with mjpeg. Would AVC not be better handled with the RtspCamera?

rosys/driving/keyboard_control_.py
```diff
@@ -33,7 +33,7 @@ def handle_keys(self, e: KeyEventArguments) -> None:
             self.steerer.update(self.direction.y, self.direction.x)
 
         # update individual speed components via arrow keys
-        if e.modifiers.shift and e.key.is_cursorkey:
```
rodja, line 36: Why removing the SHIFT modifier?

NiklasNeugebauer: This is from a merge from main. I made have made a mistake when merging?

rodja: It's not on main. It may have been accidentally introduced by myself when testing a demo. But `e.modifiers.shift` is wrong. I'll fix that on main.

rodja (approved):
Looks good so far. But I've not had the chance to test it yet. And since it's still draft I have not removed print statements etc. I just did some formatting and in some cases shorter expressions.

rodja (approved):

falkoschindler (approved):

PR #74, opened 2024-01-18
add ID_PATH and DEVNAME as fallbacks for the camera id

### LINKED ISSUE #69: udev string ID_SERIAL_SHORT might not be available
The new USB Camera Module uses udev strings to assign a unique id to each camera.
For that it currently combines `ID_VENDOR_ID`, `ID_MODEL_ID` and `ID_SERIAL_SHORT`.

We have seen one instance of a camera not providing `ID_SERIAL_SHORT`.

Note: `ID_SERIAL` does not come directly from the USB device descriptor, but is constructed from multiple strings like we do. It might be a better fit if it is always available.

NiklasNeugebauer:
This resolves #69

### DIFF
--- a/rosys/vision/usb_camera/usb_camera_scanner.py
+++ b/rosys/vision/usb_camera/usb_camera_scanner.py
@@ -9,7 +9,7 @@ def uid_from_device(device: pyudev.Device) -> Optional[str]:
         device.get('ID_MODEL_ID'),
         device.get('ID_SERIAL_SHORT') or device.get('ID_PATH') or device.get('DEVNAME'),
     ]
-    return '-'.join(parts) if all(parts) else None
+    return '-'.join(parts) if all(parts) else device.get('DEVNAME') or None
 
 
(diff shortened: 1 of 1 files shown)


falkoschindler:
I wonder if we should always use the `ID_VENDOR_ID` and `ID_MODEL_ID`, like
```py
parts = [
    device.get('ID_VENDOR_ID'),
    device.get('ID_MODEL_ID'),
    device.get('ID_SERIAL_SHORT') or device.get('ID_PATH') or device.get('DEVNAME'),
]
return '-'.join(parts) if all(parts) or None
```

🤔

NiklasNeugebauer:
Sure, I pushed that. We might run into the case of these also not being available while jsut the devpath realistically must be, but lets see if that ever happens

rodja (approved):

NiklasNeugebauer:
Actually, I added the final "only devname" case

PR #77, opened 2024-01-29
fixed estop corehandling

Johannes-Thiel:
`active = any(int(words.pop(0)) == 0 for _ in self.pins)`
if words lookes like this [0,1] this only pops the 0 and the rest is not popped. Words will look like this [1] but should be [].
This pops all estop vales first and than uses any.


falkoschindler (approved):
Oh wow! First I thought the code before and after the change is basically identical. But the lazy evaluation of `any` stops popping items at the first "0". Creating a list before calling `any` should indeed fix this bug. Thanks!

PR #81, opened 2024-02-09
Integration of IMU using quaternions

Johannes-Thiel:
Integration of IMU using eulerangles, and offsetvalues. This includes a Hardware class and a simulation to test the Emit messages.
The consturctor of the Imu expects offset vales, to be able to work with offset IMU.


Johannes-Thiel:
@falkoschindler Kannst du einmal drüber schauen?

falkoschindler (commented):
We just did a code review together and improved a few things here and there.
Before we merge, @Johannes-Thiel will check again if we multiply the measured rotation and the rotation offset in the right order.

Johannes-Thiel:
Due to the placement of the Imu in the robot, we have the Problem of gimbal lock, when using the Euler angles provided by the unit. A fix for that problem is the use of quaternions, also provided by the Imu. This implementation serves the same funktion as before, but using quaternions.

falkoschindler (approved):

PR #88, opened 2024-02-22
Mjpeg multiple cameras

NiklasNeugebauer:
Adds support for multiple cameras connected through a single ip.
In particular, ids now have an index after the mac.
For axis devices (the only ones supported right now), we ask for the the video status at 'axis-cgi/videostatus.cgi' during camera discovery and create cameras based on that rather then a single one.

The option to supply an ip to Camera.connect is also added to avoid triggering a slow arp-scan for the same ip multiple times.


falkoschindler (approved):
🎉

PR #102, opened 2024-04-29
Add Motec as a mjpeg-vendor and refactor camera providers

pascalzauberzeug:
This PR adds support for mjpeg-based [Motec ip cameras](https://www.motec-cameras.com/)

The changes are:
- add Motec as a mjpeg vendor
- change VendorType to ~~IntEnum~~ Enum for both mjpeg and rtsp
- remove check that disconnects cameras which are not in the arp-scan but are still connected
- clean up find_cameras return values
- refactor mjpeg_device.py

~~IntEnum can be changed to Enum, but may lead to errors in other projects~~

### DIFF
--- a/rosys/vision/rtsp_camera/arp_scan.py
+++ b/rosys/vision/rtsp_camera/arp_scan.py
@@ -34,7 +34,7 @@ async def find_cameras() -> AsyncIterator[tuple[str, str]]:
             continue
         mac = infos[1]
         ip = infos[0]
-        if mac in ['packets', 'arp-scan'] or ip == "Interface:":
+        if mac in ['packets', 'arp-scan'] or ip == 'Interface:':
             continue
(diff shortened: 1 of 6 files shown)


rosys/vision/rtsp_camera/arp_scan.py
NiklasNeugebauer, line 1: This was meant to ignore the lines that are do not contain ip, max pairs

NiklasNeugebauer (changes requested):
add back arp-scan output parsing

rosys/vision/rtsp_camera/arp_scan.py
NiklasNeugebauer: oh, nevermind, you added that. Was this not in the code somewhere before

NiklasNeugebauer (approved):

PR #103, opened 2024-04-30
Image Encoding

NiklasNeugebauer:
This pull request removes unnecessary en- and decoding in the camera pipeline.


pascalzauberzeug:
The rotated image had the wrong color space because cv2 uses BGR. I fixed that and reused process_ndarray_image for the function. I tested the code with a mjpeg camera

pascalzauberzeug (approved):

PR #109, opened 2024-05-07
filter network interfaces for arp scan

pascalzauberzeug:
If no network interface is provided, arp-scan will use the first one that is up.

> The network interface to use can be specified with the --interface option. If this option is not present, arp-scan will search the system interface list for the lowest numbered, configured up interface (excluding loopback).

I had the problem where arp-scan tried to use a can-bus interface, so I wrote this function to filter out unwanted interfaces.
Which are loopback, docker interfaces, virtual interfaces and can.

Two possible issues here:
- [The netifaces module is no longer maintained](https://github.com/al45tair/netifaces/issues/78), but is already in the rosys dependencies, so I used it
- It still just accepts the first correct interface. An interface parameter would be nice, but would mean a long parameter passthrough chain


falkoschindler (approved):
The PR looks good to me. Just a few thoughts:

- Should we sort the interfaces before filtering them to get predictable results? Or how is the order of `netifaces.interfaces()` defined?
- `netifaces` not being maintained since 2021 could sooner or later become a problem. @rodja Do you know why RoSys has this dependency even though it isn't used other than in this PR? And is there an alternative to switch to? Or do we need to call `netifaces` on the command line and parse the result. I'd very much prefer a Python library...
- Regarding your suggestion of an interface parameter, I'm not sure about the costs and benefits. If we don't need it at the moment, let's not add it. If we do need it, let's create a new branch/PR to find and discuss a possible implementation.

rodja:
I think netifaces is just a relict. NiceGUI already uses ifaddr: https://github.com/zauberzeug/nicegui/blob/cf61dcb2191d55fbf38fb8d38fc7867a4801c07c/nicegui/welcome.py#L11
@pascalzauberzeug could you try to switch to this dependency?

pascalzauberzeug:
@falkoschindler regarding the sorting of the interfaces, ifaddr uses the order the operating system provides. I would keep it that way to have to same order as commands like `ifconfig`

falkoschindler (approved):
Nice! ifaddr seems to work very similar to netifaces. Great to have this dependency removed.
I think we can merge into main. 🙂

PR #110, opened 2024-05-10
Provide setter for default automation

rodja:
This pull request adds `set_default_automation` to the `automator` which is handy for usage in lambda statements.


falkoschindler (approved):
@rodja Should we make the parameter `default_automation` optional, i.e. accept `Callable | None`? This way one could remove the default automation using this setter.

rodja:
@falkoschindler Sure! I just pushed the change.

PR #114, opened 2024-05-16
Arp scan specific interface

NiklasNeugebauer:
Adds ability to specify a specific network interface to use when scanning for cameras.

Branches from #109


rosys/vision/rtsp_camera/arp_scan.py
pascalzauberzeug, line 1: We should decide whether to use the term 'interface' or 'adapter'.
Arp-scan and the old library netifaces use 'interface' and ifaddr uses 'adapter'.
Personally, I would go with 'interface'

rodja: I also prefere 'interface'. Its more in line with unix nomeclatur.

pascalzauberzeug: I renamed it

pascalzauberzeug (approved):
I resolved the conflicts with main, looks good to me

PR #115, opened 2024-05-16
Motec settings api

NiklasNeugebauer:
This adds the class `MotecSettingsInterface` used in `MjpegDevice`, which is similar to `JovisionRtspInterface`.
Since this camera tends to block on requests, I made it async which requires most functions in ConfigurableCamera to be async, too.

This is changing the API since the public `set_parameters` function has to be awaited now.


NiklasNeugebauer:
MJpeg Camera related tests are failing, they should be skipped without a camera...

NiklasNeugebauer:
I have solved the race condition of a disconnect happening at the await inside a getter or setter by adding a try/except block and checking if the camera is disconnected. If so, `_apply_parameters` is aborted.

falkoschindler (commented):
Looks good to me. Only two remaining questions:

- Is `test_rtsp_camera` correct? There has been a merge conflict and I'm not sure if I resolved it correctly.
- Do we need to change the return type of setters from `None` to `Any`? I don't think `Any` is necessary just because they can be async now.

NiklasNeugebauer:
die tests in `test_mjpeg_camera.py` kommen sich immer noch in die quere, aber sie funktionieren einzeln vernünftig. Ich denke, das könnte gemerged werden

falkoschindler (approved):

falkoschindler (approved):

PR #116, opened 2024-05-17
Emitting events should call non-async functions synchronously

rodja:
This pull request ensures that non-async functions are called immediately. The usage of `invoke` was wrong because it's an async function itself and hence will always be awaitable.


falkoschindler (approved):

rodja:
@falkoschindler I think it is better to pass the cause into the `on_interrupt` callback.

falkoschindler:
@rodja I needed to revert 2aa4749af71d85e28231c0be5f1116e3cb5ac635 in d77abfa27c215bc5a57e015a388de2c38edcffd4 because it requires the event handlers to take another parameter, which breaks existing code. If we want to provide an (optional) "cause", we need to think about it once again separately.

PR #117, opened 2024-05-17
More static cameras

NiklasNeugebauer:
Two changes that I felt were hard to find a good name for:

- ip is now stored in both Rtsp and Mjpeg Network cameras such that they can quickly reconnect on startup and are less dependent on network scans
- username and passwords are not overwritten by the `MJpegCameraProvider` when loading cameras from persistence

An app should now be able to set up and configure cameras and have everything be loaded and setup as-is on the next start.

### DIFF
--- a/rosys/vision/mjpeg_camera/mjpeg_camera.py
+++ b/rosys/vision/mjpeg_camera/mjpeg_camera.py
@@ -28,7 +28,7 @@ class MjpegCamera(TransformableCamera):
                  ) -> None:
         super().__init__(id=id, name=name, connect_after_init=connect_after_init, streaming=streaming,
                          image_grab_interval=image_grab_interval, base_path_overwrite=base_path_overwrite, **kwargs)
-        self.logger = logging.getLogger(f'rosys.vision.mjpeg_camera.{self.id}')
+        self.log = logging.getLogger(f'rosys.vision.mjpeg_camera.{self.id}')
         self.username = username
         self.password = password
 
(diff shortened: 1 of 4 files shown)


rosys/vision/mjpeg_camera/mjpeg_camera.py
```diff
@@ -23,13 +23,17 @@ def __init__(self,
                  base_path_overwrite: str | None = None,
                  username: str | None = None,
                  password: str | None = None,
+                 ip: str | None = None,
                  **kwargs: Any,
                  ) -> None:
         super().__init__(id=id, name=name, connect_after_init=connect_after_init, streaming=streaming,
                          image_grab_interval=image_grab_interval, base_path_overwrite=base_path_overwrite, **kwargs)
+        self.logger = logging.getLogger(f'rosys.vision.mjpeg_camera.{self.id}')
```
pascalzauberzeug, line 31:
```suggestion
        self.log = logging.getLogger(f'rosys.vision.mjpeg_camera.{self.id}')
```

rosys/vision/rtsp_camera/rtsp_camera.py
```diff
@@ -30,8 +30,11 @@ def __init__(self,
                          streaming=streaming,
                          **kwargs)
 
+        self.logger = logging.getLogger(f'rosys.vision.rtsp_camera.{self.id}')
```
pascalzauberzeug, line 33:
```suggestion
        self.log = logging.getLogger(f'rosys.vision.rtsp_camera.{self.id}')
```

pascalzauberzeug (approved):
Looks good to me, except for the two naming suggestions

falkoschindler (approved):

PR #118, opened 2024-05-19
Add driver.velocity property

rodja:
This pull request adds a velocity property to the Driver class.


falkoschindler (commented):
We agreed to drop this PR because the velocity property isn't used by the driver.

PR #119, opened 2024-05-19
Make asyncio warnings ui container accessible and hide controls

rodja:
This pull request improves the `asyncio_warnings.ui()`.


falkoschindler (commented):
@rodja What is "hide-controls" supposed to do? I can't find it in the Quasar documentation and in a simple NiceGUI test it doesn't seem to make a difference:
```py
ui.number('coro limit', value=0.1, format='%.3f').props('suffix=s').classes('w-16')
ui.number('coro limit', value=0.1, format='%.3f').props('suffix=s hide-controls').classes('w-16')
```

rodja:
Oh, `hide-controls` should hide the spin buttons in the input field. But that does not work as you pointed out. I tried quite a few things but it seems it's not so easy. I'll just undo it.

falkoschindler (approved):

PR #122, opened 2024-05-22
Improve Mjpeg data transfer performance

NiklasNeugebauer:
This reduces the time taken to find and create an image in the incoming data stream from around 0.23 ms an average to 0.11 ms on average on my machine.
Most of that is from avoiding the call to remove_exif every time, which will of course still take ~0.1 ms each time `capture` is called.

I think using a fixed size `bytesarray` should at least remove some work from the garbage collector though.

The changes to `remove_exif` now also half its runtime on my machine (to around 50us)


rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -36,50 +35,71 @@ async def restart_capture(self) -> None:
...
-                                            pos = max(0, buffer.tell() - 1)
                                             break
-                                        image_data = buffer.getvalue()[header:footer_pos + 2]
-                                        yield remove_exif(image_data)
-                                        buffer = BytesIO(buffer.getvalue()[footer_pos + 2:])
-                                        pos = 0
+                                        image_data = buffer[header:footer_pos + 2]
+                                        yield image_data
+
+                                        buffer_view[:buffer_end - (footer_pos + 2)
```
falkoschindler, line 82: The line wraps a bit weirdly. What about this alternative:
```py
buffer_view[:buffer_end - (footer_pos + 2)] = \
    buffer_view[footer_pos + 2:buffer_end]
```
One way or the other, it's pretty hard to read. Should we introduce a helper variable like `shift = footer_pos + 2` to underline the purpose of this code block?
```py
shift = footer_pos + 2
buffer_view[:buffer_end - shift] = buffer_view[shift:buffer_end]
header = None
buffer_end -= shift
byte_search_pos = 0
```

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -36,50 +35,71 @@ async def restart_capture(self) -> None:
...
+                                        byte_search_pos = 0
                         except httpx.ReadTimeout:
                             self.log.warning('Connection to %s timed out', self.url)
-                except Exception:
-                    self.log.warning('Initial connection to %s failed. Was something disconnected?', self.url)
-
+                except Exception as e:
+                    self.log.warning('Connection to %s failed. Was something disconnected?\n%s', self.url, e)
+                    raise e
+        i = 0
```
falkoschindler, line 92: It looks like `i` is never used...

falkoschindler (commented):
Ok, as far as I can tell, the code looks good to me. There are just three remarks I added as comments.

NiklasNeugebauer:
@falkoschindler I incorporated your changes and improved the remove_exif function

falkoschindler (approved):

PR #124, opened 2024-05-25
Create Point3d from tuple and list.

rodja:
This PR adds two helpful builder functions to the `Point3d` class.

### DIFF
--- a/rosys/geometry/point3d.py
+++ b/rosys/geometry/point3d.py
@@ -12,6 +12,14 @@ class Point3d:
     y: float
     z: float
 
+    @staticmethod
+    def from_tuple(t: tuple[float, float, float]) -> Point3d:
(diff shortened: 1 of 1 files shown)


falkoschindler (commented):
What do you think about this alternative?

```py
@staticmethod
def from_sequence(seq: Sequence[float]) -> Point3d:
    return Point3d(x=seq[0], y=seq[1], z=seq[2])
```

I'm not too sure about the name `from_sequence`. But we could call it `from_tuple` (referring to the already existing `tuple` property) and accept a `Sequence[float]`, so that you can use it for tuples and lists.

rodja:
@falkoschindler I like the idea to consolidate the factory with `Sequence`. I've also added a docstring. Ready to merge?

falkoschindler (approved):

PR #126, opened 2024-05-29
Axis Settings (MjpegDevice Subclasses)

NiklasNeugebauer:
The goal here is primarily to add the ability to change the settings on axis camera devices.

This works differently from all previous interfaces since it does not change anything on the camera device itself, but rather just edits the url that is used to request images.
As a result that functionality really needs to be implemented inside the MjpegDevice and I have thus separated devices with different interfaces into subclasses of MjpegDevice.

Some logging was also added to the providers `scan_for_cameras` method.

Considerations I have for the future of this:

- The configurable parameters of a camera should really be determined by the device. We now have a discrepancy of parameters between vendors:
Axis devices have a `mirrored` parameter and  Motec devices have options for the `stream_port` and `compression` level.
(really, this was already the case for usb cameras, too, but we do not actively differentiate there). 
Additionally, the base class now has to provide mock functions for all parameters.
- Parameters should probably be adjustable even if the camera is not connected so that they are applied once it is. The downside is that we might not get feedback that settings a value failed.

### DIFF
--- a/rosys/vision/mjpeg_camera/mjpeg_camera_provider.py
+++ b/rosys/vision/mjpeg_camera/mjpeg_camera_provider.py
@@ -37,14 +37,17 @@ class MjpegCameraProvider(CameraProvider[MjpegCamera], persistence.PersistentMod
 
     def restore(self, data: dict[str, dict]) -> None:
         for camera_data in data.get('cameras', {}).values():
+            self.log.debug('restoring camera: %s', camera_data)
             camera = MjpegCamera.from_dict(camera_data)
             self.add_camera(camera)
         for camera in self._cameras.values():
             camera.NEW_IMAGE.register(self.NEW_IMAGE.emit)
 
     async def scan_for_cameras(self) -> list[tuple[str, str]]:
+        self.log.debug('scanning for cameras...')
         ids_ips: list[tuple[str, str]] = []
         async for mac, ip in find_cameras(self.network_interface):
+            self.log.debug('device found: %s %s', mac, ip)
             vendor = mac_to_vendor(mac)
             if vendor == VendorType.OTHER:
                 continue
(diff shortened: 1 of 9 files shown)


rosys/vision/mjpeg_camera/motec_mjpeg_device.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, mac: str, ip: str, *,
+                 username: Optional[str] = '',
+                 password: Optional[str] = '',
+                 control_port: Optional[int] = 8885) -> None:
+
+        super().__init__(mac, ip,
+                         username=username, password=password)
+
+        vendor = mac_to_vendor(mac)
+        if not vendor == VendorType.MOTEC:
```
pascalzauberzeug, line 18: I like it more this way, both is fine.
```suggestion
        if vendor != VendorType.MOTEC:
```

pascalzauberzeug (commented):
Looks good to me besides those minor changes

pascalzauberzeug (approved):

rosys/vision/mjpeg_camera/mjpeg_camera_provider.py
```diff
@@ -37,14 +37,17 @@ def __init__(self, *,
...
             camera = MjpegCamera.from_dict(camera_data)
             self.add_camera(camera)
         for camera in self._cameras.values():
             camera.NEW_IMAGE.register(self.NEW_IMAGE.emit)
 
     async def scan_for_cameras(self) -> list[tuple[str, str]]:
+        self.log.debug('scanning for cameras...')
         ids_ips: list[tuple[str, str]] = []
         async for mac, ip in find_cameras(self.network_interface):
+            self.log.debug('device found: %s %s', mac, ip)
```
rodja, line 50: @NiklasNeugebauer this is taking quite a lot of space in the logs (and happens quite frequently). Maybe it's better to improve this. I think we should not force every project which activates DEBUG for RoSys to fine-granularly deactivate it for the mjpeg camera provider.

PR #131, opened 2024-06-08
Simplify steering without fancy dead band regions and inversions.

rodja:
This PR reverts commit 400501c6ea20806a6dc1a2e0ee4b0d5dc82891f4 because it made driving backwards with fine adjustments to the direction very difficult.

### DIFF
--- a/rosys/driving/steerer.py
+++ b/rosys/driving/steerer.py
@@ -4,7 +4,6 @@ from enum import Enum
 import rosys
 from rosys.event import Event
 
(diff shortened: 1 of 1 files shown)


rodja:
Another problem which we might want to fix: When the max speed of the robot is reached by linear factor alone, steering is not longer possible. The angluar addition is clamped by the motors making all steering ineffective.

falkoschindler (approved):

PR #132, opened 2024-06-12
Fix camera streaming

NiklasNeugebauer:
- set streaming to True by default
- use a should_stream field for persistence (currently if the stream task ends or is not yet started, steaming=False could be written into persistence)
- show the "no image" placeholder if the camera is disconnected

### DIFF
--- a/rosys/vision/camera/camera.py
+++ b/rosys/vision/camera/camera.py
@@ -31,6 +31,8 @@ class Camera(abc.ABC):
         self.images: deque[Image] = deque(maxlen=self.MAX_IMAGES)
         self.base_path: str = f'images/{base_path_overwrite or id}'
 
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Looks correct. 👍🏻

rosys/vision/camera/camera.py
```diff
@@ -74,7 +77,7 @@ def get_image_url(self, image: Image) -> str:
 
     def get_latest_image_url(self) -> str:
         image = self.latest_captured_image
-        if image is None:
+        if image is None or not self.is_connected:
```
falkoschindler, line 80: How is this change related to this PR?

NiklasNeugebauer: it is not

PR #137, opened 2024-06-20
Fix RTSP Process Duplication

NiklasNeugebauer:
We noticed that when restarting the gstreamer process after setting a parameter, the old one might not be cleaned up properly even when the device is disconnected.

- check if there is already an active capture_task in `_start_gsteamer_task`
- check if there is already an active `capture_process` in `_run_gstreamer`
- make `shutdown` async
- only return from `shutdown` once `capture_process` and `capture_task` have finished
- add and improve logging around shutdown

- add a 1 second sleep in `test_rtsp_camera` to make sure it finishes its setup


pascalzauberzeug (approved):
Looks good to me. Can be merged if you tested it with a real camera.

PR #140, opened 2024-06-28
Add Backup schedule

codingpaula:
This PR adds a backup schedule to backup all files in the persistence folder (`~/.rosys`) once a day as suggested here #108.

The location of the backup, the time it's performed and the number of backup files that are kept is configurable.

Changes:
* move writing of an export file to `persistence.registry`
* add BackupSchedule itself
* add documentation

### DIFF
--- a/docs/examples/persistence.md
+++ b/docs/examples/persistence.md
@@ -29,8 +29,8 @@ Note that the persistence module contains a number of helper functions:
 - `replace_set`: replaces the content of a set using `from_dict` for each item
 - `replace_dataclass`: replaces the attributes of a dataclass with the values of a dictionary
 
-The persistence module also provides ui buttons for exporting the contents of your ~/.rosys directory to a single file and for importing such a file.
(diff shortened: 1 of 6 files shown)


rodja:
Cool. Only problem is that if a miss-configuration is not discovered in the time frame, it is still lost. An alternative approach would be Something like log rotatation: compare the stored version with the new data in `persistence.registry.backup()`. If they have changed save the old data in a backup folder with timestamp and then overwrite with new data.

falkoschindler:
@rodja We first wanted to catch cases where we need to restore the configuration, but the last backup is months old.
Automatically backing up after every change to the persistent storage isn't practical in most cases, since we store very frequent data like KPIs. So we would need to create ZIP archives all the time.

rodja:
I have not thought about high frequent updates like kpis. Maybe using git instead of zip would solve both cases? And later adding cloud backups by adding a remote?

PR #142, opened 2024-07-08
Make SimulatedCamera Transformable

rodja:
This pull request adds the `TransformableCamera` mixin to the `SimulatedCamera` to provide crop and and rotation properties. This makes the simulation better interchangeable with the `UsbCamera`.

### DIFF
--- a/rosys/vision/simulated_camera/simulated_camera.py
+++ b/rosys/vision/simulated_camera/simulated_camera.py
@@ -5,11 +5,12 @@ from typing_extensions import Self
 
 from ... import rosys
 from ..camera.configurable_camera import ConfigurableCamera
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):

falkoschindler:
@NiklasNeugebauer What do you think about this PR? Do you have any concerns?

By the way: `TransformableCamera` doesn't transform anything, but only holds properties like `rotation` and `crop`. Would it be possible and make sense to move functions like `process_jpeg_image` and `process_ndarray_image` here so that the `TransformableCamera` knows how to transform an image?

NiklasNeugebauer (approved):

PR #149, opened 2024-07-16
Fix breaking dependency

NiklasNeugebauer:
Issue: With updated packages, running test_examples.py sporadically creates empty `rosys.config.json` (persistence for the `Config` class), which breaks the examples that are run afterwards.
 
My best guess is that there is some kind of race condition where the backup function was called and started clearing the file but did not finish before the task was cancelled.
It seems to me that the errors never impacted normal programs.

I was able to (accidentally) lower the chance of the empty write happening by adding some debug prints (see ef750a98fb8c95985a4452d9465bb56f300ec7ea).
In the end, I decided to first write to a .tmp file and then rename the file. If the write is somehow cancelled, then only the tmp file will be empty. 

Opinions?
@rodja @falkoschindler


falkoschindler (approved):
Ok, the solution seems reasonable.
It just bugs me that we still don't know which package upgrade caused the problem. And we can't be 100% sure it won't affect backup data on real devices.

What do you think, @rodja?

rodja (approved):
Yes, thats fine with me. I don't have a better idea...

Issue #147, opened 2024-07-15
`@track` UI doesn't remove all coroutines when using `parallelize()`


falkoschindler:
The following example shows how shorter coroutines remain in the "track" UI when run with `parallelize` and the longer coroutine is started later:

```py
from nicegui import ui
import rosys

@rosys.analysis.track
async def short():
    await rosys.sleep(0.25)

@rosys.analysis.track
async def long():
    await rosys.sleep(0.75)

@rosys.analysis.track
async def run() -> None:
    await rosys.automation.parallelize(short(), long())

automator = rosys.automation.Automator(None, default_automation=run)

rosys.automation.automation_controls(automator)
rosys.analysis.track.ui()

ui.run()
```

After a while the stack accumulates to something like this:

![Screenshot 2024-07-15 at 17 13 33](https://github.com/user-attachments/assets/5666e60e-5b5d-4e66-b320-babbfcb47302)

PR #150, opened 2024-07-17
Fix @track decorator when used with parallelize()

falkoschindler:
This PR replaces the `stack` list with a dictionary, mapping a unique ID to function names. After running the function, we don't _pop_ the last function name, but delete the item with the respective ID from the dictionary. Since Python dictionaries preserve their item's order, we can simply display the current state by joining all dictionary values.

Has been successfully tested with the code snippet from https://github.com/zauberzeug/rosys/issues/147#issue-2408992831.

### DIFF
--- a/rosys/analysis/tracking.py
+++ b/rosys/analysis/tracking.py
@@ -1,3 +1,4 @@
+import uuid
 from functools import wraps
 from typing import Callable, ParamSpec, TypeVar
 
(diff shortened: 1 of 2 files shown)


NiklasNeugebauer:
Isn't the UI extremely confusing then?

The following would write: run → long → short → short
Even though the last `short` was called by long
```
from nicegui import ui
import rosys

@rosys.analysis.track
async def short():
    await rosys.sleep(2.0)

@rosys.analysis.track
async def long():
    await short()
    await rosys.sleep(2.0)

@rosys.analysis.track
async def run() -> None:
    await rosys.automation.parallelize(long(), short())

automator = rosys.automation.Automator(None, default_automation=run)

rosys.automation.automation_controls(automator)
rosys.analysis.track.ui()

ui.run()
```

falkoschindler:
Yes, the UI is far from optimal in this case. But that's how it has already been before. At least all items are correctly removed now, fixing issue #147.
We'll create a separate feature request for improving the UI for parallel coroutines.

codingpaula (approved):

codingpaula:
Feature Request: https://github.com/zauberzeug/rosys/discussions/151

PR #153, opened 2024-07-19
Chained Pose Transformations

NiklasNeugebauer:
This PR adds `Pose3d` and with it the ability to concatenate poses.

Currently, cameras implicitly define poses through their `Extrinsics` and user code must track which coordinate system this pose is defined in.
`Pose3d` must now always be defined in a parent coordinate frame (or `None` for the world).
This implicitly defines a pose chain that can be used to resolve a pose's relative coordinates into any other frame.

Poses store their frame relationships as ids and retrieve the corresponding object only when needed. This means they can be loaded and stored through persistence without defining the relationships every time.

A pose must be explicitly instatiated as its `CoordinateFrame` subclass if the pose should act as a parent frame for another pose.

## TODO
- [x] ~Should we define an explicit WORLD frame instead of using None ?~
- [x] make Python 3.10 compatible
- [x] use resolved pose in operations inside `Calibration`
- [x] write tests
- [x] write proper example
- [x] document example "coordinate frames"
- [x] review


NiklasNeugebauer:
I tried making a singleton WORLD_FRAME objects but it got pretty complicated.
It should be possible, in the end I just had some infinite recursions when checking for cycles.

Opinions?

rosys/geometry/pose3d.py
```diff
@@ -0,0 +1,168 @@
...
+        :param source_frame: The frame the point is currently in. (default: None = world frame)
+        '''
+        if source_frame is None:
+            relative_pose = self.resolve(up_to=source_frame)
+        else:
+            relative_pose = self.relative_to(source_frame, common_frame=source_frame)
+        return Point3d.from_tuple((relative_pose.rotation.matrix.T @ (point - relative_pose.translation)).tolist())
+
+    def transform_point_to(self, point: Point3d, target_frame: CoordinateFrame | None = None) -> Point3d:
+        '''Transforms the coordinates of a point from the current frame to the target frame.
```
NiklasNeugebauer, line 102: also this one

rosys/geometry/pose3d.py
```diff
@@ -0,0 +1,168 @@
...
+                raise ValueError(f'Could not resolve pose to {up_to} (end of chain)')
+
+            final_pose = current_frame @ final_pose
+            current_frame = current_frame.parent_frame
+
+        final_pose.parent_frame = up_to
+        return final_pose
+
+    def transform_point(self, point: Point3d, source_frame: CoordinateFrame | None = None) -> Point3d:
+        '''Transforms the coordinates of a point from the source frame to the current frame.
```
NiklasNeugebauer, line 90: I would love some naming suggestions for these functions

rosys/geometry/pose3d.py
NiklasNeugebauer, line 102: also this one

falkoschindler (commented):
I started to review this pull request, but haven't finished it yet. So far I have the following thoughts:

- The `scene_object` was in a bad shape (wrong docstring, unused methods), it was almost never used, and such a UI-related class didn't fit well within the geometry module. So I decided to remove it.

- [ ] I rework the example_coordinate_frame.py a bit. But it isn't references in the documentation yet.

- [x] Should we rename "CoordinateFrame" to "Frame3d"? It is shorter and can be easily extended to "Frame" for the 2D case.

- [ ] Implementing `Point3d` as a sequence seemed a bit like overkill to me. I thought we can explicetely use `.tuple` and convert to NumPy arrays where necessary. _After_ removing it and fixing all pytests, I'm somewhat convinced that it wasn't a bad idea... Sorry! 😬

- [ ] How to use the CameraSceneObject? Can we shot it in the coordinate frame example?

- [ ] I don't quite understand `test_rotate()`. Maybe we can find a more readable test case.

- [x] The `parent_frame` and `parent_frame_id` properties are implemented in `Pose3d` as well as in `CoordinateFrame`, which is confusing, because one derives from the other. Is this correct?

- [ ] `common_frame` seems to behave differently for pose and frames, which is confusing because one derives from the other.

In summary: I'll need more time and probably some pairing to understand the meat of the PR, the coordinate frame, pose3d, the changes to calibration and all related tests.

NiklasNeugebauer:
- `scene_object` was meant as a wrapper to be able to easily connect a 3d visualization with a pose. But I agree, I did not give it much thought and time and mainly used it for testing and then the examples.
- I like the rename to Frame3d
- I just saw the "no need to make point3d a sequence" commit and immediately though"no I am pretty sure this was very useful :-)"
- sure, I wanted to keep the example more simple and had thus removed cameras from it, but I think the visualization is quite useful
- `test_rotate` mainly helped me make sense of which order the rotations need to be multiplied in and I thought I'd keep it in as a test. It is supposed to check if the rotation has the right effect by testing its influence on a point
- `CoordinateFrame` checks for loops in the setter of `parent_frame_id`. Setting a parent of a `Pose3d` will never create a loop because it is a leaf node in the chain (since it is no-one's parent).
- `Pose3d` cannot use use `isinstance` with  `CoordinateFrame`, so we can only define the "full" method in `CoordinateFrame`. But we should probably just delete the one in `Pose3d`

NiklasNeugebauer:
I noticed you removed `rotation`, `rotation_array` and `center_point` in calibration.

The rotation functions previously returned the full rotation from both `intrinsics` and `extrinsics`, so I expect they are commonly used.
I think it is fine to remove them since the call may do much more (a full resolution to world) than the user may expect now and the resolution is not really part of the calibration anymore.

We should still remember to explain this in the release notes with a hint of how to replicate the behavior.

falkoschindler:
> I noticed you removed `rotation`, `rotation_array` and `center_point` in calibration.

Yes, I figured the rotation properties had been tailored specifically to the case of intrinsic and extrinsic rotations used in our outside-in tracking project to improve performance. With the new overhead of resolving the pose chain every time, we might be better off by implementing a special solution over their.

The center point was probably a convenience property which doesn't seem to be used anymore (I searched in some of our other projects). But when I think about it, it doesn't do any harm either...

When thinking about the center point, I notice that it has no frame property, which it actually could. So I wonder: If points could _have_ a frame attribute, poses might _have_ one too instead of _being_ a frame (aggregation instead of inheritance). Would that simplify things?

NiklasNeugebauer:
If I understand your suggestion correctly, that's what the parent frame is (which poses already have).
If a pose *is* a frame it just means that other poses can safely use it as their parent frame.

I did consider that points might benefit from parent frames, too.
It would certainly solve the naming problem which you have not commented on, yet :-)

falkoschindler:
Oh, I mixed it up: Frames are poses. Poses are _not_ frames, but _have_ frames. Sorry for the confusion.

Yes, I haven't looked into the naming problem, because I need a fresh mind and probably some guidance for reviewing these functions. Let's do that on Monday. 🙂

rosys/vision/calibration.py
```diff
@@ -242,10 +243,11 @@ def project_from_image(self, image_coordinates: Point | np.ndarray, target_heigh
                 return None
             return Point3d(x=world_points[0, 0], y=world_points[0, 1], z=world_points[0, 2])  # pylint: disable=unsubscriptable-object
 
+        world_extrinsics = self.extrinsics.resolve()
         image_rays = self.points_to_rays(image_coordinates.astype(np.float32).reshape(-1, 1, 2))
-        objPoints = image_rays @ self.rotation_array.T
-        Z = self.extrinsics.translation[-1]
-        t = np.array(self.extrinsics.translation)
+        objPoints = image_rays @ world_extrinsics.rotation.matrix.T
+        Z = self.extrinsics.z
```
falkoschindler, line 249:
```suggestion
        Z = world_extrinsics.z
```

@NiklasNeugebauer Shouldn't we use `world_extrinsics` here as well?

rosys/vision/simulated_camera/simulated_camera.py
```diff
@@ -37,18 +35,13 @@ def __init__(self,
                                  min_value=1, max_value=30, step=1, default_value=fps)
 
     def to_dict(self) -> dict:
-        return {
-            'id': self.id,
+        return super().to_dict() | {
```
falkoschindler, line 38: Don't we need to put the ID in the dictionary as well?

NiklasNeugebauer: Camera does that already

NiklasNeugebauer:
added back `zero()` to point3d and pose3d for convenience and to avoid breaking changes in `point3d`

falkoschindler (approved):
Ok, I think this pull request is complete. Let's merge! 🎉

PR #159, opened 2024-07-26
[DRAFT] properly update settings in rtsp devices

NiklasNeugebauer:


NiklasNeugebauer (commented):
Currently, we use `RtspDevice.update_settings` which updates them in the device.
I would like to do the same as in `MJpegDevice` and instead abstract the settings interface through setters in the device class.

NiklasNeugebauer:
This should be done but I will have to test it with a real device on Monday

NiklasNeugebauer:
This is ready to merge from my side

rosys/vision/rtsp_camera/rtsp_camera.py
```diff
@@ -109,23 +109,24 @@ async def capture_image(self) -> None:
...
-        assert self.device.settings_interface is not None
-        fps = self.device.settings_interface.get_fps(stream_id=self.jovision_profile)
-        return fps
+
+        return self.device.get_fps()
 
     def set_jovision_profile(self, profile: int) -> None:
         assert self.device is not None
-        self.jovision_profile = profile
+
```
falkoschindler, line 123: Don't we need to update `self.jovision_profile` here?

rosys/vision/rtsp_camera/rtsp_camera.py
```diff
@@ -109,23 +109,24 @@ async def capture_image(self) -> None:
...
+        self.device.set_jovision_profile(profile)
 
     def get_jovision_profile(self) -> int | None:
-        assert self.device is not None
-        return self.jovision_profile
+        assert self.device is not None\
```
falkoschindler, line 127:
```suggestion
        assert self.device is not None
```

rosys/vision/rtsp_camera/rtsp_device.py
```diff
@@ -17,9 +17,12 @@ class RtspDevice:
 
     def __init__(self, mac: str, ip: str, jovision_profile: int) -> None:
         self.log = logging.getLogger('rosys.vision.rtsp_camera.rtsp_device')
-        self.log.setLevel(logging.DEBUG)
 
         self.mac = mac
+        self.ip = ip
+
+        self.fps: int
+        self.jovision_profile: int = jovision_profile
```
falkoschindler, line 25:
```suggestion
        self.jovision_profile = jovision_profile
```

falkoschindler (commented):
I've added a few comments...

rosys/vision/rtsp_camera/rtsp_camera.py
NiklasNeugebauer: You're right, the parameter is redundant and should be removed

NiklasNeugebauer:
ready to merge from my side

falkoschindler (approved):

PR #160, opened 2024-07-26
Allow setting parameters on disconnected cameras

NiklasNeugebauer:
Currently, when setting parameters the camera checks if it is connected and simply drops the values if it is not (with the idea that if we cannot set the values on the camera then we cannot do anything).

This PR now saves the values to cache if the camera is disconnected so that upon a reconnect, the values will be applied.


NiklasNeugebauer:
As with #159 I will test this on Monday

pascalzauberzeug (approved):
Looks good to me!

PR #166, opened 2024-08-08
Add relative_direction function to Pose

pascalzauberzeug:
I wanted to know how much the robot has to turn to face a target.
You first have to get the direction to the target and then calculate the difference between the direction and your robot's current yaw.
This PR combines this approach in the `Pose.relative_direction(Point | Pose)` function to make it more easily available.

@falkoschindler the angle parameter of `rotate(self, angle: float)` collides with `rosys.helpers.angle`. I imported it as `helpers_angle` for now, but I don't really like it.

### DIFF
--- a/rosys/geometry/pose.py
+++ b/rosys/geometry/pose.py
@@ -8,6 +8,7 @@ import numpy as np
 from .line import Line
 from .point import Point
 from .point3d import Point3d
+from ..helpers import angle
 
 
 @dataclass(slots=True, kw_only=True)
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Thanks, @pascalzauberzeug!

I changed the import to `from .. import helpers` to avoid conflicts with `angle` without the need for an alias.

When reading the code, I wonder why we use `@overload` for `distance()`, `direction()` and `relative_direction()`. It might be useful in case where, e.g., the return type depends on the input type like here: https://github.com/zauberzeug/rosys/blob/e071d80820dad4340b85cd24ac4ac4229303e149/rosys/vision/calibration.py#L181-L187
or where we want to use a different docstring for each variant. But `distance()` always returns `float` and has no docstring, so I don't see a benefit over a simply union type annotation. Of course, this isn't really the point of this PR. It's just a related thought we could address here or in a subsequent PR.

PR #168, opened 2024-08-12
Provide auto-updated `Odometer.prediction_frame`

rodja:
Thanks to #153 we now can use reference frames to calculate coordinates. This PR provides a `prediction_frame` for the `Odometer` to make using coordinates on the robot more convinient.

### DIFF
--- a/examples/camera_arm/main.py
+++ b/examples/camera_arm/main.py
@@ -58,27 +58,17 @@ class Cam(persistence.PersistentModule):
         self.pose = persistence.from_dict(Frame3d, data['pose'])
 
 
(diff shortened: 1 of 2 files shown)


falkoschindler (approved):
Providing a prediction frame in odometer.py is a great idea!
I wonder if we should use frame IDs like "__odometer_prediction__" or "ODOMETER_PREDICTION" or "rosys.odometer_prediction" for built-in frames to avoid confusion with user-defined frames. What do you think?

rodja:
I like `rosys.odometer_prediction` or `rosys.odometer.prediction` because it indicates the module where it is used.

PR #169, opened 2024-08-12
Add `SimulatedCalibratableCamera`

rodja:
To avoid rewriting the very simple implementation of a `SimulatedCalibratableCamera` in every project, this PR adds it to `rosys.vision`.

~~NOTE: due to circular dependencies, the implementation was placed in the `__init__.py`.~~


falkoschindler (approved):
Strange. I just introduced simulated_calibratable_camera.py and don't get any problems with circular imports. Did I miss something?

rodja:
Strange. It now works for me, too.

PR #170, opened 2024-08-12
Improve documentation for `calibration.project_to_image`

rodja:
I noticed that the docstring was not up-to-date after adding frame-feature #153.


falkoschindler (changes requested):
You're right, the docstring hasn't been updated while adding the new `frame` parameter.

However, the new description doesn't sound right either: The result is not in world coordinates, but in the image plane. So `frame` probably indicates the frame of the input `point`. But if this is the case, why do we need the `frame` parameter at all? Can't `point` be directly defined `.in_frame(frame)`?

It might be required for the `np.array` case, which doesn't automatically come with a frame. In this case, the name `point` is also not ideal.

Maybe we need to different signatures:
```py
@overload
def project_to_image(self, point: Point3d) -> Point | None: ...

@overload
def project_to_image(self, coordinates: np.ndarray, frame: Frame3d | None = None) -> np.ndarray: ...
```

@NiklasNeugebauer What do you think?

NiklasNeugebauer:
I agree with @falkoschindler.
I added this before we added frames to points with the idea of making projection easier from various frames.

Two signatures sounds like the best solution.

falkoschindler:
I reworked the signatures and docstrings once again. Now there's a red test...

falkoschindler:
There's probably a misconception about the `frame` parameter. Earlier I wrote

> So `frame` probably indicates the frame of the input `point`.

Now I think the pytest shows that this is not the case. `frame` seems indeed to be a kind of target frame, in which we want the resulting points or coordinates. But does that makes sense? Who needs this feature? And how to fix the implementation?

falkoschindler (approved):

PR #172, opened 2024-08-13
Do not stop after driving a spline

pascalzauberzeug:
Our robot needs to drive continuously between two GNSS coordinates while adjusting its heading. For each segment we use `drive_spline`, but it stops after completing the spline drive.
This PR adds a parameter for not stopping at the end `stop_at_end` and does some refactoring in the `Driver` class

### DIFF
--- a/examples/hello_bot/main.py
+++ b/examples/hello_bot/main.py
@@ -20,11 +20,11 @@ from rosys.hardware import (
 
 
 async def drive_square():
-    corner_0: Point = Point(x=0.0, y=0.0)
-    corner_1: Point = Point(x=4.0, y=0.0)
-    corner_2: Point = Point(x=4.0, y=4.0)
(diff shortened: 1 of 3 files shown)


falkoschindler:
@pascalzauberzeug Before looking into the details of this PR, can you please

- update the docstrings according to [PEP 257](https://peps.python.org/pep-0257/) (indicative, no new line after the opening triple quotes),
- use [reSt style](https://stackoverflow.com/a/24385103/3419103) (see calibration.py for some good examples), and
- fix examples/hello_bot/main.py which is still using `drive_square`?

Thanks!

By the way, we should probably add a note about docstrings in CONTRIBUTING.md to avoid such back and forth between styles.

pascalzauberzeug:
@falkoschindler Sure, I will look into it.  To admit, copilot wrote those docstrings 😆

falkoschindler (approved):

Issue #162, opened 2024-08-04
`pynma2.parse` can break the process pool


rodja:
When using pynmea2 to parse GNSS messages, we experienced a severe problem.

```py
async def test():
    import pynmea2
    await run.cpu_bound(pynmea2.parse, 'FOOBAR')

rosys.on_repeat(test, 1)
```

This code breaks the process pool:

```
Traceback (most recent call last):
  File "/Users/rodja/Projects/rosys/rosys/rosys.py", line 158, in _repeat
    await invoke(self.handler)
  File "/Users/rodja/Projects/rosys/rosys/helpers/__init__.py", line 37, in invoke
    result = await result
             ^^^^^^^^^^^^
  File "/Users/rodja/Projects/rosys/test.py", line 37, in test
    await run.cpu_bound(pynmea2.parse, 'FOOBAR')
  File "/Users/rodja/Projects/nicegui/nicegui/run.py", line 48, in cpu_bound
    return await _run(process_pool, callback, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rodja/Projects/nicegui/nicegui/run.py", line 31, in _run
    return await loop.run_in_executor(executor, partial(callback, *args, **kwargs))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "uvloop/loop.pyx", line 2729, in uvloop.loop.Loop.run_in_executor
  File "/opt/homebrew/Cellar/python@3.11/3.11.7_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/concurrent/futures/process.py", line 791, in submit
    raise BrokenProcessPool(self._broken)
concurrent.futures.process.BrokenProcessPool: A child process terminated abruptly, the process pool is not usable anymore
finished garbage collection
error in "test"
Traceback (most recent call last):
  File "/Users/rodja/Projects/rosys/rosys/rosys.py", line 158, in _repeat
    await invoke(self.handler)
  File "/Users/rodja/Projects/rosys/rosys/helpers/__init__.py", line 37, in invoke
    result = await result
             ^^^^^^^^^^^^
  File "/Users/rodja/Projects/rosys/test.py", line 37, in test
    await run.cpu_bound(pynmea2.parse, 'FOOBAR')
[...]

Issue #43, opened 2023-10-15
Use run.io_bound and run.cpu_bound from NiceGUI to avoid code duplication


rodja:
We have made `run.io_bound` and `run.cpu_bound` available in NiceGUI and should remove the code duplication in RoSys.

falkoschindler:
I had to revert my commit because it broke tests in other projects.
Especially `running_cpu_bound_processes` needs to be updated so that the tests way for completion.
But maybe there are more things to consider.

rodja:
It would be great to have a minimal example to test the problematic cases.

PR #177, opened 2024-08-19
Use NiceGUI's run.io_bound and run.cpu_bound

falkoschindler:
This PR tries to solve issues #43 and #162 by using NiceGUI's run functions instead of own implementations.

@rodja I removed `process_pool` and `thread_pool`. Do you think this is ok? Or do we need to use our own pools for some reason?

### DIFF
--- a/rosys/run.py
+++ b/rosys/run.py
@@ -6,17 +6,15 @@ import signal
 import subprocess
 import uuid
 from collections.abc import Callable, Generator
-from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
-from concurrent.futures.process import BrokenProcessPool
 from contextlib import contextmanager
-from functools import partial, wraps
+from functools import wraps
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


rosys/run.py
```diff
@@ -137,15 +131,8 @@ def _kill(proc: subprocess.Popen) -> None:
 
 
 def tear_down() -> None:
-    # stopping process as described in https://www.cloudcity.io/blog/2019/02/27/things-i-wish-they-told-me-about-multiprocessing-in-python/
```
rodja, line 140: I think we should keep the link to the process shutdown tips & tricks

rodja (changes requested):

rosys/run.py
rodja: Oh, ok. The article does not describe the things we do in our tear_down. At least not explicitly. We can remove it!

rodja (approved):

PR #184, opened 2024-08-28
add bitrate parameter to RTSP cameras

NiklasNeugebauer:
Adds the bitrate as a parameter. The control was already supported through the jovision interface.

Aditionally, I have made some adjustments to the interface:

- username/ password in the header turned out not to be necessary
- some type hints were missing
- there was an extra channelid in the param dict
- the values to be set are easy to set as an incorrect type without noticing so I added a backup `int` cast


denniswittich (approved):
Looks good. Tests with jovision camera successful.

PR #192, opened 2024-09-09
Add focal length property to Intrinsics

pascalzauberzeug:
This PR adds a property to the camera calibration intrinsics to get the focal length which is calculated from the camera matrix


NiklasNeugebauer:
Taking the mean here seems a little odd since they could be very different if the image pixels are not square.
Consider, what you actually want in your use case (though pixels will probably be square).
Any behavior should be fine as long as it is clearly documented.

pascalzauberzeug:
You are probably right, I found a way to handle it differently in our code

PR #193, opened 2024-09-10
Async Jovision parameter setting

NiklasNeugebauer:
Uses httpx instead of requests to send and request parameters to/from Jovision RTSP cameras.

### DIFF
--- a/rosys/vision/rtsp_camera/jovision_rtsp_interface.py
+++ b/rosys/vision/rtsp_camera/jovision_rtsp_interface.py
@@ -39,6 +39,7 @@ class JovisionInterface:
 
     def __init__(self, ip: str) -> None:
         self.ip = ip
+        self.async_client = httpx.AsyncClient()
 
     @property
     def settings_url(self) -> str:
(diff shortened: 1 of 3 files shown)


rosys/vision/rtsp_camera/jovision_rtsp_interface.py
```diff
@@ -112,11 +114,12 @@ def get_current_settings(self) -> list[JovisionCameraSettings]:
...
-        params = {
-            'cmd': json.dumps(cmd),
-            '_': int(time.time() * 1000),
-        }
-        response = requests.get(self.settings_url, params=params, timeout=1)  # type: ignore
+        params = QueryParams(
+            cmd=json.dumps(cmd),
+            _=int(time.time() * 1000),
+        )
+        async with httpx.AsyncClient() as client:
```
denniswittich, line 121: Generally it is better to instantiate the client in the init function to reduce overhead.

rosys/vision/rtsp_camera/jovision_rtsp_interface.py
```diff
@@ -132,7 +135,7 @@ def get_current_settings(self) -> list[JovisionCameraSettings]:
             for stream in response.json()['result']['streams']
         ]
 
-    def get_parameter_ranges(self):
+    async def get_parameter_ranges(self):
         raise NotImplementedError
```
denniswittich, line 139: Why is there code after the `raise` ?

denniswittich (changes requested):

denniswittich (approved):

PR #194, opened 2024-09-11
Selectable lizard online version in Robot Brain control ui

pascalzauberzeug:
Before you could only download and flash the latest lizard version from the ui.
Now you can select all versions available on Github.

I tested this on our robot. Downloading works, but flashing p0 via ui is still flawed. Flashing manually via terminal did the trick, though.


codingpaula (approved):
Looks good to me 👍

PR #195, opened 2024-09-12
Deprecate camera polling

NiklasNeugebauer:
Since it is not used anymore, I would remove the whole camera polling feature. Some persistence dicts might still contain the polling parameter though, so this keeps it around but does not save it anymore, so that all persistence dicts will be corrected.


NiklasNeugebauer:
This depends on and is now integrated into #197

PR #197, opened 2024-09-16
Camera image callbacks

NiklasNeugebauer:
This PR sends camera images directly from the `Device` to the `Camera` without the camera having to poll on its own.
That causes a much more stable frame-rate and more accurate image timing.

It also improves the processing the `RtspDevice`s Pipe stream. On loop iteration we now read as much data as possible from the pipe and only process the last image. This enables the system to catch up to the real camera stream much faster.


NiklasNeugebauer:
For some reason the `SimulatedCamera` is not capturing images in `test_cameras`.
It works fine in the examples.

falkoschindler (approved):

PR #200, opened 2024-09-19
Fix camera removal

denniswittich:
This PR fixes the removal of cameras in the `multi_camera_provider`. 

The problem was that the camera removal works by modifying the dict returned by `self.cameras`. However, the `multi_camera_provider` overwrites this method to return a joined dict of all sub-providers. Consequently, the deletion had no effect. 
 
This PR also adds a deprecation warning for the `prune_images` function which is outdated because images is a dequeue now.

### DIFF
--- a/rosys/vision/camera_provider.py
+++ b/rosys/vision/camera_provider.py
@@ -1,4 +1,5 @@
 import abc
+import warnings
 from typing import Generic, TypeVar
 
 from .. import persistence, rosys
@@ -57,7 +58,7 @@ class CameraProvider(Generic[T], persistence.PersistentModule, metaclass=abc.ABC
         self.request_backup()
 
     def remove_camera(self, camera_id: str) -> None:
-        del self.cameras[camera_id]
+        del self._cameras[camera_id]
         self.CAMERA_REMOVED.emit(camera_id)
         self.request_backup()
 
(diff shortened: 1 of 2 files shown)


rosys/vision/camera_provider.py
```diff
@@ -66,6 +67,7 @@ def remove_all_cameras(self) -> None:
             self.remove_camera(camera_id)
 
     def prune_images(self, max_age_seconds: float | None = None):
+        warnings.warn('Pruning images is not required anymore because images is now a deque', stacklevel=2)
```
pascalzauberzeug, line 70: @denniswittich  the standard for deprecation warnings seems to be:
```python
import warnings

warnings.warn('Pruning images is not required anymore because images is now a deque', DeprecationWarning, stacklevel=2)
```
which prints this output:
`sys:1: DeprecationWarning: Pruning images is not required anymore because images is now a deque`
instead of this:
`sys:1: UserWarning: Pruning images is not required anymore because images is now a deque`
However, python3 normally ignores deprecation warnings.
You either have to start it via `python3 -Wd your_program.py` or add this to your file: `warnings.simplefilter('always', DeprecationWarning)`

denniswittich: Thanks!

pascalzauberzeug (approved):
Looks good to me. Approved, if it's okay with you @denniswittich that the deprecation warning is probably never shown.

PR #202, opened 2024-09-23
Make image history length configurable

NiklasNeugebauer:
reduce default number of images stored in `Camera` to 4

### DIFF
--- a/rosys/vision/camera/camera.py
+++ b/rosys/vision/camera/camera.py
@@ -16,7 +16,7 @@ from ..image_route import create_image_route
 
 
 class Camera(abc.ABC):
-    MAX_IMAGES = 256
+    MAX_IMAGES = 4
 
     def __init__(self,
                  *,
--- a/rosys/vision/camera/camera.py
+++ b/rosys/vision/camera/camera.py
@@ -16,7 +16,6 @@ from ..image_route import create_image_route
 
 
 class Camera(abc.ABC):
-    MAX_IMAGES = 4
 
     def __init__(self,
                  *,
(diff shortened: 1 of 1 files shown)


pascalzauberzeug:
@NiklasNeugebauer I'm totally for this change, but why did you choose 4?

rosys/vision/camera/camera.py
```diff
@@ -94,6 +95,7 @@ def to_dict(self) -> dict:
             'name': self.name,
             'connect_after_init': self.connect_after_init,
             'streaming': self.should_stream,
+            'image_history_length': self.image_history_length
```
falkoschindler, line 98:
```suggestion
            'image_history_length': self.image_history_length,
```

pascalzauberzeug: Oh yes

pascalzauberzeug:
@falkoschindler @NiklasNeugebauer I wasn't happy with the naming for `MAX_IMAGES`, but I'm not quite sure about `image_history_length` either. What do you think?

rosys/vision/camera/camera.py
```diff
@@ -26,12 +25,14 @@ def __init__(self,
...
                  polling_interval: float = 0.1,
                  base_path_overwrite: str | None = None,
+                 image_history_length: int = 256,
                  **kwargs) -> None:
         super().__init__(**kwargs)
         self.id: str = id
         self.name = name or self.id
         self.connect_after_init = connect_after_init
-        self.images: deque[Image] = deque(maxlen=self.MAX_IMAGES)
+        self.image_history_length = image_history_length
```
falkoschindler, line 34: Maybe we should make this field private, because we don't want the user to change it after instantiation. The deque wouldn't notice a change.

pascalzauberzeug: Or we don't save it at all and use `deque.maxlen` in to_dict

falkoschindler: Great idea!

pascalzauberzeug:
I think it's done and I tested it with this; based on the [persistence example](https://rosys.io/examples/persistence/):
```python
from typing import Any

from nicegui import ui

import rosys
from rosys import persistence
from rosys.vision import SimulatedCamera


class Model(persistence.PersistentModule):
    def __init__(self) -> None:
        super().__init__(persistence_key='camera_test')
        self.camera = SimulatedCamera(id='test_cam', width=800, height=600)

    def restore(self, data: dict[str, Any]) -> None:
        self.camera = SimulatedCamera.from_dict(data)

    def backup(self) -> dict[str, Any]:
        return self.camera.to_dict()


model = Model()
ui.button('Backup', on_click=model.request_backup)
rosys.on_startup(lambda: print(model.camera.images.maxlen))

ui.run(title='RoSys')

```

The new code will fail in projects where we altered MAX_IMAGES, since it doesn't exist anymore. But that's an easy fix

falkoschindler (approved):
Nice!

PR #205, opened 2024-10-02
feature: detector, add source when uploading image

denniswittich:
pass optional information about the image source to the detector so it can be uploaded to the learning loop.
This feature requires a detector node using nodelib version 0.10.13 or higher.
Older detectors will still do the inference but will not transport this information to the learning loop.


codingpaula (approved):
👍

PR #208, opened 2024-10-12
Timelapse: reduce flickering by removing single frames from a different cam

rodja:


rosys/analysis/timelapse_recorder.py
```diff
@@ -150,7 +157,7 @@ def _save_image(image: RosysImage, path: Path, size: tuple[int, int], notificati
     for message in notifications:
         y += 30
         _write(message, draw, x, y)
-    img.save(path / f'{image.time:.3f}.jpg', 'JPEG')
+    img.save(path / f'{image.time:.3f}_{image.camera_id[-5:].upper()}.jpg', 'JPEG')
```
falkoschindler, line 160: I don't think that we need to reduce the camera ID to the "human readable" 2-byte snippets for internal filenames. But it's not a show-stopper for me.

falkoschindler (approved):

rosys/analysis/timelapse_recorder.py
rodja: True. I updated the code accordingly.

PR #209, opened 2024-10-13
Allow svg overlays for timelapse videos

rodja:
- Adds `svg_overlays` property to image to allow storing additional augmentations.
- Render `image.svg_overlays` content with newly introduced dependency `cairosvg` to each timelapse image
- Fix issue with not existing `videos` path

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -24,102 +24,102 @@ files = [
 
 [[package]]
 name = "aiohttp"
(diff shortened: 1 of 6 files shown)


rodja:
@falkoschindler the code is updated as we discussed and ready to review.

falkoschindler (approved):

PR #210, opened 2024-10-13
Awaitable events

pascalzauberzeug:
On multiple occasions we needed to wait for the next call of an event. Instead of waiting in a loop or setting some internal variable this PR allows waiting for an event directly.

```python
from nicegui import ui

import rosys
from rosys.event import Event

event = Event()
event.register(lambda *args: print('Event fired!', *args))


async def wait_for_event_timeout(timeout: float | None = 10.0):
    print('Waiting for event with timeout', timeout)
    value = await event.emitted(timeout)
    print('Event received!', value, type(value))


async def wait_for_event():
    print('Waiting for event')
    value = await event
    print('Event received!', value, type(value))

ui.button('Wait for event with timeout', on_click=wait_for_event_timeout)
ui.button('Wait for event', on_click=wait_for_event)
ui.button('Emit', on_click=lambda: event.emit(42))
rosys.on_repeat(lambda: event.emit(123), interval=5.0)

ui.run(title='RoSys')
```

### DIFF
--- a/rosys/event.py
+++ b/rosys/event.py
@@ -81,6 +81,25 @@ class Event:
             except Exception:
                 log.exception('could not emit listener=%s', listener)
 
+    async def emitted(self, timeout: float | None = None) -> Any:
+        """Waits for an event to be emitted and returns its arguments."""
+        future: asyncio.Future[Any] = asyncio.Future()
+
+        def callback(*args):
+            if not future.done():
+                future.set_result(args[0] if len(args) == 1 else args if args else None)
+
+        self.register(callback)
+        try:
+            return await asyncio.wait_for(future, timeout)
+        except asyncio.TimeoutError as error:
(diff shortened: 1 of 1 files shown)


falkoschindler (commented):
Really nice, @pascalzauberzeug!

One idea though: Couldn't `wait_for` be a method of `Event`, like: `value = await event.emitted(timeout=10.0)`? For clarity `timeout` would be a keyword argument.

We could even implement `Event.__await__` so that you can write `value = await event` (with a default timeout).

pascalzauberzeug:
@falkoschindler I like the `Event.__await__` idea and will implement it when I find the time

pascalzauberzeug:
As we discussed in person, it's now possible to use await directly, without a timeout for an event as well as with a timeout.
```
 value = await event
 value = await event.emitted()
 value = await event.emitted(timeout)
```

falkoschindler (approved):

PR #220, opened 2024-10-30
Persistent image transformation for cameras

NiklasNeugebauer:
adds transformations (rotation and crop) to the persistence dict used by the camera providers


NiklasNeugebauer (commented):
I really dislike that I have to explicitly handle the crop as a dict now. If someone knows a better solution improve it please

rosys/vision/camera/transformable_camera.py
```diff
@@ -6,15 +7,23 @@
 
 class TransformableCamera(Camera):
 
-    def __init__(self, **kwargs) -> None:
+    def __init__(self, crop: Rectangle | dict | None = None, rotation: ImageRotation | int = ImageRotation.NONE, **kwargs) -> None:
         super().__init__(**kwargs)
-        self.crop: Rectangle | None = None
+        self.crop: Rectangle | None = crop if isinstance(
```
rodja, line 12: The line breaking should be fixed.

rodja (approved):
I also have no better idea for the persistence.

denniswittich (approved):

rosys/vision/usb_camera/usb_camera.py
NiklasNeugebauer, line 1: I would keep it in the mixin since it uses attributes from it and might be used by others.
The others are using `img_size.get_size(bytes)`, maybe we should just use this, too

PR #221, opened 2024-10-30
Earlier image timestamping

NiklasNeugebauer:
Gets the timestamp for images in each camera earlier (mainly before any further image procession).
This is not critical but should make the timestamps a little more accurate.


denniswittich (approved):
I hijacked this branch a bit to do some chores. looks good now :)

PR #223, opened 2024-11-04
image route max_size parameter

NiklasNeugebauer:
adds a `max_size` parameter to determine the necessary shrink based on the image size

resulting shrink is always the maximum between (`shrink`, `shrink_from_max_size`)
meaning you can still apply a base shrink AND set a maximum

Here, "size"  (in `max_size`) refers to the maximum dimension of the image. I would like some feedback on that name. 

I was also considering a maximum pixel area instead (e.g. 1.5 MP), but felt like that would be more confusing.

### DIFF
--- a/rosys/vision/image_route.py
+++ b/rosys/vision/image_route.py
@@ -13,6 +13,7 @@ from .. import run
 from .calibration import Calibration
 from .image import Image
 
+
 if TYPE_CHECKING:
     from .camera import Camera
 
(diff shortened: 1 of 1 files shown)


denniswittich:
Max size is fine. For me it was clear it refers to width or height, whatever is larger

denniswittich (approved):
As discussed I extended the EP to allow floating point shrink factors.

falkoschindler (approved):

PR #225, opened 2024-11-05
Fix intersection of parallel line segments

falkoschindler:
When intersecting two parallel line segments, e.g. `(0, 0)--(1, 0)` and `(0, 1)--(1, 1)`, the determinant `d` becomes zero and the division is undefined. We should return `None` in this case.


pascalzauberzeug (approved):
Approved

PR #227, opened 2024-11-07
Add type annotation and kwargs to `run.io_bound` and `run.cpu_bound`

falkoschindler:
As noticed by @denniswittich, IO- and CPU-bound functions in RoSys are missing type annotations for parameters and return values. Apart from that `run.cpu_bound` is missing `kwargs`.

While this PR adds both, it unveils several type errors in connection with `run.cpu_bound` and `run.io_bound`:
```
rosys/run.py:26: error: Missing return statement  [return]
rosys/run.py:28: error: Return value expected  [return-value]
rosys/run.py:46: error: Missing return statement  [return]
rosys/run.py:48: error: Return value expected  [return-value]
rosys/vision/usb_camera/usb_device.py:85: error: Incompatible types in assignment (expression has type "Awaitable[Any] | None", variable has type "tuple[bool, Mat | ndarray[Any, dtype[integer[Any] | floating[Any]]]]")  [assignment]
rosys/vision/usb_camera/usb_camera.py:92: error: Argument 2 to "io_bound" has incompatible type "ndarray[Any, Any]"; expected "list[ndarray[Any, Any]]"  [arg-type]
rosys/vision/usb_camera/usb_camera.py:94: error: Argument 1 to "cpu_bound" has incompatible type "Callable[[bytes, ImageRotation, Rectangle | None], bytes | None]"; expected "Callable[[bytes, ImageRotation, Rectangle | None], bytes]"  [arg-type]
rosys/vision/mjpeg_camera/mjpeg_camera.py:94: error: Argument 1 to "cpu_bound" has incompatible type "Callable[[bytes, ImageRotation, Rectangle | None], bytes | None]"; expected "Callable[[bytes, ImageRotation, Rectangle | None], bytes]"  [arg-type]
```

We should fix them before merging into main.

### DIFF
--- a/rosys/vision/mjpeg_camera/mjpeg_camera.py
+++ b/rosys/vision/mjpeg_camera/mjpeg_camera.py
@@ -91,9 +91,10 @@ class MjpegCamera(TransformableCamera, ConfigurableCamera):
 
     async def _handle_new_image_data(self, image: bytes, timestamp: float) -> None:
         if self.crop or self.rotation != ImageRotation.NONE:
-            image = cast(bytes, await rosys.run.cpu_bound(process_jpeg_image, image, self.rotation, self.crop))
-        if image is None:
-            return
+            image_ = await rosys.run.cpu_bound(process_jpeg_image, image, self.rotation, self.crop)
+            if image_ is None:
+                return
+            image = image_
         try:
             final_image_resolution = get_image_size_from_bytes(image)
         except ValueError:
--- a/rosys/vision/mjpeg_camera/mjpeg_camera.py
+++ b/rosys/vision/mjpeg_camera/mjpeg_camera.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any
+from typing import Any, cast
 
 from typing_extensions import Self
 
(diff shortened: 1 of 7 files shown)


falkoschindler:
Ok, now there's only one mypy issue left related to an `ndarray` in usb_camera.py.
@NiklasNeugebauer Maybe you can have a look. 🙏

NiklasNeugebauer:
@falkoschindler This code for handling raw MJPEG images might not be used by anyone right now. 
I was unable to figure out which type the result of capture.read might have in this case. 
I doubt it is a list, but it might be an array with an extra dimension that is removed here by calling [0].
I decided to keep the current behavior, but execute it in the device and then deliver the raw bytes from there.

rosys/vision/mjpeg_camera/mjpeg_camera.py
```diff
@@ -1,5 +1,5 @@
 import logging
-from typing import Any
+from typing import Any, cast
```
falkoschindler, line 2:
```suggestion
from typing import Any
```

PR #230, opened 2024-11-13
detectors: image creation date

denniswittich:
enable to set custom creation date when uploading an image or when running inference using a detector. This requires detectors that use learning loop node library 0.11.1 or higher.


falkoschindler (approved):
Ok, I just added some line breaks for consistency and some trailing commas for smaller diffs when adding more parameters in the future.
And I don't think we need to explain assertions that are primarily for mypy and can speak for themselves.

NiklasNeugebauer (approved):

PR #231, opened 2024-11-14
Add GNSS hardware and geometry

pascalzauberzeug:
## ToDos:

- [x] add tests
- [x] fix pipeline
- [x] go back to GeoPoint not GeoPose
- [x] everything included that we liked in other implementations?
- [x] is `current_geo_reference` a good name?
- [x] fix test with 45, 179 degrees error (`test_pose_cartesian`)
- [x] Add noise to simulated lat, lon?


rodja:
@falkoschindler, @codingpaula and @pascalzauberzeug I'm still not convinced that passing the reference as a parameter is the right design decision. If you look at https://github.com/zauberzeug/field_friend/pull/231 the code becomes quite more cluttered and complicated. I really think having the reference globally available like we do with [frames](https://github.com/zauberzeug/rosys/blob/675e64c9949d41c6a9e3307c4c5ae10e7f3163f1/rosys/geometry/frame3d_registry.py) would be much better. Parameter passing would only be necessary if there are multiple different reference points to be used in parallel. And that is not something we need to do, right?

pascalzauberzeug:
@rodja The old field_friend implementation was easy to use, but wasn't easy to change and had problems when imported in the wrong way.
This implementation is a hassle in the app code, yes.

Multiple references may not be needed at the same time, but at least the reference should be easily changeable.
Claude.ai also suggested doing it the same way as with frames and I think that keeps the best of both worlds.

Something like this.
```python
# geo_registry.py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from .geo_reference import GeoReference

_global_reference: dict[str, 'GeoReference'] = {}  # Empty dict, will be initialized later

def get_reference(key: str = 'default') -> 'GeoReference':
    """Get the global geo reference."""
    return _global_reference[key]

def set_reference(reference: 'GeoReference', key: str = 'default') -> None:
    """Set the global geo reference."""
    _global_reference[key] = reference
```
That way you can initialize a default reference for the robot, but also can set an extra reference for a special area, which could make local coordinates easier to work with.

rodja:
> Multiple references may not be needed at the same time, but at least the reference should be easily changeable.

What do you mean by "easily changeable, @pascalzauberzeug? Code plasticity, eg. a changeable architecture? Or update-able variable at runtime? For what scenarios would we need the latter? I think it is fine to restart RoSys if the reference point changes.

pascalzauberzeug:
@falkoschindler @codingpaula I think it's done now and can be tested with this script:

```python
from nicegui import ui

import rosys
from rosys.geometry import GeoPoint, GeoReference
from rosys.hardware import GnssSimulation, WheelsSimulation


def main():
    wheels = WheelsSimulation()
    GeoReference.update(GeoReference(GeoPoint.from_degrees(lat=52, lon=7)))
    gnss = GnssSimulation(wheels=wheels)
    with ui.row():
        with ui.column():
            gnss.developer_ui()

    ui.button('Set Reference', on_click=lambda: GeoReference.update(GeoReference(gnss.last_measurement.pose.point)))
    ui.label().bind_text_from(GeoReference.current, 'origin', lambda p: f'Reference: {p}')


rosys.on_startup(main)
ui.run(title='RoSys')
```

First I tried to make current a property and save the reference in _current, but you can't chain `@classmethod` and `@property` anymore. The alternative, a classmethod named get_current or something, didn't look as smooth.
Sadly, I had to remove the `MissingGeoReferenceError`, because when accessing the reference directly there's no clean way to raise it, I think.

falkoschindler (approved):

pascalzauberzeug:
Tested on our robot, will merge now

PR #237, opened 2024-12-11
Check Strapping Pins before Flashing P0

LukasBaecker:
In [Lizard's PR #95] (https://github.com/zauberzeug/lizard/pull/95) a check of the pins that could block the flashing was implemented.

This PR added a check and notifications for the UI to warn the user if flashing cannot be performed.

- [x]  Move EspPins to Robotbrain so that they are accessible
- [x]  get_pin_value function for a single Esp pin
- [x]  Function to check the status of p0 strapping pins
- [x]  Prevent flashing when strapping pin is high


falkoschindler (approved):

pascalzauberzeug:
Just tested it on our robot 👍

PR #243, opened 2025-01-08
improve docstrings in Autoupload enum

denniswittich:
Rephrases the docstrings in the Autoupload enum to better reflect its role as a configuration type rather than an active component.

Changes:
- Rewrites class docstring to describe the enum's purpose as configuration options
- Changes member docstrings from action statements to descriptive mode definitions
- Removes misleading implications about the enum performing actions itself

The enum is a pure type definition, and these changes make its purpose clearer to developers.

### DIFF
--- a/rosys/vision/detector.py
+++ b/rosys/vision/detector.py
@@ -13,16 +13,16 @@ from .uploads import Uploads
 
 
 class Autoupload(Enum):
-    """Configures the auto-submitting of images to the Learning Loop"""
+    """Configuration options for image auto-upload behavior to the Learning Loop"""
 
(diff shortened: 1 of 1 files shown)


NiklasNeugebauer (approved):

NiklasNeugebauer:
> @NiklasNeugebauer Did you check if this variant still works e.g. on ubuntu 22?
tests are running fine without libgl.
The Docker image works with the new packages, though I also doubt it really needs the libraries. 
If anyone has other specific needs, like for OpenCV visualization, they should have this expressed in their system configuration or Dockerfile.

PR #244, opened 2025-01-09
remove detectors upload routine

denniswittich:
Removes the detectors upload mechanism via internal queue.

This way of uploading images was redundant to the async upload method which is from now on the only way to upload images to the learning loop.


rosys/vision/detector_hardware.py
```diff
@@ -46,23 +46,10 @@ def on_sio_disconnect() -> None:
...
     def is_connected(self) -> bool:
         return self.sio.connected
 
-    async def step(self) -> None:
-        if not self.is_connected:
-            self.log.info('trying reconnect %s', self.name)
-            if not await self.connect():
-                self.log.error('connection to %s at port %s failed; trying again', self.name, self.port)
-                await rosys.sleep(3.0)
-                return
```
codingpaula, line 61: I'd say we still need this part to continuously try to reconnect, since we also use the `self.sio` connection for the `upload` function.

rosys/vision/image.py
```diff
@@ -30,10 +31,15 @@ class Image:
...
     is_broken: bool | None = None
-    tags: set[str] = field(default_factory=set)
+    tags: set[str] = field(default_factory=set, metadata={'deprecated': True})
 
     DEFAULT_PLACEHOLDER_SIZE: ClassVar[tuple[int, int]] = (320, 240)
 
+    def __post_init__(self) -> None:
+        if self.tags:
+            warn('The tags field is deprecated and will be removed in a future version.',
+                 DeprecationWarning, stacklevel=2)
```
falkoschindler, line 41: Nice! I didn't know about `warnings.warn`. In NiceGUI I created `warn_once` for this purpose:
https://github.com/zauberzeug/nicegui/blob/3d834e539b4e8f54c96e06b8ee6732bf5960beec/nicegui/helpers.py#L14-L21

But it looks like DeprecationWarnings are hidden by default. Is this ok for us? Or should we simply "forget" to specify the warning type?

rosys/vision/image.py
```diff
@@ -30,10 +31,15 @@ class Image:
     data: bytes | None = None
     _detections: dict[str, Detections] = field(default_factory=dict)
     is_broken: bool | None = None
-    tags: set[str] = field(default_factory=set)
+    tags: set[str] = field(default_factory=set, metadata={'deprecated': True})
```
falkoschindler, line 34: What impact does the "deprecated" flag have? Is it a dataclass feature, or just some convention?

rosys/vision/detector_hardware.py
denniswittich: Uf.. yes absolutely correct!!! Maybe with a larger delay like 10 Secs

codingpaula: Sounds good!

rosys/vision/image.py
denniswittich: seems to be just convention... I'd say it serves as documentation for developers. 
If you don't like it, I am fine with removing this.

rosys/vision/image.py
codingpaula: The [docs](https://docs.python.org/3.11/library/warnings.html#warning-categories) say that it's only ignored if it's not triggered by code in `__main__`. I tested it in another project that's using RoSys, it won't show up with `DeprecationWarning`, but without I get a `UserWarning`.

denniswittich:
@falkoschindler I will assume you merge this PR whenever you are done with the review

falkoschindler (approved):

codingpaula (approved):

PR #246, opened 2025-01-14
dahua profile setting

NiklasNeugebauer:
feat: use substream parameter for dahua cameras

DEPRECATION: rename jovision_profile to substream in `RtspCamera` and `RtspCameraProvider`

### DIFF
--- a/rosys/helpers/deprecation.py
+++ b/rosys/helpers/deprecation.py
@@ -1,0 +1,19 @@
+import warnings
+from functools import wraps
+
+
+def deprecated_param(param_name):
+    """Usage:
+    @deprecated_param('my_param')
+    def my_function(param1, my_param=None):"""
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            if param_name in kwargs:
+                warnings.warn(
+                    f"The '{param_name}' parameter is deprecated and will be removed in future versions.",
+                    DeprecationWarning,
+                )
+            return func(*args, **kwargs)
+        return wrapper
+    return decorator
--- a/rosys/helpers/deprecation.py
+++ b/rosys/helpers/deprecation.py
@@ -13,6 +13,7 @@ def deprecated_param(param_name):
                 warnings.warn(
                     f"The '{param_name}' parameter is deprecated and will be removed in future versions.",
                     DeprecationWarning,
+                    stacklevel=2,
                 )
             return func(*args, **kwargs)
         return wrapper
--- a/rosys/helpers/deprecation.py
+++ b/rosys/helpers/deprecation.py
@@ -3,17 +3,40 @@ from collections.abc import Callable
 from functools import wraps
 
 
-def deprecated_param(param_name: str, *, remove_in_version: str | None = None) -> Callable:
+def deprecated_param(param_name: str, *, remove_in_version: str | None = None, stacklevel: int = 2) -> Callable:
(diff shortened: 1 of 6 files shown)


NiklasNeugebauer:
I would like to rename the `jovision_profile` in `RtspCamera` to `substream` also.
This change would break both user code and and persistence.
Should we mark the argument as deprecated and support both first?

NiklasNeugebauer:
Opinions on the deprecation?

falkoschindler:
I expected the decorator to hide type information of decorated functions. But as far as I can tell it seems to work. Mypy and pylint are happy too. 👍

pascalzauberzeug (approved):
> will be removed in future versions

How do we want to keep track of deprecations? Keep them for x versions, for an amount of time or just until we feel like removing it?

NiklasNeugebauer:
Versions would be easiest to keep track of but is very arbitrary since we have not release schedule or major versions

falkoschindler:
I'd suggest to tag deprecated code with `# DEPRECATED` and mention the version when a feature will be removed. This way we can find and clean up such code more easily in the future.

NiklasNeugebauer:
I added the version to the decorator and a comment where it is not used

pascalzauberzeug:
@NiklasNeugebauer What do you think about a decorator not just for parameters but also for functions and classes?
I like the way it handles _remove_in_version_

rosys/helpers/deprecation.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def deprecated_param(param_name: str, *, remove_in_version: str | None = None) -> Callable:
+    """Mark a function parameter as deprecated."""
+    def decorator(func: Callable) -> Callable:
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            if param_name in kwargs:
+                warnings.warn(
+                    f'The parameter "{param_name}" is deprecated and will be removed in {"Rosys " + remove_in_version if remove_in_version else "a future version"}.',
```
falkoschindler, line 13: I think this will look funny if `remove_in_version` is `None`:
"...and will be removed in Rosys a future version."

By the way, it's "RoSys" with an uppercase "S".

NiklasNeugebauer: ah yes, that should have parentheses. Good catch

NiklasNeugebauer:
> @NiklasNeugebauer What do you think about a decorator not just for parameters but also for functions and classes? I like the way it handles _remove_in_version_. I'm not a fan of the comment

I had the same thought but was too lazy. I'll add one for functions.

rosys/helpers/deprecation.py
falkoschindler: Oh, I'm wrong. `if` has higher precedence, so it should already work correctly.

NiklasNeugebauer:
And I found another bug, the parameter code is really too complicated for this...

rosys/vision/rtsp_camera/rtsp_camera.py
```diff
@@ -73,7 +84,7 @@ async def connect(self) -> None:
             return
 
         self.device = RtspDevice(mac=self.id, ip=self.ip,
-                                 jovision_profile=self.parameters['jovision_profile'],
+                                 substream=self.parameters['jovision_profile'],
```
falkoschindler, line 87: Is it correct to access `self.parameters['jovision_profile']`?

PR #251, opened 2025-01-22
Only show deprecation warnings on call

NiklasNeugebauer:
- add proper wrapper to function decarator
- replace `__init__` in class decorator


NiklasNeugebauer:
Tests are green

falkoschindler (approved):

PR #253, opened 2025-01-23
deal with model_info is none

denniswittich:
There was a small bug when a detector returned the value None for the key 'model_info' in the response of the about dict.

### DIFF
--- a/rosys/vision/detector_hardware.py
+++ b/rosys/vision/detector_hardware.py
@@ -198,7 +198,7 @@ class DetectorHardware(Detector):
             raise DetectorException('Invalid response from detector')
 
         try:
-            model_info_dict = response.get('model_info', {})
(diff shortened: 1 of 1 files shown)


NiklasNeugebauer (approved):

rosys/vision/detector_hardware.py
```diff
@@ -198,7 +198,7 @@ async def fetch_detector_info(self) -> DetectorInfo:
             raise DetectorException('Invalid response from detector')
 
         try:
-            model_info_dict = response.get('model_info', {})
+            model_info_dict = response.get('model_info', {}) or {}
```
falkoschindler, line 201: Tiny simplification:
```suggestion
            model_info_dict = response.get('model_info') or {}
```

falkoschindler (approved):

PR #257, opened 2025-02-05
Use `args_from_dict` in all camera classes

NiklasNeugebauer:
- remove overwrite of `from_dict` in `UsbCamera` and `MjpegCamera`
- use `args_from_dict` in `RtspCamera`

### DIFF
--- a/tests/test_cameras.py
+++ b/tests/test_cameras.py
@@ -4,13 +4,7 @@ import platform
 import pytest
 
 from rosys.testing import forward
-from rosys.vision import (
-    RtspCamera,
-    RtspCameraProvider,
-    SimulatedCamera,
-    UsbCamera,
-    UsbCameraProvider,
-)
+from rosys.vision import RtspCamera, RtspCameraProvider, SimulatedCamera, UsbCamera, UsbCameraProvider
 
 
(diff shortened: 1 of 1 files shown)


codingpaula:
I added some tests for this. What do you think?
If they run successfully, I'd say we're ready to merge.

codingpaula (approved):

NiklasNeugebauer:
> I added some tests for this. What do you think?
Good idea.

Some notes:
I think it is quite confusing to test something that is part of `Camera` inside of a specific test. For example, e.g. `test_storing_configurable_camera_as_dict` is the only one that checks history length and base path are restored. This can be difficult to understand if a test fails and somewhat difficult to maintain.
 
 We do not actually test the functionality that we just fixed. A base `CalibratableCamera` was always able to load its calibration but the subclasses overwrote it.
 
 I'll look into fixing those two points.

NiklasNeugebauer:
seems a little overkill in hindsight but that is what I ended up with

tests/test_cameras.py
falkoschindler, line 1: I don't think we need to wrap this line. It's shorter than 120 characters.

PR #258, opened 2025-02-05
Hashable Points

NiklasNeugebauer:
add `__hash__` to `geometry.Point`


NiklasNeugebauer:
Since the point coordinates are currently mutable (not frozen) we have decided to leave this to the user to implement for now.

PR #264, opened 2025-02-28
typed Events

NiklasNeugebauer:
I have have had a lot of situations where I was not exactly sure which values an event would produce or these values changed but the consumer was not changed correctly with it.

This PR adds a Generic ParamSpec to the `Event` class which enables users to optinally initialize their events as
```py
my_event = Event[int, str]()
```
or equivalently
```py
my_event: Event[int, str] = Event
```
to let the type checker correctly check the types of the callback in `my_event.register()` and the other functions.

### DIFF
--- a/rosys/automation/app_controls_.py
+++ b/rosys/automation/app_controls_.py
@@ -39,7 +39,7 @@ class AppControls:
     """
 
     def __init__(self, robot_brain: RobotBrain, automator: Automator) -> None:
-        self.APP_CONNECTED = Event()
+        self.APP_CONNECTED = Event[[]]()
         """an app connected via bluetooth (used to refresh information or similar)"""
 
         self.robot_brain = robot_brain
--- a/rosys/automation/automator.py
+++ b/rosys/automation/automator.py
@@ -26,22 +26,22 @@ class Automator:
                  steerer: Steerer | None, *,
                  default_automation: Callable | None = None,
                  on_interrupt: Callable | None = None) -> None:
-        self.AUTOMATION_STARTED = Event()
+        self.AUTOMATION_STARTED = Event[[]]()
         """an automation has been started"""
 
-        self.AUTOMATION_PAUSED = Event()
+        self.AUTOMATION_PAUSED = Event[str]()
         """an automation has been paused (string argument: description of the cause)"""
 
-        self.AUTOMATION_RESUMED = Event()
+        self.AUTOMATION_RESUMED = Event[[]]()
         """an automation has been resumed"""
 
-        self.AUTOMATION_STOPPED = Event()
+        self.AUTOMATION_STOPPED = Event[str]()
         """an automation has been stopped (string argument: description of the cause)"""
 
-        self.AUTOMATION_FAILED = Event()
+        self.AUTOMATION_FAILED = Event[str]()
         """an automation has failed to complete (string argument: description of the cause)"""
 
-        self.AUTOMATION_COMPLETED = Event()
+        self.AUTOMATION_COMPLETED = Event[[]]()
         """an automation has been completed"""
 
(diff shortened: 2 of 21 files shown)


NiklasNeugebauer (commented):
The current issue with this is:
MyPy wants a type annotation (even `my_event: Event = Event()` would be enough).
This means either:
1. We find a way to make this work with MyPy without needing the annotation
2. All event initializations need to be updated

In my opinion all events should be typed so I would tend to option 2, but of course that would place a slight burden on users using MyPy, too.

falkoschindler:
@NiklasNeugebauer I also like option 2. But we have to see how it turns out once we fixed all linter and type errors:
Missing type annotations can be easily added for event definitions.
But there are also errors that "Cannot infer type of lambda", where Cursor suggests to simply add `# type: ignore`.

NiklasNeugebauer:
Hm, you're right, it seems simply typing empty events as `Event` is not enough.
The correct way to type empty events would be:
```py
empty = Event[[]]()
empty.register(lambda: print('empty'))
```
We could add a global `EmptyEvent = Event[[]]` to improve ease of use.

Otherwise, we could maybe keep the `Event` class as is and add a `TypedEvent` Subclass?

NiklasNeugebauer:
As far as I can tell, ParamSpec defaults have been introduced in 3.13

https://docs.python.org/3/whatsnew/3.13.html
https://peps.python.org/pep-0696/
https://typing.python.org/en/latest/spec/generics.html#type-parameter-defaults
example from the last link (docs):
```py
DefaultP = ParamSpec("DefaultP", default=[str, int])

class Foo(Generic[DefaultP]): ...

reveal_type(Foo)                  # type is type[Foo[DefaultP = [str, int]]]
reveal_type(Foo())                # type is Foo[[str, int]]
reveal_type(Foo[[bool, bool]]())  # type is Foo[[bool, bool]]
```

This would be very nice to have

NiklasNeugebauer:
This is the best I could come up with for a warning

```py
def __class_getitem__(cls, item):
        print('__class_getitem__', cls, item)
        setattr(cls, '__getitem_called', True)
        return super().__class_getitem__(item)

    def __new__(cls, *args, **kwargs):
        print('__new__', cls, args, kwargs)
        if not hasattr(cls, '__getitem_called'):
            log.warning(
                'The Event class was instantiated without a type argument. '
                'If you wanted to indicate that the event does not have any arguments, '
                'use Event[[]] instead.'
            )
            return Event[[]].__new__(cls)
        return super().__new__(cls)
```

falkoschindler:
I've got some mypy/pylint complaints, so I changed it to
```py
    def __class_getitem__(cls, _item: Any) -> type[Event[P]]:
        setattr(cls, '__getitem_called', True)
        return cls

    def __new__(cls, *args, **kwargs) -> Event[P]:
        if not hasattr(cls, '__getitem_called'):
            log.warning(
                'The Event class was instantiated without a type argument. '
                'If you wanted to indicate that the event is emitted without any arguments, use Event[[]] instead.'
            )
            return Event[[]].__new__(cls)
        return super().__new__(cls, *args, **kwargs)
```

But I wonder if we should raise an exception with a helpful text instead of "fixing" user code. This would also allow to write a pytest more easily.

NiklasNeugebauer:
> I did not test this properly before. This does not really work since all Events are of the same class and thus there is only one single `__getitem_called__` flag. This means if Event was ever called with a type (as is now done by rosys itself), then it will never trigger again.
> 
> I suggest we just don't do it then.

@falkoschindler @codingpaula I would love you to comment on that, is this PR approved then?

falkoschindler (approved):
@NiklasNeugebauer Oh, I didn't notice that the PR is ready for a final review.
But yes, let's merge this one!

PR #268, opened 2025-03-25
Add 3d visualization for Frame3d 

pascalzauberzeug:
I want to share `Frame3dSceneObject` that I built for easier debugging of 3d frames. You only need nicegui's 3d scene and a `Frame3d`, nothing else is needed and they update their positions themselves.

Here I modified the [camera arm example](https://github.com/zauberzeug/rosys/blob/main/examples/camera_arm/main.py)

https://github.com/user-attachments/assets/c5836746-ebcb-4ee9-ad24-41aff7128b4c

I replaced the given page function
https://github.com/zauberzeug/rosys/blob/10c73721be0cd5da5bff8acce540f0e67e94ea9a/examples/camera_arm/main.py#L68-L96

with this:

```python
def page():
    keyboard_control(steerer)
    with ui.row():
        with ui.scene() as scene:
            scene.move_camera(y=-2, z=2)
            frame_3d_object(anchor_frame)
            frame_3d_object(arm1.base, name='Arm 1 Base')
            frame_3d_object(arm2.base, name='Arm 2 Base')
            frame_3d_object(cam.pose, name='Camera')

        with ui.column():
            joystick(steerer, size=50, color='blue')
            ui.slider(min=-math.pi / 2, max=math.pi / 2, step=0.01, value=0, on_change=lambda e: arm1.pitch(e.value))
            ui.slider(min=-math.pi / 2, max=math.pi / 2, step=0.01, value=0, on_change=lambda e: arm2.pitch(e.value))
            ui.slider(min=-math.pi / 4, max=math.pi / 4, step=0.01, value=0, on_change=lambda e: cam.pitch(e.value))
```

### DIFF
--- a/examples/camera_arm/main.py
+++ b/examples/camera_arm/main.py
@@ -6,7 +6,7 @@ from nicegui import ui
 
 from rosys import config, persistence
 from rosys.driving import Driver, Odometer, Steerer, joystick, keyboard_control
-from rosys.geometry import Frame3d, Pose3d, Rotation
+from rosys.geometry import Frame3d, Pose3d, Rotation, frame_3d_object
 from rosys.hardware import RobotSimulation, WheelsSimulation
 
 wheels = WheelsSimulation()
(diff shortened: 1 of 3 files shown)


NiklasNeugebauer:
I like that this derives from group, makes it versatile to be use with custom models.

falkoschindler (commented):
@pascalzauberzeug I just reviewed your code and just made a few minor changes and added the new object to the "camera arm" example. In my point of view, there are only two open questions:

- Is there a better name than `Frame3dObject`? I could imagine naming it `AxesObject` or `CoordinateAxes`.
- The default length of 0.15 seems a bit arbitrary. I tend to set a default of 1.0 and adjust it in user code if needed.

What do you think?

pascalzauberzeug:
@falkoschindler to throw a third one in, how about `FrameAxes`? I don't like the "Object" names, but they are used throughout the code and clearly show that it is an object for the 3d scene. So we probably should stick to it.

Well, it is arbitrary but perhaps more fitting for visualizing real world 3d objects than 1.0m. If you use your hands, it should be more like 0.1m. But from a code perspective, it should be 1.0m, yes

NiklasNeugebauer:
Regarding the name:
I saw this also as an option to simply pin a 3D representation of an object to a frame.
In the robot arm example we could replace the update step entirely with this.

In this case Frame3dObject would work well.
Maybe, this is not the best API for this kind of usage anyway and so keeping it named for the intended use is better.

falkoschindler (approved):

PR #269, opened 2025-03-26
Fix tag upload in detector hardware

NiklasNeugebauer:
Fixes tag upload by properly adding 'tags' to the data dictionary that is sent via socketio.


pascalzauberzeug (approved):
Just tested it on a robot, works!

PR #274, opened 2025-04-11
use generic camera type `T` for `CAMERA_ADDED` event parameter

NiklasNeugebauer:
changes the event parameter type to the already existing generic type so that subtypes use their corresponding camera types (e.g. RtspCameraProvider.CAMERA_ADDED now has type `Event[RtspCamera]`)


codingpaula (approved):
Nice, works as intended! 👍

PR #276, opened 2025-04-13
Allow multiple usb cameras to be processed in parallel

rodja:
This PR reworks the usb camera internals to allow seamless usage of multiple v4l2 devices. There had to be made quite a lot of changes to make this possible:

- improving layout and usability of the camera control example (which was used for testing)
- make sure we only add real capturing devices (linux often creates two devices (video0 and video1 for a single usb camera, where only video0 can be used to capture data)
- set mjpeg format for usb cameras before trying to connect -- otherwise the bandwidth might not be sufficient for multiple cameras to get connected
- use [`linuxpy.video.device`](https://tiagocoutinho.github.io/linuxpy/user_guide/video/) instead of opencv -- because opencv auto-detection for devices does not work in all cases and often comes without gstreamer support in containers; also `linuxpy` provides an async API while `cv2` does not
- simplified `UsbDevice` and renaming it to `V4l2Device` to make clear that it uses the v4l2 APIs from [`linuxpy.video.device`](https://tiagocoutinho.github.io/linuxpy/user_guide/video/)

**ToDos**

- [ ] Is the frame rate fast enough?
- [ ] Implement/rethink setting of width/height/fps/exposure/auto-exposure for `V4l2Device`.
- [ ] Test with four cameras.
- [ ] Test/implement/verify YUV fallback with a single camera.

NOTE: if cameras only show black images they might have gotten into an undefined state. Try unplug/replug them or restart the host system.

### DIFF
--- a/.syncignore
+++ b/.syncignore
@@ -10,7 +10,6 @@ __pycache__/
 .ruff_cache/
 .venv/
 .vscode/
-docs/
 examples/
 tests/
 *.env
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ RUN apt update && apt install -y \
     libgl1 libglx-mesa0 ffmpeg libsm6 libxext6 \
     avahi-utils iputils-ping \
     jq \
+    lsof \
     gstreamer1.0-tools gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-libav \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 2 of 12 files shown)


NiklasNeugebauer:

**Verdict: valuable direction, but the branch is beyond practical rebase — recommend closing and rebooting on current main.** It predates the poetry→uv migration (it edits `poetry.lock`, which no longer exists), the `UsbDevice` it deletes has since evolved (and gains reconnect logic in #423), and the PR's own TODO list (frame rate, 4-camera test, YUV fallback, width/height/fps setters) is unresolved after 14 months.

The linuxpy insights are worth carrying into the reboot:

- `linuxpy.video.device` over cv2 for async capture and reliable enumeration in containers — still the right call.
- Setting MJPG format *before* connecting so USB bandwidth fits multiple cameras — keep.
- Filtering metadata-only `/dev/video*` nodes: the reboot should query V4L2 capabilities instead of the `min(video_ids)` heuristic in `find_video_id`.
- In the new code, avoid: the capture loop ending silently on the first exception (no retry — #423 now defines the pattern), `__del__`-based release of V4L2 handles, and take a fresh look at whether V4L2 capture belongs in a capture subprocess alongside #425.

PR #277, opened 2025-04-14
Add properties to Rotation to get roll, pitch and yaw in degrees

pascalzauberzeug:
This PR adds handy properties to `Rotation` to get roll, pitch and yaw in degrees, like the ones we have in `Pose`.

For context, the old description:

> Our robot's gnss antennas are quite high above the ground and to calculate the correct position on ground level, we have to add the z coordinate to the `antenna_pose`.
> I also added handy degree properties to `Rotation`, like the ones we have in `Pose`.

### DIFF
--- a/rosys/geometry/rotation.py
+++ b/rosys/geometry/rotation.py
@@ -46,12 +46,24 @@ class Rotation:
         return np.arctan2(self.R[2][1], self.R[2][2])
 
     @property
+    def roll_deg(self) -> float:
+        return np.rad2deg(self.roll)
+
+    @property
     def pitch(self) -> float:
         return np.arctan2(-self.R[2][0], np.sqrt(self.R[2][1]**2 + self.R[2][2]**2))
 
     @property
+    def pitch_deg(self) -> float:
+        return np.rad2deg(self.pitch)
+
+    @property
     def yaw(self) -> float:
         return np.arctan2(self.R[1][0], self.R[0][0])
+
+    @property
+    def yaw_deg(self) -> float:
+        return np.rad2deg(self.yaw)
 
     @property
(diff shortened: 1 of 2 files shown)


falkoschindler (commented):
As far as I can tell, this PR only changes `self.antenna_pose: Pose` to `self.antenna_pose: Pose3d` to accommodate an additional z coordinate (which isn't actually used anywhere in gnss.py).

I have some concerns:

1. To avoid yet another breaking change, we could allow for `antenna_pose: Pose | Pose3d | None` and generate a `Pose3d` from `Pose` if needed. This could, however, cause confusion is someone passes a 2D pose, changes a coordinate and assumes that `gnss.antenna_pose` contains the updated coordinate. This is probably a rare edge case though.

2. More importantly: `GnssMeasurement.pose` is supposed to contain the "robot pose". At least that's what `robot_pose` in `_run` suggests. But it doesn't even use the z coordinate.

3. Similar to point 2, one could expect that roll and pitch of the antenna pose is somehow used for computing the robot pose. This is getting pretty complicated, and it's probably even undetermined because we can't measure the roll angle.

When thinking about it, is it even possible to _use_ the z coordinate for the robot pose? Without being able to measure the roll angle, we can't distinguish between roll and translation.

To summarize: Does it really make sense to extend the antenna pose to the third dimension?

pascalzauberzeug:
We talked about this and decided to keep the antenna pose 2d to avoid complications due to typing and keep the module simple.
If your antennas have a special setup, you should correct that in your project's code, like we will do with https://github.com/zauberzeug/field_friend/pull/232.

This PR will keep the property additions in `Rotation` though.

falkoschindler (approved):

PR #278, opened 2025-04-23
Add retry utility for handling function retries

pascalzauberzeug:
We encounter the situation quite often where we want to run a hardware controlling function that can fail because of timings for example. But running it in a loop until it's done is problematic aswell.

So this PR adds the `retry` function wrapper, that runs a given function with a maximum number of tries and and other optional parameters like a timeout and a `on_failed` function that is called when an attempt fails.

@falkoschindler I'm not sure whether we should return a boolean or the return value of the function or both. Probably both?

### DIFF
--- a/rosys/run.py
+++ b/rosys/run.py
@@ -8,7 +8,7 @@ import uuid
 from collections.abc import Callable, Generator
 from contextlib import contextmanager
 from functools import wraps
-from inspect import Parameter, signature
+from inspect import signature
 from pathlib import Path
 from typing import Any, ParamSpec, TypeVar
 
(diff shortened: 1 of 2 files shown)


falkoschindler (commented):
I did some small adjustments to the docstring and added a pytest covering most cases.

But somehow the two callbacks feel a bit strange:
- `on_failed` runs after each failed attempt (sync or async).
- `logging_callback` also runs after each failed attempt, but with two arguments (sync only).

Since they are so similar, we could improve `on_failed` by passing arguments on demand (like https://github.com/zauberzeug/nicegui/blob/3cc700df69baa606a7cf1abeac88465537e3c88f/nicegui/events.py#L423-L426). Then it would fulfil all needs and we could drop the `logging_callback`.

pascalzauberzeug:
@falkoschindler I implemented both our ideas, so `attempt` and `max_attempts` are now optional keyword arguments and `retry` returns both a success boolean and the return value of the function

pascalzauberzeug:
We talked about this and decided to make the function less bloated and more pythonic. Meaning `raise_on_failure` was removed and it will always raise an exception if all attempts fail and we introduced `OnFailedArguments` as an optional argument for `on_failed`. Which makes it extendable in the future and less strict like the keyword arguments before.

falkoschindler (approved):
Great! I just refactored the pytest a little since it got pretty long for a single test case. And using `pytest.raises` is a more convenient way to assert exceptions.
Let's merge! 🚀

PR #279, opened 2025-04-23
Fix Gnss noise calculation in simulation

pascalzauberzeug:
While setting up a test for a new gnss antenna configuration, I noticed that the location noise of the gnss simulation is wrong.
When using `_lat_std_dev=1` and `_lon_std_dev=0`, I got this result which makes sense, because we have two cases:

1. noise_lat is positive, which leads to both a positive direction and distance for polar, resulting in a spread on positive x
2. noise_lat is negative, which leads to both negative values canceling each other out, also resulting in a positive spread

Screenshot 2025-04-23 at 05 49 16

After the fix, I recorded both `_lat_std_dev` and `_lon_std_dev` alternating between 0 and 1, which led to an expected cross:
Screenshot 2025-04-23 at 05 51 04

Setting both to 1 also shows the expected pattern:
Screenshot 2025-04-23 at 06 00 22

### DIFF
--- a/rosys/hardware/gnss.py
+++ b/rosys/hardware/gnss.py
@@ -282,12 +282,11 @@ class GnssSimulation(Gnss):
 
         noise_lat = np.random.normal(0, self._lat_std_dev)
         noise_lon = np.random.normal(0, self._lon_std_dev)
-        noise_length = np.sqrt(noise_lat**2 + noise_lon**2)
-        noise_heading = np.random.normal(0, math.radians(self._heading_std_dev))
-
-        direction = math.atan2(noise_lon, noise_lat)
-        noise_point = geo_pose.point.polar(noise_length, direction)
-        noise_pose = GeoPose(lat=noise_point.lat, lon=noise_point.lon, heading=geo_pose.heading + noise_heading)
+        noise_magnitude = np.sqrt(noise_lat**2 + noise_lon**2)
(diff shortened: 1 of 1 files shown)


falkoschindler (commented):
Oh! Clearly, the original line can't be right:
```py
noise_point = geo_pose.point.polar(noise_lat, direction)
```
But do we really need to use `.polar()`? Can't we simply say
```py
geo_pose = GeoPose.from_pose(self.wheels.pose)
noise_pose = GeoPose(lat=np.random.normal(geo_pose.lat, self._lat_std_dev),
                     lon=np.random.normal(geo_pose.lon, self._lon_std_dev),
                     heading=np.random.normal(geo_pose.heading, math.radians(self._heading_std_dev)))
```

---

By the way: First I thought converting `self._heading_std_dev` to radians is wrong. But according to the documentation it is in degrees, so the conversion is necessary. It is just a bit unconventional because we usually keep angles in radians. Anyway, it's probably not worth a breaking change.

pascalzauberzeug:
@falkoschindler `_lat_std_dev` and `_lon_std_dev` are in meters that's why we can't use them directly. We also can't use GeoPoint.shift_by because that is relative to the orientation of the GeoReference.

I think we kept the in meters and degrees, because the Septentrio SimpleRTK3b returns them like that.

PR #280, opened 2025-04-28
Add configurable logging level and nicegui keyword arguments to notify function

pascalzauberzeug:
The `rosys.notify` function always logged as info from rosys.rosys and did not relay NiceGUI keyword arguments so that i.e. the position could not be changed.

Before:
```
2025-04-28 10:49:53.654 [INFO] rosys/rosys.py:77: Test Message
```
After this change:
```
2025-04-28 10:51:02.194 [your loglevel] your_project/your_file.py:line: Test Message
```

### DIFF
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -71,8 +71,10 @@ def notify(message: str,
                'warning',
                'info',
                'ongoing',
-           ] | None = None) -> None:
-    log.info(message)
+           ] | None = None, *,
+           log_level: int = logging.INFO,
+           **kwargs) -> None:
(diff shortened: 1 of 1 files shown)


pascalzauberzeug:
@falkoschindler I just noticed that it does use the correct log level but doesn't respect the logging configuration of your file, since it still uses the RoSys logger.
Do we keep it that way or do we add the logger as a parameter?

pascalzauberzeug:
We decided to keep it that way. Another idea is to create a seperate logger for RoSys notifcations, but if the need that we can post another PR

PR #281, opened 2025-04-28
Estop and bumper use active

Johannes-Thiel:
The bumper and estops still use the `level` of the Inputs. With the introduction of the `active` property to Lizard in https://github.com/zauberzeug/lizard/pull/38 we can make the code more readable and understandable. This also allows us to invert the inputs for specific robot configurations.
With this change we will need to reconfigure our robots to use `active` instead of `level`. This is a breaking change.

### DIFF
--- a/rosys/hardware/bumper.py
+++ b/rosys/hardware/bumper.py
@@ -31,20 +31,24 @@ class BumperHardware(Bumper, ModuleHardware):
                  expander: ExpanderHardware | None = None,
                  name: str = 'bumper',
                  pins: dict[str, int],
-                 estop: EStop | None = None) -> None:
+                 estop: EStop | None = None,
+                 inverted: bool = False) -> None:
         self.name = name
         self.pins = pins
+        self.inverted = inverted
         lizard_code = ''
         for pin, number in pins.items():
             lizard_code += f'{name}_{pin} = {expander.name + "." if expander else ""}Input({number})\n'
-        core_message_fields = [f'{name}_{pin}.level' for pin in pins]
+            if inverted:
+                lizard_code += f'{name}_{pin}.inverted = true\n'
+        core_message_fields = [f'{name}_{pin}.active' for pin in pins]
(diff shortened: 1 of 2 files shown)


rosys/hardware/bumper.py
```diff
@@ -31,20 +31,24 @@ def __init__(self, robot_brain: RobotBrain, *,
                  expander: ExpanderHardware | None = None,
                  name: str = 'bumper',
                  pins: dict[str, int],
-                 estop: EStop | None = None) -> None:
+                 estop: EStop | None = None,
+                 inverted: bool = False) -> None:
         self.name = name
         self.pins = pins
+        self.inverted = inverted
```
falkoschindler, line 38: This member variable isn't used. Should we remove it?

rosys/hardware/estop.py
```diff
@@ -38,19 +38,22 @@ class EStopHardware(EStop, ModuleHardware):
     The module expects a dictionary of pin names and pin numbers.
     """
 
-    def __init__(self, robot_brain: RobotBrain, *, name: str = 'estop', pins: dict[str, int]) -> None:
+    def __init__(self, robot_brain: RobotBrain, *, name: str = 'estop', pins: dict[str, int], inverted: bool = True) -> None:
         self.name = name
         self.pins = pins
+        self.inverted = inverted
```
falkoschindler, line 44: This member variable isn't used. Should we remove it?

rosys/hardware/bumper.py
Johannes-Thiel: I thought it might be interesting for other parts of the code to know if the pins are inverted, especially for debugging purposes.

rosys/hardware/estop.py
Johannes-Thiel: I thought it might be interesting for other parts of the code to know if the pins are inverted, especially for debugging purposes.

falkoschindler (approved):

PR #283, opened 2025-05-05
Add another dahua mac prefix

denniswittich:


evnchn:
Dahua is one of the most prolific surveilance camera companies from China. They sell tons. Always see those cameras when I watch videos where technicians repair the network. So the name definitely rings a bell. 

Given the high volume, I'm guessing that they are running our of MAC addresses. As a result, they seem to have 31 (!) MAC prefixes which belong to them. 

Source: https://standards-oui.ieee.org/oui/oui.txt, as accessed by https://github.com/bauerj/mac_vendor_lookup/blob/master/mac_vendor_lookup.py, which powers the `mac-vendor-lookup` Python library. 

Note that: 
- `E4246C`, `3CE36B`, `D4430E` and `FC5F49` all belongs to "Zhejiang Dahua Technology Co., Ltd." located at "No.1199,Waterfront Road" / "No.1199, Waterfront Road" post code 310053 (there are 24 MAC prefixes under this)
- While `30DDAA` belongs to "ZHEJIANG DAHUA TECHNOLOGYCO.,LTD" located at "No. 1399, Binxing Road" post code 311200 (4 MAC prefixes under this)
- 3 more MAC prefix is under Dahua at "NO.1199 BinAn Road" / "NO.1199 Bin An Road" post code also 310053

From the postcode, Bin An Road should me Waterfront road. 

So, there's 2 branches of Dahua, then. And the MAC address introduced in this PR is from another branch. 

Let's consider: 
- Are the Dahua cameras from the different branch behave differently? 
- Should we add all prefixes?

evnchn:
All are under Dahua. Confirmed with their WeChat customer support representative. 







But I cannot guarantee the cameras of 31 MAC address prefix all work the same and allow you to go to the path as defined in `mac_to_url`.

NiklasNeugebauer:
Thanks for the heads up. 
Let's be conservative and add them as needed.

PR #285, opened 2025-05-07
Improve GNSS message processing to avoid time discrepancies between NMEA messages and RoSys

rodja:
This PR makes various fixes to ensure we do not run into time discrepancies between NMEA messages and RoSys:

- clearing the serial buffer before starting the processing loop ensures we start fresh and do not spend time processing old messages
-  increasing the baud rate ensures we are reading as fast as possible
- warn when timestamps between RoSys and NMEA diverge more than 50 ms (which is kind of a barrier for a functioning Kalman Filter when driving about 1 km/h)
- debug log the time diff for easier debugging

### DIFF
--- a/rosys/hardware/gnss.py
+++ b/rosys/hardware/gnss.py
@@ -177,7 +177,11 @@ class GnssHardware(Gnss):
                 time_obj = datetime.strptime(nema_timestamp, '%H%M%S.%f').time()
                 utc_time = datetime.combine(today, time_obj).replace(tzinfo=timezone.utc)
                 timestamp = utc_time.timestamp()
-                self.log.debug('timestamp diff = %s', round(timestamp - rosys_time, 3))
+                diff = round(((timestamp - rosys_time + 43200) % 86400) - 43200, 3)
+                if diff > 0.05:  # NOTE: above 50 ms we will get issues with Kalman filter
+                    self.log.warning('timestamp diff = %s', diff)
+                else:
+                    self.log.debug('timestamp diff = %s', diff)
                 if parts[0] == '$GPGGA':
                     if parts[2] == '' or parts[4] == '':
                         continue
--- a/rosys/hardware/gnss.py
+++ b/rosys/hardware/gnss.py
@@ -183,9 +183,9 @@ class GnssHardware(Gnss):
                 field_index = {'$GPGGA': 1, '$GPGST': 1, '$PSSN': 2}
                 if parts[0] not in field_index:
                     continue
-                nema_timestamp = parts[field_index[parts[0]]]
+                nmea_timestamp = parts[field_index[parts[0]]]
                 today = datetime.now().date()
-                time_obj = datetime.strptime(nema_timestamp, '%H%M%S.%f').time()
+                time_obj = datetime.strptime(nmea_timestamp, '%H%M%S.%f').time()
                 utc_time = datetime.combine(today, time_obj).replace(tzinfo=timezone.utc)
                 timestamp = utc_time.timestamp()
                 diff = round(((timestamp - rosys_time + 43200) % 86400) - 43200, 3)
--- a/rosys/hardware/gnss.py
+++ b/rosys/hardware/gnss.py
@@ -18,6 +18,9 @@ from ..event import Event
 from ..geometry import GeoPoint, GeoPose, Pose
 from ..run import io_bound
 
+SECONDS_DAY = 86400
+SECONDS_HALF_DAY = 43200
+
 
 class GpsQuality(IntEnum):
     INVALID = 0
@@ -110,8 +113,6 @@ class GnssHardware(Gnss):
 
     # Maximum allowed timestamp difference in seconds (default is ok for Kalman Filter with about 1 km/h)
     MAX_TIMESTAMP_DIFF = 0.05
-    SECONDS_DAY = 86400
-    SECONDS_HALF_DAY = 43200
 
     def __init__(self, *, antenna_pose: Pose | None, reconnect_interval: float = 3.0) -> None:
         """
@@ -190,7 +191,7 @@ class GnssHardware(Gnss):
                 time_obj = datetime.strptime(nmea_timestamp, '%H%M%S.%f').time()
                 utc_time = datetime.combine(today, time_obj).replace(tzinfo=timezone.utc)
                 timestamp = utc_time.timestamp()
-                diff = round(((timestamp - rosys_time + self.SECONDS_HALF_DAY) % self.SECONDS_DAY) - self.SECONDS_HALF_DAY, 3)
+                diff = round(((timestamp - rosys_time + SECONDS_HALF_DAY) % SECONDS_DAY) - SECONDS_HALF_DAY, 3)
                 if abs(diff) > self.MAX_TIMESTAMP_DIFF:
                     self.log.warning('timestamp diff = %s (exceeds threshold of %s)', diff, self.MAX_TIMESTAMP_DIFF)
                     continue
--- a/rosys/hardware/gnss.py
+++ b/rosys/hardware/gnss.py
@@ -144,6 +144,10 @@ class GnssHardware(Gnss):
         last_gga_timestamp = 0.0
         last_gst_timestamp = 0.0
         last_pssn_timestamp = 0.0
+
+        # Maximum allowed timestamp difference in seconds (default is ok for Kalman Filter with about 1 km/h)
+        MAX_TIMESTAMP_DIFF = 0.05
+
         while True:
             if not self.is_connected:
                 try:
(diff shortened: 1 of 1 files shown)


rosys/hardware/gnss.py
```diff
@@ -217,10 +224,12 @@ def _find_device(self) -> str:
                 return port.device
         raise RuntimeError('No GNSS device found')
 
-    def _connect_to_device(self, port: str, *, baudrate: int = 115200, timeout: float = 0.2) -> serial.Serial:
+    def _connect_to_device(self, port: str, *, baudrate: int = 460800, timeout: float = 0.2) -> serial.Serial:
```
falkoschindler, line 227: Is there an important reason for changing the default baudrate?

rosys/hardware/gnss.py
```diff
@@ -170,7 +172,12 @@ async def _run(self) -> None:
             self.log.debug(sentence)
             parts = sentence.split(',')
             try:
-                timestamp = parts[{'$GPGGA': 1, '$GPGST': 1, '$PSSN': 2}[parts[0]]]
+                nema_timestamp = parts[{'$GPGGA': 1, '$GPGST': 1, '$PSSN': 2}[parts[0]]]
+                today = datetime.now().date()
+                time_obj = datetime.strptime(nema_timestamp, '%H%M%S.%f').time()
+                utc_time = datetime.combine(today, time_obj).replace(tzinfo=timezone.utc)
+                timestamp = utc_time.timestamp()
+                self.log.debug('timestamp diff = %s', round(timestamp - rosys_time, 3))
```
falkoschindler, line 180: I think it would be more correct and more robust (around midnight) to compare times mod 24 hours instead of datetimes. But since it's only used for logging, having a 24-hour difference from time to time is _ok_ I guess...

rosys/hardware/gnss.py
rodja: Yes. Increasing the baud rate ensures we are reading as fast as possible. We saw issues where RoSys time and GNSS time diverged.

rosys/hardware/gnss.py
rodja: Oh, very good. I added the 24h code because its also good to warn if time differs to much.

rosys/hardware/gnss.py
```diff
@@ -140,9 +141,9 @@ async def _run(self) -> None:
         last_num_satellites = 0
         last_hdop = 0.0
         last_altitude = 0.0
-        last_gga_timestamp = ''
-        last_gst_timestamp = ''
-        last_pssn_timestamp = ''
+        last_gga_timestamp = 0.0
+        last_gst_timestamp = 0.0
+        last_pssn_timestamp = 0.0
```
falkoschindler, line 146: When changing these variable types, please update the line
```py
if last_gga_timestamp == last_gst_timestamp == last_pssn_timestamp != '':
```
accordingly. I guess we need to compare with 0.0 instead.

rosys/hardware/gnss.py
```diff
@@ -201,7 +201,7 @@ async def _run(self) -> None:
                     last_pssn_timestamp = timestamp
                     last_raw_heading = float(parts[4] or 0.0)
                     last_heading_accuracy = float(parts[7] or 'inf')
-                if last_gga_timestamp == last_gst_timestamp == last_pssn_timestamp != '':
+                if last_gga_timestamp > 0 and last_gst_timestamp > 0 and last_pssn_timestamp > 0:
```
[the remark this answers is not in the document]
rodja: You are right! And I am sorry 🤦
I think I now have it fixed. We simply need to ensure that all three values are equal. Compare for `> 0` is not necessary because the code is only reached if we have processed a NEMA message with a time stamp.

falkoschindler:
> Compare for > 0 is not necessary because the code is only reached if we have processed a NEMA message with a time stamp.

Oh ok. I would have assumed that the GNSS device might be sending other sentences than the one we are processing. So all three timestamps could be 0.0 at this moment. But than the line
```py
nema_timestamp = parts[{'$GPGGA': 1, '$GPGST': 1, '$PSSN': 2}[parts[0]]]
```
would already break, so you're right, we can't reach the line
```py
if last_gga_timestamp == last_gst_timestamp == last_pssn_timestamp:
```
with all timestamps being 0.0.

pascalzauberzeug:
I will test this later today on another robot

pascalzauberzeug:
@rodja your latest change crashes the serial connection
```
rosys_1     | 2025-05-07 17:00:14.752 [ERROR] nicegui/app/app.py:125: (32, 'Broken pipe')
rosys_1     | Traceback (most recent call last):
rosys_1     |   File "/home/zauberzeug/.local/lib/python3.11/site-packages/nicegui/background_tasks.py", line 52, in _handle_task_result
rosys_1     |     task.result()
rosys_1     |   File "/app/rosys/hardware/gnss.py", line 165, in _run
rosys_1     |     self.serial_connection.send_break(duration=0.25)
rosys_1     |   File "/home/zauberzeug/.local/lib/python3.11/site-packages/serial/serialposix.py", line 701, in send_break
rosys_1     |     termios.tcsendbreak(self.fd, int(duration / 0.25))
rosys_1     | termios.error: (32, 'Broken pipe')
```

rosys/hardware/gnss.py
```diff
@@ -146,7 +148,6 @@ async def _run(self) -> None:
         last_pssn_timestamp = 0.0
 
         # Maximum allowed timestamp difference in seconds (default is ok for Kalman Filter with about 1 km/h)
```
falkoschindler, line 150: outdated comment...

rodja: oh it was just at the wrong position.

rosys/hardware/gnss.py
```diff
@@ -217,10 +240,11 @@ def _find_device(self) -> str:
                 return port.device
         raise RuntimeError('No GNSS device found')
 
-    def _connect_to_device(self, port: str, *, baudrate: int = 115200, timeout: float = 0.2) -> serial.Serial:
+    def _connect_to_device(self, port: str, *, baudrate: int = 460800, timeout: float = 0.2) -> serial.Serial:
         self.log.debug('Connecting to GNSS device "%s"...', port)
         try:
-            return serial.Serial(port=port, baudrate=baudrate, timeout=timeout)
+            serial_connection = serial.Serial(port=port, baudrate=baudrate, timeout=timeout)
+            return serial_connection
```
falkoschindler, line 247: I think we can revert this change:
```suggestion
            return serial.Serial(port=port, baudrate=baudrate, timeout=timeout)
```

rosys/hardware/gnss.py
falkoschindler: Just to be clear: Does this mean we'll have to adjust the device configuration after deploying this PR?

pascalzauberzeug: @falkoschindler no, I just deployed it to F20 and it worked

pascalzauberzeug (approved):

rosys/hardware/gnss.py
```diff
@@ -217,7 +240,7 @@ def _find_device(self) -> str:
                 return port.device
         raise RuntimeError('No GNSS device found')
 
-    def _connect_to_device(self, port: str, *, baudrate: int = 115200, timeout: float = 0.2) -> serial.Serial:
+    def _connect_to_device(self, port: str, *, baudrate: int = 460800, timeout: float = 0.2) -> serial.Serial:
```
pascalzauberzeug, line 243: @rodja I just noticed that the [ROS driver](https://github.com/septentrio-gnss/septentrio_gnss_driver/blob/2d91a27f711a29455df3b4cf79ac8fb87f0209f2/config/gnss.yaml#L8) for our board, that is maintained by the manufacturer, uses  a baudrate for 921600. Should we just use the same?

falkoschindler: We should test it though.

pascalzauberzeug: @falkoschindler I will test it on monday, but I think it should work, because @JensOgorek already tested the ROS node

evnchn: Found more docs: https://docs.px4.io/main/en/gps_compass/septentrio.html#serial-configuration

- Fallback default is `230400`, which `115200` is lower than default, so we are too low to start with. 
- > When using RTK positioning or higher receiver output rates, values above 460800 are recommended.

This leaves **460800 and 921600** as the two standard baud rates you can try your hands on. 

<details>
<summary>Sidenote on 500000, 576000, 1000000, 1500000 baud rates</summary>
I can also see why 500000, 576000, 1000000, 1500000 are non-standard and should be avoided. Recall when I was still doing Arduino in my childhood, depending on the crystal frequency, some baud rates are "better than others". Check https://wormfood.net/avrbaudcalc.php, but this does not quite apply to Septentrio GNSS Receivers because we don't know how it works internally. 
</details>

pascalzauberzeug: 921600 works, I will leave it like that

pascalzauberzeug:
@rodja @falkoschindler ready to merge from my side, do you have anything else?

falkoschindler (approved):

PR #286, opened 2025-05-11
Fix compatibility with NiceGUI 2.16.x

pascalzauberzeug:
We had problems installing the newest NiceGUI version 2.16.1 together with RoSys that we didn't really understand, since the version requirements don't contradict themselves:

```
# 15 19.86 ERROR: Cannot install -r requirements.txt (line 11), -r requirements.txt (line 8) and uvicorn[standard]==0.34.0 because these package versions have conflicting dependencies.
# 15 19.86
# 15 19.86 The conflict is caused by:
# 15 19.86   rosys 0.23.0 depends on uvicorn!=0.29.0
# 15 19.86   nicegui 2.16.1 depends on uvicorn>=0.22.0
# 15 19.86   uvicorn[standard] 0.34.0 depends on uvicorn 0.34.0 (Installed)
```

I still don't quite understand the underlying issue, but requiring `httpx = "^0.25.0"` instead of `httpx = "^0.24.0"` in RoSys does fix it.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -444,13 +444,13 @@ xcb = ["xcffib (>=1.4.0)"]
 
 [[package]]
 name = "cairosvg"
-version = "2.7.1"
+version = "2.8.0"
 description = "A Simple SVG Converter based on Cairo"
 optional = false
-python-versions = ">=3.5"
+python-versions = ">=3.9"
 files = [
-    {file = "CairoSVG-2.7.1-py3-none-any.whl", hash = "sha256:8a5222d4e6c3f86f1f7046b63246877a63b49923a1cd202184c3a634ef546b3b"},
-    {file = "CairoSVG-2.7.1.tar.gz", hash = "sha256:432531d72347291b9a9ebfb6777026b607563fd8719c46ee742db0aef7271ba0"},
+    {file = "cairosvg-2.8.0-py3-none-any.whl", hash = "sha256:f15836feea67a680f79938edf2efdd162793526b0b1cc2fd9ec257ebeafe2946"},
+    {file = "cairosvg-2.8.0.tar.gz", hash = "sha256:d60fe75c238ffba3f8e21e06dad8727ecd2bcef8d3d55fc11f9286c08c5cc710"},
 ]
 
 [package.dependencies]
@@ -461,7 +461,7 @@ pillow = "*"
 tinycss2 = "*"
 
 [package.extras]
-doc = ["sphinx", "sphinx-rtd-theme"]
+doc = ["sphinx", "sphinx_rtd_theme"]
 test = ["flake8", "isort", "pytest"]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):

evnchn:
@pascalzauberzeug Did you upgrade pip? Before upgrading pip (on 24.3.1) I indeed get the same message as you. After upgrading to pip 25.1.1, I get this instead. Try it on your machine, too!

```
ERROR: Cannot install httpx and nicegui==2.16.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    nicegui 2.16.1 depends on h11>=0.16.0
    httpcore 0.17.3 depends on h11<0.15 and >=0.13
    httpcore 0.17.2 depends on h11<0.15 and >=0.13
    httpcore 0.17.1 depends on h11<0.15 and >=0.13
    httpcore 0.17.0 depends on h11<0.15 and >=0.13
    httpcore 0.16.3 depends on h11<0.15 and >=0.13
    httpcore 0.16.2 depends on h11<0.15 and >=0.13
    httpcore 0.16.1 depends on h11<0.15 and >=0.13
    httpcore 0.16.0 depends on h11<0.15 and >=0.13
    httpcore 0.15.0 depends on h11<0.13 and >=0.11
```

@falkoschindler I think the issue may have to be intrinsically with NiceGUI itself. Though a simple fix in RoSys is much appreciated, we should proceed with more care and caution.

evnchn:
https://poe.com/s/oNjKLZHR8DJHL6ys77hw

https://github.com/encode/httpx/compare/0.24.1...0.25.0#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

image

https://github.com/encode/httpcore/commit/d1e17c53993fb5f3bfafedcd5aaca52a7264f0ed

{F3B55F17-2B5C-4560-B039-737186A01870}

{7B6C3530-A53C-46B4-B01A-36CA36DD0BEF}

evnchn:
Boiled things down further:

```
pip install nicegui==2.16.1
pip install rosys==0.23.0
```

Then:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
rosys 0.23.0 requires httpx<0.25.0,>=0.24.0, but you have httpx 0.28.1 which is incompatible.
```

**Realization: `^0.24.0` actually means `<0.25.0,>=0.24.0`**

So RoSys before the PR excluding [httpx 0.25.1](https://github.com/encode/httpx/releases/tag/0.25.1) (aka the release of httpx where httpcore 1.0 support was introduced) means it marks itself incompatible with h11 0.16.0, which is only supported by [httpcore 1.0.9](https://github.com/encode/httpcore/releases/tag/1.0.9) as of two weeks ago. 

Therefore, it is right to do `https = "^0.25.0"` in RoSys, and not touch NiceGUI in any way. Though I wonder why not use `>=`

falkoschindler:
@evnchn `httpx = "^0.25.0"` means >=0.25.0 but <1.0.0. This can be a reasonable limitation in case 1.0.0 introduces breaking changes in the future.

evnchn:
From the above, where `"^0.25.0"` is resolved to `httpx<0.25.0,>=0.24.0`, I'm not sure if that's the case.

codingpaula:
According to the poetry documentation [here](https://python-poetry.org/docs/dependency-specification/#caret-requirements), `^0.24.0` means `>=0.24.0, <0.25.0` since its major version is below 1. NiceGUI has `>=0.24.0` for httpx and uses `0.28.1` (see poetry.lock), RoSys has `^0.24.0` which causes the clash.

pascalzauberzeug:
@evnchn @codingpaula and I decided to use `>= 0.25.1` and just fix it when/if it breaks

falkoschindler (approved):

PR #288, opened 2025-05-15
Add gnss timestamp to `GnssMeasurement`

Johannes-Thiel:
When working with gnss it can be useful to access the timestamp of the `GnssMeasurement` that has come from the gnss device. At the moment we only use the `rosys.time`, so the time that the `NEW_MEASURMENT` Event is callled and not the actual time of the nss measurement in the gnss device. To introduce the timestamp of the incoming Gnss measurment we add the `gnss_time` property to the `GnssMeasurement` dataclass.

### DIFF
--- a/rosys/hardware/gnss.py
+++ b/rosys/hardware/gnss.py
@@ -317,10 +317,10 @@ class GnssSimulation(Gnss):
         noise_point = geo_pose.point.polar(noise_magnitude, noise_direction)
         noise_heading = np.random.normal(geo_pose.heading, math.radians(self._heading_std_dev))
         noise_pose = GeoPose(lat=noise_point.lat, lon=noise_point.lon, heading=noise_heading)
-
+        timestamp = rosys.time()
         self.last_measurement = GnssMeasurement(
-            time=rosys.time(),
-            gnss_time=rosys.time(),
+            time=timestamp,
+            gnss_time=timestamp,
             pose=noise_pose,
             latitude_std_dev=self._lat_std_dev,
             longitude_std_dev=self._lon_std_dev,


rosys/hardware/gnss.py
```diff
@@ -318,6 +320,7 @@ def simulate(self) -> None:
 
         self.last_measurement = GnssMeasurement(
             time=rosys.time(),
+            gnss_time=rosys.time(),
```
pascalzauberzeug, line 323:
```suggestion
        timestamp = rosys.time()
        self.last_measurement = GnssMeasurement(
            time=timestamp,
            gnss_time=timestamp,
```

How about this to make sure the timestamps are the same

evnchn:
Just to confirm, the SimpleRTK3b does not output any timestamp data in its serial output, right? (which is odd since I thought it is already talking to the GPS network so surely it'd know the time from there)

If it did (or if it's possible to turn on the feature), and we did not extract it from serial, then we would be wasting the GPS time calibration that we would otherwise get for free and which can enable easier multi-device operation. 

My reference: https://www.ardusimple.com/how-to-connect-ardusimple-simplertk2b-3b-receivers-to-arduino/

image

Doesn't look like a timestamp?

pascalzauberzeug:
@evnchn the first number after the type is the timestamp without the date.
In your example `$GNGGA,093052.002,...` stands for 9:30 AM and 52.022 seconds.
For the full timestamp we could activate the [ZDA message](https://receiverhelp.trimble.com/alloy-gnss/en-us/NMEA-0183messages_ZDA.html).

You are correct, we currently don't take advantage of the direct synchronization with the GNSS module

falkoschindler (approved):

pascalzauberzeug (approved):

evnchn:
As it stands this PR is OK. But I need to clarify on my former point, otherwise I fear I lead you astray. 

Personally I believe if you do want to read the timestamp from the receiver, then you also need to have a clock somewhere which is synchronized to the receiver, and only use said clock with reference to receiver-provided timestamps. Best if that clock is also the system clock, then you don't have to worry about which clock to use. 

Otherwise, if the system clock and the receiver's clock is out-of-sync (which it sure is), and we try and compare system time to the timestamp from the receiver (`rosys.time() - timestamp_from_receiver`), then we may risk having wonky stuff such as the code calculating that it is receiving messages in the past. 

Having a time reference which is consistent within each robot but _inconsistent_ across robots, is **much better** than a time reference which is _inconsistent_ within each robot but consistent across robots. 

In short, either always use system timestamp, or always use GPS timestamp (at least for GPS-time-sensitive code). There should only be one source of truth. Latter is better than the former, of course, but a mix is not acceptable.

pascalzauberzeug:
@evnchn Yes we will always use `rosys.time()` in code. Adding `gnss_time` in this case is more for debugging and information purposes.
If we want to sync the robot with GNSS time, we should sync the system clock of the jetson without touching it in RoSys

PR #292, opened 2025-05-22
Improve ESP ready state handling

pascalzauberzeug:
We had issues where hardware commands were sent before the ESP was done booting and instead of just trying to re-send the commands, this PR does:

1. Add `RobotBrain.is_ready` property based on `RobotBrain.hardware_time`
2. Make use of new `espresso.py` optional for enable and reset, but keep `flash.py` as default for now
3. Raise new `EspNotReadyException`, when trying to send a Lizard command when the ESP is not ready
4. Remove old enable etc. notifications, that were not accurate and quite spamy
5. Implement new features in `AppControls` and `BmsHardware`, because they failed when I tested this PR

### DIFF
--- a/rosys/automation/app_controls_.py
+++ b/rosys/automation/app_controls_.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 import inspect
+import logging
 from collections.abc import Callable
 from dataclasses import dataclass
 
 from .. import rosys
 from ..event import Event
-from ..hardware import RobotBrain
+from ..hardware import EspNotReadyException, RobotBrain
 from .automator import Automator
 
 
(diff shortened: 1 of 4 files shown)


rosys/automation/app_controls_.py
```diff
@@ -53,7 +56,7 @@ def __init__(self, robot_brain: RobotBrain, automator: Automator) -> None:
         }
         self.extra_buttons: dict[str, AppButton] = {}
 
-        rosys.on_startup(self.sync)
+        rosys.on_startup(self._startup_sync)
```
pascalzauberzeug, line 59: @falkoschindler I just thought about not running sync `on_startup` but just register it to the `ESP_CONNECTED`

falkoschindler (approved):

PR #295, opened 2025-05-23
Add a notification to warn about wrong Lizard configuartion

Johannes-Thiel:
### Motivation
When working on robots and introducing new properties to be checked with lizard, it is easy to forget to configure the robot. This is also relevant when changing branches on the robot. This PR aims to notify the user that there is something wrong with the configuration.

### Implementation
To detect a faulty configuration we are comparing the number of arguments in the `output_fields` with the length of the provided `core_message`. I chose this way of comparison due to the ease of implementation. This will cover most wrong configurations, the only problem will be if two configurations have the same number of paramters, but these parameters are diffrent. This is not a usual case since in practice we mostly add new parameters or remove them. Not change the parameter.

We add `+2` the the `output_length` due to the `core_message` looking like this:
``` 
core <millis> <parameters>
```
And needing to cover `core` and `<millis>` too.

I added a notification timer, in order to prevent message spam. This was chosen, so that the user will be continually informed about the problem and not just once.

### Progress
- [x] Tested on real hardware.

### DIFF
--- a/rosys/hardware/robot.py
+++ b/rosys/hardware/robot.py
@@ -58,7 +58,7 @@ class RobotHardware(Robot):
             rdyp.on()
             en3.on()
         ''')
-        self.output_length = len(output_fields)+2  # add 2 for core and core.millis
+        self.output_length = len(output_fields) + 2  # account for the two words "core" and `core.millis`
         return code
 
     async def update(self) -> None:
--- a/rosys/hardware/robot.py
+++ b/rosys/hardware/robot.py
@@ -74,10 +74,10 @@ class RobotHardware(Robot):
                 if len(words) != self.output_length:
                     current_time = rosys.time()
                     if current_time - self._last_warning_time >= self._warning_cooldown:
-                        rosys.notify(
-                            f'Lizard output configuration is incorrect, expected {self.output_length} parameters, got {len(words)}',
-                            type='negative',
-                            log_level=logging.ERROR)
+                        rosys.notify('Lizard output configuration is incorrect, '
+                                     f'expected {self.output_length} parameters, got {len(words)}',
+                                     type='negative',
+                                     log_level=logging.ERROR)
                         self._last_warning_time = current_time
                     return
                 words.pop(0)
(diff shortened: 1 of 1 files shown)


rosys/hardware/robot.py
```diff
@@ -32,6 +32,9 @@ class RobotHardware(Robot):
     def __init__(self, modules: list[Module], robot_brain: RobotBrain) -> None:
         super().__init__(modules)
         self.robot_brain = robot_brain
+        self.output_length = 0
+        self._warning_cooldown = 30.0
+        self._last_warning_time = 0.0
```
pascalzauberzeug, line 37:
```suggestion
        self.output_length: int = 0
        self._warning_cooldown: float = 30.0
        self._last_warning_time: float = 0.0
```

rosys/hardware/robot.py
```diff
@@ -67,6 +71,15 @@ async def update(self) -> None:
...
                 continue
             if words[0] == 'core':
+                if len(words) != self.output_length:
+                    current_time = rosys.time()
+                    if current_time - self._last_warning_time >= self._warning_cooldown:
+                        rosys.notify(
+                            f'Lizard output configuration is incorrect, expected {self.output_length} parameters, got {len(words)}',
+                            type='negative',
+                            log_level=logging.ERROR)
+                        self._last_warning_time = current_time
```
pascalzauberzeug, line 81: I don't really like the idea of having to throttle the logging here. Like we talked about, we should rework the `rosys.notify` function to allow for that.

rosys/hardware/robot.py
falkoschindler: I don't think we need explicit type annotation when assigning values with the same type. Type checkers can infer the type reliably.

rosys/hardware/robot.py
```diff
@@ -55,6 +58,7 @@ def generate_lizard_code(self) -> str:
             rdyp.on()
             en3.on()
         ''')
+        self.output_length = len(output_fields)+2  # add 2 for core and core.millis
```
falkoschindler, line 61:
```suggestion
        self.output_length = len(output_fields) + 2  # account for the two words "core" and `core.millis`
```

rosys/hardware/robot.py
falkoschindler: Let's discuss this subject in person.

falkoschindler (approved):

PR #301, opened 2025-06-04
Use GC natively

denniswittich:
### Motivation

Rosys based systems lagged once every minute when the gc.collect() was called. 

### Implementation

Remove the mechanism that removd the native gc and instead collected once every minute.

### Progress

- [ ] Discuss if the manual collection should be optional
- [ ] Remove unused code

### DIFF
--- a/rosys/analysis/memory.py
+++ b/rosys/analysis/memory.py
@@ -50,24 +50,24 @@ def compare_tracemalloc_snapshots(snapshot, prev_snapshot):
 
 def observe_memory_growth(with_tracemalloc: bool = False) -> None:
     log.info('Observing memory growth')
-    prev_memory: int = 0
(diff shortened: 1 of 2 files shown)


evnchn:
By the way, if GC performance is important, consider `gc.collect(2)` then `gc.freeze()` after initialization and just before the server starts, so that those objects which are not collected at time are not considered ever in the GC process.

NiklasNeugebauer:
@evnchn yes, good point. We should look into that since we do have a lot of long living objects after settings up most systems (and nicegui, rosys)

evnchn:
PyPy and NiceGUI is not going to work without a fight...

And I'm fighting over at https://github.com/zauberzeug/nicegui/discussions/5159

💪🏻

PR #303, opened 2025-06-10
Make throttle at end ramp configurable

SeaTechRC:
### Motivation
Currently, the stopping behaviour with throttle_at_end in Driver.drive_spline is not configurable.
This leads to robots (such as the Field Friend) taking a long time to come to a stop.
This could be optimized, but the ramp settings are not configurable.

### Implementation
To allow for changing of the throttling behaviour, we simply swap the ramp arguments with values we add to the driver parameters.
Thus, we can set the min speed (as a factor) and the distance to start throttling.
Perhaps, the code needs some more changes if the throttle distance is larger than the hook distance.

### Progress

### DIFF
--- a/rosys/driving/driver.py
+++ b/rosys/driving/driver.py
@@ -25,7 +25,8 @@ class DriveParameters(ModificationContext):
     hook_bending_factor: float = 0
     minimum_drive_distance: float = 0.01
     throttle_at_end_distance: float = 0.5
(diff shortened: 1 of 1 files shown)


pascalzauberzeug (approved):

falkoschindler:
Thanks for the pull request, @SeaTechRC!
Making these parameters configurable is a great idea.
Let's merge it and create a separate PR later in case large throttle distances turn out to be an important use case.

PR #306, opened 2025-06-23
Disable persistence in tests

pascalzauberzeug:
### Motivation

We noticed that the persistence failed to setup multiple pytest fixtures, because the persistence keys will already be claimed by the first fixture that is set up.
Clearing the instances at the start of the fixture with `Persistable.instances.clear()` helped, but we decided that persistence should not be tested in our projects' code anyway. As it also wasn't before the rework.

### Implementation

Before setting up the directory and claiming the persistence key, the `Persistable` will check if it is running in a pytest environment. In that case it will just return `self` without doing anything.
If you need to test the persistence, this can be overwritten with the `allow_tests` parameter that is `False` by default.

### Progress

- [x] Fix cyclic import from rosys.is_test

### DIFF
--- a/rosys/persistence/persistable.py
+++ b/rosys/persistence/persistable.py
@@ -12,7 +12,6 @@ from nicegui import run
 from typing_extensions import Self
 
 from .. import core
-from ..helpers import is_test
 
 KEY_PATTERN = re.compile(r'^[a-zA-Z0-9_\-\.]+$')
 
@@ -48,7 +47,7 @@ class Persistable(abc.ABC):
         :param backup_interval: The interval to backup the object (default: None)
         :param disable_in_tests: Whether to disable persistence in tests (default: True)
         """
-        self._disabled = is_test() and disable_in_tests
+        self._disabled = core.is_test and disable_in_tests
         if self._disabled:
             return self
         if self._filepath is not None:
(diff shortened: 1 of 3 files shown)


falkoschindler:
@pascalzauberzeug What cyclic import do you mean? Alls checks are green. What's wrong with the current implementation?

rosys/persistence/persistable.py
```diff
@@ -31,6 +32,7 @@ def persistent(self, *,
                    restore: bool = True,
                    backup_check_interval: float | None = 10.0,
                    backup_interval: float | None = None,
+                   allow_tests: bool = False,
```
falkoschindler, line 35:
```suggestion
                   allow_in_tests: bool = False,
```
Maybe `allow_in_tests` is a better name?

rosys/persistence/persistable.py
```diff
@@ -103,6 +108,8 @@ def sync_backup(self) -> None:
         To avoid blocking the main thread, use ``request_backup()`` instead.
         """
         if self._filepath is None:
+            if is_test():
```
falkoschindler, line 111: Why do we need these extra checks in `sync_backup()` and `sync_restore()`?
Any why only `if self._filepath is None`?

falkoschindler (approved):

rosys/persistence/persistable.py
pascalzauberzeug: We introduced the `_disabled` variable to circumvent this

PR #307, opened 2025-06-24
save the position, from witch images have been taken.

SeaTechRC:
### Motivation

The detection process for an image detector can take time (or could be delayed, etc.). During this time a robot could move. The position prom witch the image has been taken is not saved, thus it is not possible to accurately calculate the positions of detected objects in an image.

### Implementation

The concept is to save the pose from witch an image has been taken as soon as possible after receiving the image data, thus saving the most accurate position possible.

A potential trade-off is additional memory usage. Compared to the image data, this should be minimal.

### Progress

### DIFF
--- a/rosys/vision/calibration.py
+++ b/rosys/vision/calibration.py
@@ -253,15 +253,16 @@ class Calibration:
         """
 
     @overload
-    def project_from_image(self, image_coordinates: FloatArray, target_height: float = 0) -> FloatArray:
+    def project_from_image(self, image_coordinates: FloatArray, target_height: float = 0, frame: Frame3d | None = None) -> FloatArray:
         """Project points in image coordinates to a plane in world coordinates.
 
         :param image_coordinates: (Nx2) The image coordinates to project.
         :param target_height: The height of the plane in world coordinates.
+        :param frame: The coordinate frame from which to project the image coordinates to.
         :return: (Nx3) The world coordinates of the projected points.
         """
 
-    def project_from_image(self, image_coordinates: Point | list[Point] | FloatArray, target_height: float = 0) -> Point3d | None | list[Point3d | None] | FloatArray:
+    def project_from_image(self, image_coordinates: Point | list[Point] | FloatArray, target_height: float = 0, frame: Frame3d | None = None) -> Point3d | None | list[Point3d | None] | FloatArray:
         if isinstance(image_coordinates, Point):
             return self.project_from_image([image_coordinates], target_height=target_height)[0]
         if not isinstance(image_coordinates, np.ndarray):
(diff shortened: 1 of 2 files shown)


rosys/vision/calibration.py
```diff
@@ -253,23 +253,24 @@ def project_from_image(self, image_coordinates: list[Point], target_height: floa
...
+    def project_from_image(self, image_coordinates: Point | list[Point] | FloatArray, target_height: float = 0, frame: Frame3d | None = None) -> Point3d | None | list[Point3d | None] | FloatArray:
         if isinstance(image_coordinates, Point):
             return self.project_from_image([image_coordinates], target_height=target_height)[0]
         if not isinstance(image_coordinates, np.ndarray):
             image_array = np.array([point.tuple for point in image_coordinates], dtype=np.float64)
             world_array = self.project_from_image(image_array, target_height=target_height)
             return [Point3d(x=point[0], y=point[1], z=point[2]) if not np.isnan(point).any() else None for point in world_array]
 
-        world_extrinsics = self.extrinsics.resolve()
+        world_extrinsics = self.extrinsics.resolve() if frame is None else self.extrinsics.transform_with(frame.resolve())
```
pascalzauberzeug, line 273: I think you can just use:
```suggestion
        world_extrinsics = self.extrinsics.relative_to(frame)
```
Because `resolve` also just calls it with `None`. [Link](https://github.com/zauberzeug/rosys/blob/af186412fe38e4c10597927df2d72a7995be3ee5/rosys/geometry/object3d.py#L27-L45)

NiklasNeugebauer:
What is the motivation for adding the frame parameter to `project_from_image`? Why only this one overload?

SeaTechRC:
I followed the overloads of `project_to_image`. Perhaps the frame kwarg should be present in all overloads?

The motivation is, that currently, `project_from_image` always resolves the extrinics (world position) when called.
If the image significantly older than the call to `project_from_image`, it will calculate a world position for the camera, that is different from the world position the image has been taken from.

NiklasNeugebauer:
The `project_to_image` functions are called either with `Point3d` or a numpy array. `Point3d` can have a frame attached to it since every point must be in some frame. For numpy arrays we wanted to give the option to also provide one.

In `project_to_image`, your implementation adds the ability to specify a new target frame in which the horizontal plane lies that the points are projected onto. There would be no reason not to add it to all overloads.

However, while this _can_ be quite useful, esp. since it implicitly adds the ability to project to arbitrary planes,
I don't quite see how it relates to the problem of outdated camera poses.
I believe what you want is the ability to specify a frame from which to proejct the points (i.e. an alternative to `self.extrinsics`.

pascalzauberzeug:
@NiklasNeugebauer I think you are right. 
In the use case of old data, you would have to resolve `self.extrinsics` when taking the image, to later be able to project a pixel to the correct position at that point in time, correct? So `world_extrinsics` has to be an argument.

NiklasNeugebauer:
> Currently, it also resolves the extrinsics with the given frame instead of using the given frame as the extrinsics.

The frame parameter is different from you suggestions. This is the frame in which the points to be projected lie. We should not remove that.

PR #311, opened 2025-07-08
Allow creating repeaters in stopped state

pascalzauberzeug:
### Motivation

I encountered the situation where I wanted to create a `Repeater` to start and stop it later, but it will always start when created.
In this case I didn't want that.

Using this does not work, because when `Repeater.stop()` is called, the repeater is not running yet:
```python
repeater = rosys.on_repeat(FUNCTION, INTERVAL)
repeater.stop()
```

### Implementation

This is why I propose the `start_on_creation` parameter for `rosys.on_repeat`, that is `True` by default.
I'm not sure with the naming, since `start_on_creation` is a little long. At first I just used `start` which I didn't like because the function that is called, is also named `start()`.

This PR will not break anything and will not change the default behaviour.

### Progress

### DIFF
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -198,9 +198,9 @@ class Repeater:
             repeater.cancel()
 
 
-def on_repeat(handler: Callable, interval: float, start: bool = True) -> Repeater:
+def on_repeat(handler: Callable, interval: float, start_on_creation: bool = True) -> Repeater:
     repeater = Repeater(handler, interval)
-    if start:
+    if start_on_creation:
         repeater.start()
     return repeater
 
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -198,9 +198,10 @@ class Repeater:
             repeater.cancel()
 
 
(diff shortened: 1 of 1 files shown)


falkoschindler (commented):
@pascalzauberzeug

- > At first I just used `start` which I didn't like because the function that is called, is also named `start()`.

    Why is this a problem? To me it seems like a very intuitive name.

- On the other hand,
    ```py
    repeater = rosys.on_repeat(..., start=False)
    ```
    sounds like a contradiction. `on_repeat` says "do something" and `start=False` says "don't do it".

- What about instantiating a repeater like this:
    ```py
    repeater = rosys.Repeater(...)
    ```
    This creates the repeater without starting it. And the code is very readable.

So maybe we don't need to change the API at all and keep `on_repeat` as it is: always "doing something".

pascalzauberzeug:
@falkoschindler Oh yes, I didn't think about instantiating `Repeater` directly, because I always just used `on_repeat`, thanks. I will close this

PR #312, opened 2025-07-10
add camera parameter `h265`

denniswittich:
### Motivation

Add support for h265 codec

### DIFF
--- a/rosys/vision/rtsp_camera/rtsp_device.py
+++ b/rosys/vision/rtsp_camera/rtsp_device.py
@@ -17,7 +17,8 @@ from .vendors import VendorType, mac_to_url, mac_to_vendor
 class RtspDevice:
 
     def __init__(self, mac: str, ip: str, *,
-                 substream: int, fps: int, on_new_image_data: Callable[[bytes, float], Awaitable | None]) -> None:
+                 substream: int, fps: int, on_new_image_data: Callable[[bytes, float], Awaitable | None],
+                 h265: bool = False) -> None:
         self._mac = mac
         self._ip = ip
         self.log = logging.getLogger('rosys.vision.rtsp_camera.rtsp_device.' + self._mac)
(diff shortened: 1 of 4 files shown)


rosys/vision/rtsp_camera/rtsp_device.py
```diff
@@ -103,7 +105,10 @@ async def stream() -> AsyncGenerator[bytes, None]:
             url = self.url
             self.log.debug('[%s] Starting gstreamer pipeline for %s', self._mac, url)
             # to try: replace avdec_h264 with nvh264dec ! nvvidconv (!videoconvert)
-            command = f'gst-launch-1.0 rtspsrc location="{url}" latency=0 protocols=tcp ! rtph264depay ! avdec_h264 ! videoconvert ! jpegenc ! fdsink'
+            if self._h265:
+                command = f'gst-launch-1.0 rtspsrc location="{url}" latency=0 protocols=tcp ! rtph265depay ! avdec_h265 ! videoconvert ! jpegenc ! fdsink'
+            else:
+                command = f'gst-launch-1.0 rtspsrc location="{url}" latency=0 protocols=tcp ! rtph264depay ! avdec_h264 ! videoconvert ! jpegenc ! fdsink'
```
falkoschindler, line 111:
```suggestion
            avdec = 'h265' if self._h265 else 'h264'
            command = f'gst-launch-1.0 rtspsrc location="{url}" latency=0 protocols=tcp ! rtp{avdec}depay ! avdec_{avdec} ! videoconvert ! jpegenc ! fdsink'
```

falkoschindler (approved):
Looks good to me. I just have one minor suggestion for reducing code duplication.

falkoschindler:
> I dislike the use of a bool where a user can choose between two options.

Ok, but I personally prefer literals over enums. From the user's perspective I can simply type the value without needing to import the enum first. And we all use type checkers nowadays, don't we?

NiklasNeugebauer:
> I personally prefer literals over enums.

I like the extra abstraction of enums,
but I'm fine with literals here long as they're typed properly.

PR #320, opened 2025-08-12
add a pnp wrapper to SpatialResection

NiklasNeugebauer:
### Motivation

We currently have the method `lsa_with_lines` in the `SpatialResection` class.
`lsa_with_lines` computes a camera pose from image to world point correspondences using generic least squares adjustment. Here, the correspondences can be exact point to point or line correspondences where we know that a given image point lies on a line in the world.  

While flexible, the use of generic least squares adjustment means that the solution depends strongly on the initial guess. Even with a good guess the result will likely not be the optimal solution.

This PR adds a second method which implements various solutions of the pnp (`perspective n points`) problem, some of which are shown to be optimal in specific circumstances.

### Implementation

This PR adds a second method `pnp_with_lsa` which acts as a wrapper for the `cv2.solvePnP` function from opencv.
We try to pick the optimal method flag for the function depending on the input. For example, there is a special algorithm for coplanar world points which we automatically pick if the input points are close to coplanar.

As a last step the function refines the pose further using `cv2.solvePnPRefineLM`.

### Progress


falkoschindler:
@NiklasNeugebauer Can you, please, provide some description for this PR? I'm not even sure what "pnp" is supposed to mean.

NiklasNeugebauer:
@falkoschindler I added the motivation and implementation section. Sorry about that.

PR #328, opened 2025-08-27
Introduce `@automation.uninterruptible` decorator

falkoschindler:
### Motivation

@pascalzauberzeug wondered if we could implement a decorator for marking parts of an automation as uninterruptible so that they keep running until completed when pausing an automation.

### Implementation

With the help of GPT-5 I came up with such a decorator that seems to work nicely:

```py
async def a():
    for i in range(10):
        print(f'a - {i}')
        await asyncio.sleep(0.1)

@rosys.automation.uninterruptible
async def b():
    for i in range(10):
        print(f'b - {i}')
        await asyncio.sleep(0.1)

async def c():
    for i in range(10):
        print(f'c - {i}')
        await asyncio.sleep(0.1)

async def automation():
    await a()
    await b()
    await c()

automator = rosys.automation.Automator(steerer=None, default_automation=automation)
ui.button('start automation', on_click=automator.start)
ui.button('pause automation', on_click=lambda: automator.pause('button clicked'))
ui.button('resume automation', on_click=automator.resume)
ui.button('stop automation', on_click=lambda: automator.stop('button clicked'))
```

While "a" and "c" pause immediately, "b" keeps counting until reaching "c - 0".

### Progress

- [x] Let decorator add `sleep(0)` after calling the function.
- [x] Find better name? "uninterruptible"? ~~"unstoppable"?~~
- [x] Documentation has been added.
- [x] Check parallel automations.

### DIFF
--- a/docs/examples/automations/README.md
+++ b/docs/examples/automations/README.md
@@ -1,0 +1,9 @@
+# Automations
+
+This example demonstrates how to start, pause, resume and stop automations.
+The robot will drive forward and then turn left in an endless loop.
+The `turn_left` function is marked with `@rosys.automation.uninterruptible` which means that it will continue to run even if the automation is interrupted by pausing or stopping it.
+
+```python
+{! examples/automations/main.py !}
+```
--- a/docs/examples/automations/main.py
+++ b/docs/examples/automations/main.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+import math
+
+from nicegui import ui
+
+import rosys
+import rosys.helpers
+
+
+async def run() -> None:
+    while True:
+        await forward()
+        await turn_left()
+
+
+async def forward() -> None:
+    start = odometer.prediction
+    while start.distance(odometer.prediction) < 1.0:
+        await wheels.drive(1.0, 0.0)
+        await rosys.sleep(0.1)
+    await wheels.stop()
+
+
+@rosys.automation.uninterruptible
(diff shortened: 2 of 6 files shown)


pascalzauberzeug (changes requested):
While b continues uninterupted correctly, I think the first loop of c should not be run:
```
b - 5
b - 6
b - 7
b - 8
b - 9
It should stop here!
c - 0
```

which leads to a bigger problem, when we create a b_twice function, which runs the atomic b two times, but itself is not atomic:
```
async def b_twice():
    await b()
    # it should pause here, but b loops through two times before pausing
    await b()

async def automation():
    await a()
    await b_twice()
    await c()
```

pascalzauberzeug (commented):
Looks good for now. I'm curious which problems we encounter with parallel automations

falkoschindler:
Two separate automators seem to work independently - as expected. (I basically duplicated the code above and tried to break it.)

But an uninterruptible coroutine `b` prevents parallel coroutines `a` and `c` from being interrupted:
```py
async def automation():
    await rosys.automation.parallelize(a(), b(), c())
```

We could argue that this behavior is ok, because the coroutines wouldn't be "parallel" if one of them is paused and another one keeps running.

pascalzauberzeug:
@falkoschindler maybe there is a way to check in `rosys.automation.parallelize` if one of the coroutines is uninterruptible and print a warning if that's the case

PR #337, opened 2025-09-17
Handle projection of 0 points

denniswittich:
The current implementation will trow errors when one of the projection functions is called with an empty array or a list of zero points. This is quite annoying when projecting an automatically generated set of points.


NiklasNeugebauer (approved):

denniswittich:
@NiklasNeugebauer what was your motivation to test with empty arrays? Empty point lists are converted to empty arrays anyways, or did I miss something?!

NiklasNeugebauer:
Just for completeness to be less dependent on implementation.
Tests should still work if the underlying implementation changes.

PR #341, opened 2025-09-23
Add image recorder / replay module

denniswittich:
### Motivation

We needed a lightweight solution to record image data and play them back for development purpose. 

### Implementation

Adds `ImageRecorder` that saves all incoming images for one camera provider.
Adds `ReplayCameraProvider` that acts as a camera provider but emits the previously recorded images.

### Progress

### DIFF
--- a/rosys/vision/__init__.py
+++ b/rosys/vision/__init__.py
@@ -13,6 +13,7 @@ from .detector_simulation import DetectorSimulation, SimulatedObject
 from .image import Image, ImageSize
 from .mjpeg_camera import MjpegCamera, MjpegCameraProvider
 from .multi_camera_provider import MultiCameraProvider
+from .record_replay import ImageRecorder, ReplayCamera, ReplayCameraProvider
 from .rtsp_camera import RtspCamera, RtspCameraProvider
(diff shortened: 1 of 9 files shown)


NiklasNeugebauer:
I am done with my part.

I changed the ReplayCamera to *only* emit an image if the time is past its timestamp. Before, we would at least emit the first one. A request of an earlier timestamp currently does not happen but I like the guarantee :)

Otherwise, for testing we need real image data that can be loaded, so I made the `SimulatedDevice` generate real images even during tests. We would have to clear that with the others, esp. @rodja @falkoschindler @pascalzauberzeug
I doubt this causes any issues.

NiklasNeugebauer (approved):

denniswittich:
@NiklasNeugebauer I fixed two bugs:
- Time was practically evolving when replay was paused because _last_update_time was not updated on play.
- When jumping to zero (stop) only one camera jumped to the first frame because your search algorithm returned -1

After these fixes I am happy now. Let's go with this version for now

NiklasNeugebauer (dismissed):

PR #345, opened 2025-10-07
Migrate to NiceGUI 3.0

falkoschindler:
### Motivation

Now that NiceGUI 3.0 is out, we should make sure RoSys is compatible.

### Implementation

- Upgrade NiceGUI dependency to ^3.0.0.
- Add `sanitize=false` to `ui.html`.
- Fix imports for `ui.scene`.
- Remove the condition on `client.shared`.
- Use `on_delete` instead of `on_disconnect`.
- Adjust to new upload event arguments.

### Progress

    - [ ] Should we use a sanitizer for `ui.html` on the object graph page?
    - [ ] Should we remove or at least deprecate RoSys events in favor of NiceGUI's `Event`?
- [x] Pytests are not necessary.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2157,14 +2157,14 @@ test = ["codecov (>=2.1)", "pytest (>=7.2)", "pytest-cov (>=4.0)"]
 
 [[package]]
 name = "nicegui"
-version = "2.16.1"
+version = "3.0.3"
 description = "Create web-based user interfaces with Python. The nice way."
 optional = false
-python-versions = "<4.0,>=3.8"
+python-versions = "<4.0,>=3.9"
 groups = ["main"]
 files = [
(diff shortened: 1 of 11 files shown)


falkoschindler:
We noticed that previous tests relied on a special mechanism that checked event handling tasks for exceptions while running `forward` and raised a RuntimeError if there were any. After moving the `Event` class over to NiceGUI, this special treatment is gone. And we think that's good, because there won't be a RuntimeError in production either. The exception will only be printed to the command line.

We thought about letting tests fail if there are unexpected error logs. But I have difficulties getting it to work, because the `rosys_integration` fixture rewrites the log configuration so that NiceGUI logs are not captured anymore. I think, however, we can postpone this change because it doesn't have much to do with the upgrade to NiceGUI 3.0. Logs haven't been checked before, so we can do that in a separate PR.

PR #346, opened 2025-10-07
Add dynamic versioning for Poetry

pascalzauberzeug:
### Motivation

For a long time now, the `pyproject.toml` file showed the version of RoSys to be 0.1.0.
That lead to pip also thinking that RoSys is on version 0.1.0 while we are already working towards 0.30.0 right now.

In our FieldFriend repository I used [extra release commits](https://github.com/zauberzeug/field_friend/commit/fdaac9a41f4b40075ffd7b705fcffc0110b018d7) to tackle this, but we didn't like this solution because it adds an extra step that is also error prone.
We also couldn't just edit the version in our build pipeline, because if the build pipeline adds another commit, the tagged version will not contain this commit. Which means, the version info will always lag behind one release.

### Implementation

[poetry-dynamic-versioning](https://github.com/mtkennerly/poetry-dynamic-versioning) to the rescue!
This will automatically generate the correct version info when you install RoSys.
When you install this branch via pip it will return this, since this branch has added 8 commits after 0.29.1 and is currently on commit `32bee1a9`.

```
Successfully installed r0.29.1.post8.dev0+32bee1a9
```

When installed locally, it will also inform about uncommitted changes. With `dirty`, like this:
```
❯ poetry-dynamic-versioning show
0.29.1.post8.dev0+32bee1a9
```

~The default `pep440` style is quite long `0.29.1.post4.dev0+a1cd5875` and hasn't got the `dirty` tag. That's why I (or more the AI) added a custom jinja format.~

I tested it locally and installed the branch via pip. That works fine. I didn't test the build pipeline yet, but that should work the same.

### Progress


falkoschindler (commented):
Great idea to use dynamic versioning, @pascalzauberzeug! I never heard of it, but it looks very promising.

Just two thoughts:

1. With the removal of setuptools we probably should also remove setup.py. I'm not too familiar with these build tools, but this is how GPT-5 argues:
    > The project now builds via PEP 517 (poetry-core with poetry-dynamic-versioning) defined in pyproject.toml, so setup.py is unused and relies on `distutils` (removed in Python 3.12). This avoids confusion around deprecated `python setup.py ...` workflows; builds/installs should use modern PEP 517 tooling (poetry, pip) instead."

2. I lean towards removing the Jinja template and simply stick with the PEP440 default. When building locally, maybe with uncommitted changes, the version number isn't too important. And even if it is used somewhere, a few characters more or less don't matter.

pascalzauberzeug:
@falkoschindler I removed setup.py and the Jinja format.
Sadly, with the way we inject our local changes into the robot's code, poetry can't detect the dirty tag, since it's not installed.

falkoschindler (approved):

PR #350, opened 2025-10-20
Introduce `MultiQueueLazyWorker`

ftilde:
This can be seen as a generalization of the `LazyWorker` with a configurable number of slots for running coroutines and fair scheduling between a number of different queues where each queue has a size of one where newer requests displace older ones.

### Motivation

On a system with multiple cameras, but one (or multiple) shared detector(s), images from one camera can repeatedly displace the detector-requests from another when using `LazyWorker` (for example because of differing frame rates). `MultiQueueLazyWorker` manages access to the shared resource of detectors fairly in that case. Having multiple running-slots enables better utilization of detector resources, e.g. in cases where we have multiple detectors running in a pipeline.

### Implementation

In contrast to `LazyWorker` the `run` method takes a `queue_id` argument, which is used to distinguish clients for fair distribution. This could be a `camera_id` in the above example. For each queue, we keep track of the last time a coroutine was scheduled to run and when selecting a new coroutine we use this time as the criterion for scheduling.
For multiple running slots, a Semaphor-like counter is used.
Synchronization is achieved via a state-enum and an `asyncio.Event` per queue.

Notably, the implementation is not thread-safe (although obviously async-safe, i.e. when not preempted between await-points).

### Progress

### DIFF
--- a/rosys/helpers/__init__.py
+++ b/rosys/helpers/__init__.py
@@ -13,12 +13,12 @@ import numpy as np
 from nicegui import app
 
 from .lazy_worker import LazyWorker
-from .multi_client_displacement_scheduler import MultiClientDisplacementScheduler
+from .multi_queue_lazy_worker import MultiQueueLazyWorker
 
 __all__ = [
     'LazyWorker',
     'ModificationContext',
-    'MultiClientDisplacementScheduler',
+    'MultiQueueLazyWorker',
     'PackagePathFilter',
     'angle',
     'eliminate_2pi',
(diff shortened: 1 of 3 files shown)


rosys/helpers/multi_client_displacement_scheduler.py
```diff
@@ -0,0 +1,118 @@
...
+from collections.abc import Coroutine
+from enum import StrEnum
+from typing import Any, TypeVar
+
+import rosys
+
+_T = TypeVar('_T')
+
+
+class MCDSSubmissionState(StrEnum):
```
denniswittich: I would remove the MCDS prefix for the helper classes (as they are only used within the class)

denniswittich (commented):
Let's discuss the open suggestions

PR #356, opened 2025-10-29
Change image representation to pixel array

ftilde:
### Motivation
Currently, the internal representation of `Image` is data, an encoded bytes array (usually assumed to be JPEG bytes). This is useful, if we assume that we get a JPEG from a camera and directly present it as such to the user, without any further processing. The moment we do any [processing](https://github.com/zauberzeug/rosys/blob/2f0b816dc7f2b7e2b9b3da2b2806aaa73307b9cb/rosys/vision/image_processing.py#L35), this advantage falls out of the window and we do more work that we must.

### Implementation

The `data` property no longer exists. Instead, the `array` property can be used to interact with the image data. `Image.from_jpeg_bytes` can be used to conveniently construct an image in places where JPEG bytes are given, and `Image.to_jpeg_bytes` when JPEG bytes are required for further processing.
Notably, array can no longer be `None` (i.e., we can no longer have images without any pixel data). I found no problems with that in the rosys code base, but I'm not sure if there are any use cases in downstream projects.

⚠️ The `bytes` property has been removed from `Image`. Use `from_jpeg_bytes`/`to_jpeg_bytes` (or preferably `from_array` and the new `array` property) instead.
⚠️ The properties `is_broken` and `tags` have been removed from `Image`.
⚠️ Default parameter values of `Image.create_placeholder` have been changed.
⚠️ `get_image_size_from_bytes` has been removed
⚠️ `image_history_length` parameter has been removed from the `Camera` constructor. Use the global `IMAGE_HISTORY_LENGTH` instead. The default value has also reduced from 256 (or 128 for some specialized Cameras) to 16.

### Progress

### DIFF
--- a/rosys/vision/calibration.py
+++ b/rosys/vision/calibration.py
@@ -527,9 +527,7 @@ class Calibration:
 
         if isinstance(image, Image):
             image_array = self.undistort_image(image.array, crop=crop)
-            img = Image.from_array(image_array, camera_id=image.camera_id, time=image.time)
-            img.metadata = image.metadata
-            return img
+            return Image.from_array(image_array, camera_id=image.camera_id, time=image.time, metadata=image.metadata)
 
         if image.shape[0] != self.intrinsics.size.height or image.shape[1] != self.intrinsics.size.width:
             log.warning('Image size does not match calibration size (image: %s, calibration: %s)',
--- a/rosys/vision/image.py
+++ b/rosys/vision/image.py
@@ -32,7 +32,7 @@ class Image:
     time: float  # time of recording
     _array: ImageArray
     _detections: dict[str, Detections] = field(default_factory=dict)
-    metadata: dict[str, Any] = field(default_factory=dict)
+    metadata: dict[str, Any]
 
     DEFAULT_PLACEHOLDER_SIZE: ClassVar[tuple[int, int]] = (320, 240)
 
@@ -64,30 +64,43 @@ class Image:
         return f'{self.camera_id}/{self.time}'
 
     @classmethod
-    def create_placeholder(cls, text: str, *, camera_id: str = 'from_placeholder', time: float | None = None, shrink: int = 1) -> Self:
+    def create_placeholder(cls, text: str, *,
+                           camera_id: str = 'from_placeholder',
+                           time: float | None = None,
+                           metadata: dict[str, Any] | None = None,
+                           shrink: int = 1) -> Self:
         h, w = cls.DEFAULT_PLACEHOLDER_SIZE
         img = PIL.Image.new('RGB', (h // shrink, w // shrink), color=(73, 109, 137))
         d = PIL.ImageDraw.Draw(img)
         d.text((img.width / 2 - len(text) * 3, img.height / 2 - 5), text, fill=(255, 255, 255))
-        return cls.from_array(np.array(img), camera_id=camera_id, time=time)
+        return cls.from_array(np.array(img), camera_id=camera_id, time=time, metadata=metadata)
 
     @classmethod
-    def from_pil(cls, pil_image: PIL.Image.Image, *, camera_id: str = 'from_pil', time: float | None = None) -> Self:
+    def from_pil(cls, pil_image: PIL.Image.Image, *,
+                 camera_id: str = 'from_pil',
+                 time: float | None = None,
+                 metadata: dict[str, Any] | None = None) -> Self:
         """Create an image from a PIL image."""
-        return cls.from_array(np.array(pil_image), camera_id=camera_id, time=time)
+        return cls.from_array(np.array(pil_image), camera_id=camera_id, time=time, metadata=metadata)
 
     @classmethod
-    def from_jpeg_bytes(cls, jpeg_bytes: bytes, *, camera_id: str = 'from_jpeg_bytes', time: float | None = None) -> Self | None:
+    def from_jpeg_bytes(cls, jpeg_bytes: bytes, *,
+                        camera_id: str = 'from_jpeg_bytes',
+                        time: float | None = None,
+                        metadata: dict[str, Any] | None = None) -> Self | None:
         """Create an image from plain JPEG bytes. This runs a jpeg decode. Returns None if decoding fails."""
         array = decode_jpeg_image(jpeg_bytes)
         if array is None:
             return None
         if len(array.shape) == 2:
             array = np.repeat(np.expand_dims(array, -1), repeats=3, axis=2)
-        return cls.from_array(array, camera_id=camera_id, time=time)
+        return cls.from_array(array, camera_id=camera_id, time=time, metadata=metadata)
 
     @classmethod
-    def from_array(cls, array: np.ndarray, *, camera_id: str = 'from_array', time: float | None = None) -> Self:
+    def from_array(cls, array: np.ndarray, *,
+                   camera_id: str = 'from_array',
+                   time: float | None = None,
+                   metadata: dict[str, Any] | None = None) -> Self:
         """Create an image from a NumPy array."""
 
         # Convert array to a c-contiguous representation (or do nothing if that
@@ -97,7 +110,7 @@ class Image:
         assert array.dtype == np.uint8, 'Array must have dtype np.uint8'
         assert len(array.shape) == 3, 'Array must have shape (height, width, channels)'
         assert array.shape[2] == 3, 'Image should have 3 channels'
-        return cls(camera_id=camera_id, time=time or rosys.time(), _array=array)
+        return cls(camera_id=camera_id, time=time or rosys.time(), _array=array, metadata=metadata or {})
 
     def to_pil(self) -> PIL.Image.Image:
         """Convert the image to a PIL image."""


rosys/vision/usb_camera/usb_device.py
```diff
@@ -98,6 +98,8 @@ async def _capture_image(self) -> None:
                     return
                 result = self._on_new_image_data(bytes_, timestamp)
             else:
+                # convert bgr to rgb
+                frame = cv2.cvtColor(frame, cv2.COLOR_BGR2RGB)
```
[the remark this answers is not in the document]
falkoschindler: GPT:

> **One caveat: not every downstream consumer accepts negative-stride arrays.**
> Some OpenCV functions choke on views that use negative strides, because they expect contiguous memory. If you later pass this array into some OpenCV calls, you might end up with hidden copies or errors. But the act of flipping the channels itself is still free.

Just something to keep in mind.

rosys/vision/detector_hardware.py
```diff
@@ -26,9 +26,9 @@ class DetectorHardware(Detector):
     It automatically connects and reconnects, submits and receives detections and sends images
     that should be uploaded to the [Zauberzeug Learning Loop](https://zauberzeug.com/products/learning-loop).
 
-    Note: Images must be smaller than ``MAX_IMAGE_SIZE`` bytes (default: 10 MB).
+    Note: Images must be smaller than ``MAX_IMAGE_SIZE`` bytes (default: 64 MB).
     """
-    MAX_IMAGE_SIZE = 10 * 1024 * 1024
+    MAX_IMAGE_SIZE = 64 * 1024 * 1024
```
NiklasNeugebauer: If I remember correctly, there is a limit set on the detector node side. Is this large enough to accommodate 64MB?

NiklasNeugebauer:
I am finished with my review for now

uv.lock
```diff
@@ -0,0 +1,3 @@
+version = 1
+revision = 3
+requires-python = ">=3.12"
```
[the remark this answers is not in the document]
ftilde: Whoops. I've removed it again.

rosys/vision/detector_hardware.py
NiklasNeugebauer: It is currently `2e7`, so  `20 MB`, or `20 * 1000 * 1000`.  
We could consider increasing it though if we expect images to be that large.

rosys/vision/image.py
```diff
@@ -27,19 +26,13 @@ def tuple(self) -> tuple[int, int]:
 @dataclass(slots=True, kw_only=True)
```
denniswittich: I am wondering if we should somehow document that images are not (no longer?) persistable as the array is neither a dataclass nor a simple python data type.. it can be converted to a dictionary but restoring it fails.

This may also be problematic when processing images using `cpu_bound` (even though it seems to work)?

@falkoschindler any thoughts on this?

denniswittich: I think dataclasses should not be usued for mutable data structures.
Another problem is that im1==im2 would no longer return true unless we write a custom eq_fn

rosys/vision/usb_camera/usb_device.py
```diff
@@ -98,6 +98,8 @@ async def _capture_image(self) -> None:
                     return
                 result = self._on_new_image_data(bytes_, timestamp)
             else:
+                # convert bgr to rgb
+                frame = frame[:,:,::-1]
```
denniswittich: FYI, this breaks contiguousity!! (cf. comment in `Image`)

rosys/vision/image.py
```diff
@@ -27,19 +26,13 @@ def tuple(self) -> tuple[int, int]:
...
-    tags: set[str] = field(default_factory=set)
+    metadata: dict[str, Any] = field(default_factory=dict)
 
     DEFAULT_PLACEHOLDER_SIZE: ClassVar[tuple[int, int]] = (320, 240)
-
-    def __post_init__(self) -> None:
-        if self.tags:
-            warnings.warn('The "tags" field is deprecated and will be removed in a future version.',
-                          DeprecationWarning, stacklevel=2)
+    array_type = np.ndarray[tuple[int, int, int], np.dtype[np.uint8]]  # pixel data with shape (height, width, RGB)
```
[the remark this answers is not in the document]
ftilde: I've updated the comment and changed the case to PascalCase (since it is a type).

rosys/vision/image.py
ftilde: For 1.: Once we have constructed the internal representation, we can create an array in "zero copy" fashion using [np.frombuffer](https://numpy.org/doc/stable/reference/generated/numpy.frombuffer.html). That's cheap. But for some reason the "inverse" operation [np.tobytes](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tobytes.html#numpy.ndarray.tobytes) copies the data.

For 2.: I wasn't aware that dataclass brings a notion of "cheap". I would not qualify an image as "cheap" anyways... So, I vote for option 2. 

Do we even want to define an equality between images? If so, how is it defined? Should the metadata be taken into account or not? Maybe we are only interested if we point to the same image object? If the user wants pixel equality (not cheap to compute), they can just compare `array`.

rosys/vision/camera/camera.py
```diff
@@ -27,7 +27,7 @@ def __init__(self,
                  streaming: bool | None = None,
                  polling_interval: float | None = None,
                  base_path_overwrite: str | None = None,
-                 image_history_length: int = 256,
+                 image_history_length: int = 16,
```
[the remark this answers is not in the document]
ftilde: That's also a possibility. And do we want to remove the non-static member (and constructor arguments) or do we want to keep them? In that case we would have `Camera.DEFAULT_IMAGE_HISTORY_LENGTH`, I think.

rosys/vision/usb_camera/usb_device.py
denniswittich: Just checked, calling `ascontiguousarray` would work and make it `C_CONTIGUOUS` . The function name `ascontiguousarray` is misleading.. it actually makes it `C_CONTIGUOUS`.

rosys/vision/calibration.py
```diff
@@ -526,14 +526,13 @@ def undistort_image(self, image: NDArray, *, crop: bool = False) -> NDArray:
...
+            img = Image.from_array(
                 camera_id=image.camera_id,
-                size=image.size,
                 time=image.time,
-                data=cv2.imencode('.jpg', self.undistort_image(image.to_array(), crop=crop))[1].tobytes(),
-                is_broken=image.is_broken,
-                tags=image.tags,
+                array=self.undistort_image(image.array, crop=crop),
             )
+            img.metadata = image.metadata
```
falkoschindler: Why don't we make `metadata` a parameter of `from_array`?

falkoschindler (commented):
Among some comments throughout the code I only have two more thoughts:

- I don't get why app_controls_.py and bluetooth.py are part of the diff. This is a bit worrying, but at least the change looks like the one from PR #364.

- The `Image.from_*` class methods seem a bit inconsistent: Some default to `time=0`, some to `time=rosys.time()`. But since `time=None` is also valid, why not simply forward `time` without modification?
    Same for `camera_id`: Defaulting to `camera_id='no_cam_id'` sounds paradox. And even if we want to fill a missing ID by `'no_cam_id'`, we should do it using a default value for the parameter. Otherwise you can't explicitly set `camera_id=None` from outside.

rosys/vision/rtsp_camera/rtsp_camera.py
```diff
@@ -89,19 +88,10 @@ async def disconnect(self) -> None:
...
-        transformed_image_bytes = await rosys.run.cpu_bound(process_jpeg_image, image_bytes, self.rotation, self.crop)
-        if transformed_image_bytes is None:
-            return
-
-        try:
-            final_image_resolution = get_image_size_from_bytes(transformed_image_bytes)
-        except ValueError:
-            return
+    async def _handle_new_image_data(self, image_array: ImageArray, timestamp: float) -> None:
+        transformed_image_array = process_ndarray_image(image_array, self.rotation, self.crop)
```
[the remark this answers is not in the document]
denniswittich: In most cases we do not apply any transformation, so moving to another process would be just overhead.

rosys/analysis/timelapse_recorder.py
```diff
@@ -160,20 +158,21 @@ def _save_image(image: RosysImage,
...
     frame_info = ', ' + frame_info if frame_info else ''
     _write(f'{datetime.fromtimestamp(image.time):%Y-%m-%d %H:%M:%S}' + frame_info, draw, x, y)
+
     for message in notifications:
         y += 30
         _write(message, draw, x, y)
     if overlay:
         style = 'position:absolute;top:0;left:0;pointer-events:none'
-        viewbox = f'0 0 {image.size.width} {image.size.height}'
+        viewbox = f'0 0 {image.array.shape[1]} {image.array.shape[0]}'
```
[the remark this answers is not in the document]
ftilde: I agree on the implementation detail. Looking into the history I think this diff is an artifact of me being confused about `Protocol`/`RosysImage`. I've modified `RosysImage` now so that we use `image.size.width` (and `height`) again. I feel like going over the ImageSize property is fine, but we could add direct readonly properties for `width`/`height`.

PR #357, opened 2025-10-30
Emit an event when the battery is below a threshold

pascalzauberzeug:
### Motivation & Implementation

We want to be notified when the battery of our robot falls below a set threshold. That's why this PR introduces the `BATTERY_LOW` event and the optional keyword argument `battery_low_threshold` for the BMS.

### Progress

### DIFF
--- a/rosys/hardware/bms.py
+++ b/rosys/hardware/bms.py
@@ -53,9 +53,10 @@ class Bms(Module, abc.ABC):
         elif not is_charging and self.state.is_charging:
             self.CHARGING_STOPPED.emit()
 
-    def _send_battery_low_event(self, current_percentage: float, previous_percentage: float) -> None:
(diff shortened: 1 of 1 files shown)


rosys/hardware/bms.py
```diff
@@ -18,15 +18,18 @@ class Bms(Module, abc.ABC):
     The BMS module provides measured voltages as an event.
     """
 
-    def __init__(self, **kwargs) -> None:
+    def __init__(self, *, battery_low_threshold: float | None = None, **kwargs) -> None:
```
falkoschindler: Intuitively I would have expected the threshold to be in volts, but according to the implementation it is in percent. This shows how important it is to document all parameters. 🫢

falkoschindler (commented):
I know I'm late to the party, but I added two comments to this PR.

PR #358, opened 2025-10-31
Randomize ARP scan order for eventual guaranteed device discovery

ftilde:
This should result in eventual discovery of all devices of the network, even if they are too underpowered to cope with the full flood of ARP requests.

### Motivation

I've had trouble to connect to the JoVision camera in the office via autoscan. Investigation showed that either a router in between or the device itself (likely the latter) does not respond to requests if they are "flooded", where the 255 requests for a /24 subnet already constitute flooding. It seems that the hardware (or network-stack) of the JoVision camera in the office is so underpowered that they only process the first 32 or so requests.

### Implementation

The arp scan works by sending individual requests/packets for all `2^(32-n) -1` ip addresses in a network, i.e. for a 192.168.7.1/24:
"What's the mac address to 192.168.7.1?"
"What's the mac address to 192.168.7.2?"
"What's the mac address to 192.168.7.3?"
...
"What's the mac address to 192.168.7.254?"
"What's the mac address to 192.168.7.255?"

One way to handle this would be to throttle the arp-scan by a delay between each packet, but I've found this to work only work reliably with a delay of ~100ms, which would result in a total scan duration of 25s for a /24 subnet. Additionally, other devices on the network may scan simultaneously and thus temporarily overload our target device so that even with throttling there is no guarantee that a scan discovers the camera. Finally, we would probably want to make this delay configurable which requires tuning for specific devices and networks.

Instead I propose that we use the `-R` option of `arp-scan` which randomizes the order in which requests are sent, i.e. for example:
"What's the mac address to 192.168.7.128?"
"What's the mac address to 192.168.7.5?"
"What's the mac address to 192.168.7.222?"
...
"What's the mac address to 192.168.7.1?"
"What's the mac address to 192.168.7.25?"

This does not guarantee that we find the camera on the first scan, but we should find it eventually, when we (by chance) ask for its mac-address in the first number of requests until it is overloaded. Also, we don't delay discovery of well-behaved devices.

Notably, in contrast to more sophisticated scheduling, this proposed change strictly improves the situation from the status quo, without any further configuration.

### Progress

### DIFF
--- a/rosys/vision/rtsp_camera/arp_scan.py
+++ b/rosys/vision/rtsp_camera/arp_scan.py
@@ -30,7 +30,17 @@ async def run_arp_scan(interface: str | None = None) -> str:
     if os.getuid() != 0:
         arpscan_cmd = f'sudo {arpscan_cmd}'
 
-    cmd = f'{arpscan_cmd} --localnet'
+    # NOTE: -R randomizes the order in which the mac addresses for the given
+    # localnet IP address range are requested. That means:
+    # Without -R: 192.168.7.1, 192.168.7.2, 192.168.7.3, ...
+    # With -R (e.g.): 192.168.7.193, 192.168.7.12, 192.168.7.7, ...
+    # If a device (camera) we are interested in has a "high" IP address in our
+    # range (and thus at the end of our list) and if it is unable to cope with
+    # the flood of ARP-requests (e.g. due to underpowered hardware) it may
+    # never respond.
+    # To avoid such a scenario, we add -R so that any device will eventually be
+    # at the start of the list and thus be able to respond.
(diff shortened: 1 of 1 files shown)


pascalzauberzeug:
@ftilde have you tried limiting the requests with `--interval` and `--retry`? https://linux.die.net/man/1/arp-scan

falkoschindler (approved):

ftilde:
@pascalzauberzeug I tried, but it needed ~100ms delay with `--interval` for it to work reliably. I don't think we would want that in all cases, so it would have to be configurable. We can still do that, but I would argue that `-R` is a strict improvement right now.
`--retry` seemed to make things worse, probably due to retries being sent for other devices and thus generating even more traffic for the weak camera to handle.

PR #360, opened 2025-11-03
Add a /logs page to download log files from the UI

Johannes-Thiel:
### Motivation
When using rosys on a remote device it can be cumbersome to download the `*.log` files using ssh. To make this easier I want to introduce a `/logs` page to make the logs downloadable using the ui.

### Implementation
I choose to take inspirtaion from  `video_page.py`. In a similar fashion we have a list of `*.log` and `*.log.*` files. These can be downloaded or inspected in the browser. This function could ahev been added to the `/logging` page. Instead I choose to make a seperate page, in order to prevent the `/logging` page from getting to complex.

### Progress

### DIFF
--- a/rosys/analysis/logs_page_.py
+++ b/rosys/analysis/logs_page_.py
@@ -1,7 +1,9 @@
 from datetime import datetime
 from pathlib import Path
 
-from nicegui import run, ui
+from fastapi import HTTPException
+from fastapi.responses import FileResponse
+from nicegui import app, run, ui
 
 
 class LogsPage:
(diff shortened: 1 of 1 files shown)


rosys/analysis/__init__.py
```diff
@@ -29,6 +30,8 @@
     'date_to_str',
     'kpi_page',
     'logging_page',
+    'logs_page',
+    'logs_page',
```
falkoschindler:
```suggestion
    'logs_page',
```

rosys/analysis/logs_page_.py
```diff
@@ -0,0 +1,83 @@
+from datetime import datetime
+from pathlib import Path
+
+from nicegui import run, ui
+
+LOG_FILES = Path('~/.rosys').expanduser()
```
pascalzauberzeug: I think this should be configurable

rosys/analysis/logs_page_.py
```diff
@@ -0,0 +1,74 @@
...
+    logs = await run.io_bound(logs_dir.glob, '*.log')
+    rotated = await run.io_bound(logs_dir.glob, '*.log.*')
+    paths = list({p.resolve(): p for p in [*logs, *rotated]}.values())
+
+    def mtime(p: Path) -> float:
+        try:
+            return p.stat().st_mtime
+        except FileNotFoundError:
+            return 0.0
+    return sorted(paths, key=mtime, reverse=True)
```
falkoschindler: I think this `mtime` function is overkill, because we just globed the files, so a `FileNotFoundError` seems impossible.

```suggestion
    return sorted(paths, key=lambda p: p.stat().st_mtime, reverse=True)
```

falkoschindler (approved):
Looks good to me. I just improved(?) some minor things:

- I introduced a new endpoint to serve individual log files. (It accepts arbitrary filenames, so the whole log directory is readable via HTTP. Is this ok?)
- The whole log item is clickable and opens the log file in a new browser tab.
- The download icon moved to the right. This feels more intuitive to me, since downloading is now a secondary action after viewing in the browser.
- I swapped mtime and size because the mtime has almost constant width, so everything aligns better this way. And mtime is the key for sorting the file list.
- I moved `_human_size` further down, because we prefer to have calling functions above called functions.

PR #361, opened 2025-11-04
Add methods to switch the RobotBrain's power control outputs

pascalzauberzeug:
### Motivation & Implementation

While cleaning up our robot's code, I removed these functions from an ui widget. I thought these may be helpful some time.

### Progress

    - [x] Check back with electronics to make sure to write correct docstrings

### DIFF
--- a/rosys/hardware/battery_control.py
+++ b/rosys/hardware/battery_control.py
@@ -28,14 +28,14 @@ class BatteryControlHardware(ModuleHardware):
     async def release_battery_relay(self) -> None:
         self.log.info('releasing battery relay')
         await rosys.sleep(1)
-        await self._relay_on()
+        await self.relay_on()
         await rosys.sleep(0.5)
-        await self._relay_off()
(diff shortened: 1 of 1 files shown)


rosys/hardware/battery_control.py
```diff
@@ -28,6 +28,14 @@ def handle_core_output(self, _: float, words: list[str]) -> None:
...
         await rosys.sleep(0.5)
+        await self._relay_off()
+
+    async def _relay_on(self) -> None:
+        """Switch on the relay that is used to cut charging power to the battery."""
+        await self.robot_brain.send(f'{self.name}_reset.on()')
+
+    async def _relay_off(self) -> None:
+        """Switch off the relay that is used to cut charging power to the battery."""
         await self.robot_brain.send(f'{self.name}_reset.off()')
```
falkoschindler: > I thought these may be helpful some time.

I wonder how these methods could be useful, since they are private. 🤔

PR #362, opened 2025-11-07
Remove second subscription to `NEW_IMAGE` event

denniswittich:
### Motivation

Some camera providers (MJPG, USB, Replay) subscribe twice to the NEW_IMAGE events of the cameras. This results in every image being emmited twice. 

### Implementation

Remove second subscription


falkoschindler (approved):
Ah, you mean their `restore_from_dict` calls `add_camera` (which already subscribes to  the event) and then `NEW_IMAGE.subscribe` which causes a second subscription.

I guess this is a regression due to PR #345. RoSys' event system ignored duplicate event registrations, but NiceGUI's `Event` doesn't. I wasn't aware that the duplication filter has actually been used.

PR #363, opened 2025-11-09
Improve camera ID conversion in Record/Replay module

denniswittich:
### Motivation

When using the recorder/replay modules, the restored cameras did not have the same camera ids like te ones used during recording. The reason for this is that we had to replace `:` by `-` to have a valid file path.

This is problematic, e.g. when camera ids are hard-coded in the application code. 

### Implementation

The idea of this PR is to revert the character encoding when cameras are created. However, as the symbol `-` may be used in the camera id, the sequence `--` is used to encode colons.

### Note

To completely solve this problem we would need to create a metadata file that stores the camera names. This approach could be beneficial as we could also store additional information about the cameras (e.g. the name). However, as this would require some further discussion I propose to do this simple modification first.

### DIFF
--- a/tests/test_recorder_replay.py
+++ b/tests/test_recorder_replay.py
@@ -26,7 +26,7 @@ async def test_record_images(recordings_dir: Path):
     await forward(seconds=1.1)
 
     # ASSERT
-    cam_dir = recordings_dir / 'cam-0'
+    cam_dir = recordings_dir / 'cam--0'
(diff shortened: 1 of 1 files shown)


NiklasNeugebauer (changes requested):
Please adapt the tests accordingly (with --).
Looks good otherwise

NiklasNeugebauer (approved):

PR #372, opened 2025-12-01
Fix Bumper and EStop event emit order and add arguments to EStop events

pascalzauberzeug:
### Motivation

I noticed that the events of `Bumper` and `EStop` were fired before the new values were written, which led to wrong values in my user code. Especially since the `EStop` events have no argument.

### Implementation

1. Write values before emitting events
2. Add the EStop's name as an argument for `ESTOP_TRIGGERED` and `ESTOP_RELEASED`
3. The software EStop is now handled exactly like the other EStops, but I kept the `is_soft_estop_active` property

⚠️ The order could change event handling in your user code
⚠️ Your event handling functions could fail, because the events will now provide an argument
⚠️ `EStop.pressed_estops: list[int]` was renamed and retyped to `EStop.active_estops: set[str]`
⚠️ `Bumper.active_bumpers: list[str]` retyped to `Bumper.active_bumpers: set[str]`

### Progress

### DIFF
--- a/rosys/hardware/bumper.py
+++ b/rosys/hardware/bumper.py
@@ -51,16 +51,16 @@ class BumperHardware(Bumper, ModuleHardware):
 
     def handle_core_output(self, time: float, words: list[str]) -> None:
         previous_active_bumpers = self.active_bumpers.copy()
-        states: dict[str, bool] = {name: words.pop(0) == 'true' for name in self.pins}
         self.active_bumpers.clear()
-        self.active_bumpers.update(name for name, active in states.items() if active)
+        self.active_bumpers.update(name for name in self.pins if words.pop(0) == 'true')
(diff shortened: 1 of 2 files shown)


pascalzauberzeug:
@falkoschindler I change it to _could fail_

rosys/hardware/bumper.py
```diff
@@ -50,16 +50,17 @@ def __init__(self, robot_brain: RobotBrain, *,
                          estop=estop)
 
     def handle_core_output(self, time: float, words: list[str]) -> None:
+        active_bumpers = list(self.active_bumpers)
```
falkoschindler: Why wrapping with `list()`?
And maybe `previous_active_bumpers` is a better name?

falkoschindler (approved):

PR #375, opened 2025-12-09
Install esptool via Lizard's requirements.txt

falkoschindler:
### Motivation

A Dependabot alert complained about ecdsa, a dependency of esptool. But it is strange that RoSys depends on esptool and never uses it directly.

### Implementation

This PR removes esptool from pyproject.toml and extends the installation of requirements after unpacking the Lizard release artefact. Instead of installing prompt-toolkit, we now install all requirements.txt.

### Progress

- [x] Pytests are not necessary.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,9 +33,7 @@ RUN CURL="curl -s https://api.github.com/repos/zauberzeug/lizard/releases" && \
     unzip *zip && \
     rm *zip && \
     ls -lha
-
-# for Lizard monitor
-RUN pip install --no-cache prompt-toolkit
+RUN pip install --no-cache -r requirements.txt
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -33,9 +33,7 @@ RUN CURL="curl -s https://api.github.com/repos/zauberzeug/lizard/releases" && \
     unzip *zip && \
     rm *zip && \
     ls -lha
-
-# for Lizard monitor
-RUN pip install --no-cache prompt-toolkit
+RUN pip install --no-cache -r requirements.txt
```
pascalzauberzeug: I think `--no-cache` is weird, but we can keep it like this.

PR #377, opened 2025-12-18
Lower RtspDevice childprocess pipe size limit to default

ftilde:
### Motivation
The only effect of the larger limit is visibile in situations where we cannot keep up with the frames from there camera.
Then we fail to apply backpressure, leading to the gstreamer process stuffing the process pipe with old images. This is undesirable because 1) we want newer images and 2) this eats up memory.
Notably, readexact on the stream (which we use) buffers beyond the limit to fulfill our request anyway.

### Implementation

We remove the argument to apply the default (64KB).

### Progress


NiklasNeugebauer (approved):
I cannot verify this at the moment, but this sounds reasonable.

PR #383, opened 2026-01-21
Let Odometer extend PoseProvider and rework naming

pascalzauberzeug:
### Motivation

We already have the `PoseProvider` protocol but the Odometer, which is basically provides a pose, currently does not extend `PoseProvider`.

### Implementation

Now, `Odometer` does extend from `PoseProvider`.
For easier use I moved the protocols `PoseProvider` and `VelocityProvider` to their own files.

⚠️ `Driver.odometer` was renamed to `Driver.pose_provider` and `Driver.prediction` to `Driver.pose`.

### Progress

    - [x] Rename `prediction` to `pose` in `Driver`?
    - [x] Write Breaking Change warning

### DIFF
--- a/rosys/driving/driver_object.py
+++ b/rosys/driving/driver_object.py
@@ -30,7 +30,7 @@ class DriverObject(Group):
         ui.timer(config.ui_update_interval, self.update)
 
     def update(self) -> None:
-        robot_pose = self.driver.prediction
+        robot_pose = self.driver.pose
         self.robot_group.move(robot_pose.x, robot_pose.y).rotate(0, 0, robot_pose.yaw)
 
         if self.driver.state:


rosys/driving/driver.py
```diff
@@ -243,12 +236,12 @@ def _throttle(self, linear: float, angular: float) -> tuple[float, float]:
...
-        if not isinstance(self.odometer, Odometer):
+        if not isinstance(self.pose_provider, Odometer):
             raise ValueError('max_detection_age_ramp requires an odometer')
-        if self.odometer.detection is None:
+        if self.pose_provider.detection is None:
             return 0
         age_ramp = self.parameters.max_detection_age_ramp
-        age = rosys.time() - self.odometer.detection.time
+        age = rosys.time() - self.pose_provider.detection.time
         return ramp(age, age_ramp[0], age_ramp[1], 1.0, 0.0, clip=True)
```
pascalzauberzeug: @falkoschindler Can we rewrite that logic to avoid needing Odometer.detection? I would like to avoid needing the pose provider to have a detection attribute.

PR #385, opened 2026-02-02
Integrate automatic camera reconnection into Camera base class

NiklasNeugebauer:
### Motivation

Cameras can lose connection due to network or power issues. Currently, when a camera disconnects, it requires another camera scan through the \`CameraProvider\` to be reconnected.

This PR adds automatic reconnection behavior directly to the \`Camera\` base class. One can then set \`auto_scan=False\` in the camera provider and scan only when necessary (e.g. on user input), but still keep any previously connected cameras connected.

This is less heavy than a full scan and avoids finding potential new cameras that we currently do not want connected (happens in test/development settings).

### Implementation

Every \`Camera\` now has:
- A \`reconnect_interval\` parameter (default 5.0 seconds) controlling how often reconnection is attempted
- An \`activate()\` method that connects and starts a background \`Repeater\` that periodically calls \`connect()\` when the camera is disconnected
- A \`deactivate()\` method that stops reconnection and disconnects
- An \`is_reconnecting\` property indicating whether the reconnection repeater is running
- A mutex ensuring \`deactivate()\` and the reconnect attempt cannot race

Usage:

\`\`\`python
camera = RtspCamera(id='my_camera', ip='192.168.1.100')
await camera.activate()   # connect + start auto-reconnect
await camera.deactivate() # stop reconnect + disconnect
\`\`\`

### Progress


NiklasNeugebauer:
@falkoschindler @pascalzauberzeug Do any of you care to comment on this PR?

falkoschindler (changes requested):
Hm, Opus 4.6 found several issues:

---

### Overview

This PR adds a `ReconnectingCamera` mixin that automatically reconnects cameras when they lose connection. It uses the existing `Repeater` infrastructure to periodically check connection status and retry `connect()`. The MRO-based mixin pattern (`class MyCamera(ReconnectingCamera, ConcreteCamera)`) is clean and consistent with Python conventions.

---

### Issues & Suggestions

#### 1. Missing serialization for `reconnect_interval`

The base `Camera` class and all subclasses (`MjpegCamera`, `RtspCamera`, etc.) override `to_dict()` to persist their configuration. `ReconnectingCamera` does not, which means `reconnect_interval` will be lost on save/restore. If a user creates a `ReconnectingRtspCamera` with `reconnect_interval=10.0` and the camera provider persists it, it will silently revert to the default `5.0` on reload.

**Suggestion:** Add `to_dict` override:
```python
def to_dict(self) -> dict:
    return super().to_dict() | {'reconnect_interval': self.reconnect_interval}
```

(`args_from_dict` doesn't need an override since the base implementation passes `data` through as kwargs, and `reconnect_interval` will be picked up by `__init__`.)

#### 2. Mutating `connect_after_init` as a side effect of `connect()`

reconnecting_camera.py:43:
```python
self.connect_after_init = True
```

This silently overwrites the user's original `connect_after_init=False` setting. It also changes serialization output — calling `to_dict()` after `connect()` will produce a different value than before. The PR description acknowledges this edge case but dismisses it. I think this line should at minimum be documented with a comment explaining *why* it's there, or better yet, be removed if the reconnect task itself handles reconnection on restart (which it does).

**Question for author:** What scenario does this line protect against? If the purpose is to ensure the camera reconnects after a restart/deserialize, wouldn't the `is_activated` / `_reconnect_task` state already handle that? The current approach changes observable state as a side-effect.

#### 3. `__del__` is unreliable for cleanup

reconnecting_camera.py:63-64:
```python
def __del__(self) -> None:
    if self._reconnect_task:
        self._reconnect_task.stop()
```

`__del__` is not guaranteed to be called in CPython (circular references, interpreter shutdown), and referencing `self._reconnect_task` in `__del__` can fail if the instance wasn't fully initialized. Since `Repeater.tasks` is a class-level set that's already cleaned up on shutdown (and `disconnect()` handles the normal cleanup path), this `__del__` provides a false sense of safety.

**Suggestion:** Remove `__del__` or convert to a comment explaining that cleanup is handled by `disconnect()` and system shutdown.

#### 4. `_try_reconnect` should log camera ID on failure

reconnecting_camera.py:58:
```python
log.error('reconnection attempt failed: %s', e)
```

This doesn't include the camera ID, unlike the other log messages (`log.warning('[%s] connect failed: %s', self.id, e)`). With multiple cameras, this makes it hard to diagnose which camera is failing.

**Fix:**
```python
log.error('[%s] reconnection attempt failed: %s', self.id, e)
```

#### 5. Race condition between `_try_reconnect` and `disconnect`

If `disconnect()` is called while `_try_reconnect()` is mid-execution (after the `is_connected` check but before `super().connect()` completes), the sequence would be:

1. `_try_reconnect` checks `is_connected` → False
2. `disconnect()` runs: stops task, sets `_reconnect_task = None`, calls `super().disconnect()`
3. `_try_reconnect` calls `super().connect()` → camera reconnects unexpectedly

Since `_try_reconnect` uses `await super().connect()` (not `self.connect()`), it won't restart the repeater, but the camera will end up connected with no reconnect task — an inconsistent state. The commit message for 46e0428f mentions this was considered, but the fix only prevents the task from restarting, not the connect itself.

**Suggestion:** Use the existing `device_connection_lock` in `_try_reconnect`:
```python
async def _try_reconnect(self) -> None:
    async with self._device_connection():
        if not self.is_connected:
            ...
```

#### 6. `.gitignore` change should use `.venv/` not `.venv/*`

The `.venv/*` pattern technically works but is unconventional. `.venv/` is the standard gitignore pattern for directories and is more consistent with the rest of the file (which uses `site/`, `.nicegui/`, etc.).

---

### Minor Nits

- The `log.debug` at line 42 fires even if the task was already running (idempotent but slightly noisy).
- Tests are solid and cover the key scenarios. Consider adding a test for the race condition (disconnect during active reconnect attempt).

---

### What Looks Good

- Clean mixin design using Python MRO — `super().connect()` correctly delegates through the chain.
- Good use of existing `Repeater` infrastructure instead of hand-rolling async loops.
- The fix in commit 46e0428f to use `super().connect()` in `_try_reconnect` (avoiding re-triggering the mixin's `connect()`) shows careful thought about the dispatch chain.
- Documentation and tests are included and well-structured.
- CI is green across all Python versions (3.11, 3.12, 3.13).

---

### Summary

The overall design is sound and addresses a real need. The main areas to address are the **missing serialization** for `reconnect_interval`, the **inconsistent log message**, and the **race condition** between disconnect and reconnect. The `connect_after_init` mutation and `__del__` usage are worth discussing but less critical.

---

And another nitpick from myself: Double colon at `Example usage::`. Is this intentional?

NiklasNeugebauer:
I fixed all issues except for `Mutating connect_after_init as a side effect of connect()`.

It feels natural for a camera to keep its active state after a restart but it might depend on the application.

NiklasNeugebauer:
Other question: should I
A. keep this as an optional mixin
B. add this to all cameras
C. integrate it into `Camera`

NiklasNeugebauer:
@falkoschindler

falkoschindler:
@NiklasNeugebauer Hm, I'd follow the recommendation of Opus 4.6:

---

Good question. Here's my take on the three options:

**A. Optional mixin (current)** — Forces boilerplate for every camera type (`class ReconnectingRtspCamera(ReconnectingCamera, RtspCamera)`) and relies on users getting MRO order right. But it's cleanly separated.

**B. Provide reconnecting variants of all cameras** — Combinatorial explosion, clearly not worth it.

**C. Integrate into Camera with opt-in flag** — I'd lean toward this.

The reconnection logic is simple enough (a repeater + a lock + one float parameter) that it doesn't meaningfully bloat `Camera`. You could make it opt-in with a single parameter:

```python
class Camera:
    def __init__(self, *, reconnect_interval: float | None = None, ...):
```

Where `None` means no auto-reconnect (current default behavior), and any float value enables it. The repeater and lock only get created when `reconnect_interval` is set.

**Why I'd prefer C:**

- No MRO gymnastics — users don't need to understand cooperative MI or remember the class order
- Works with `from_dict` / `to_dict` naturally since it's just another Camera parameter
- In robotics, physical cameras losing connection is the norm, not the exception — this feels like core camera behavior, not an optional extra
- The PR author already used it in two projects, which suggests it'll be needed broadly
- Avoids the `device_connection_lock` breaking change since it stays on Camera

**The one argument for A** is if you want the Camera base class to stay minimal and purely abstract. But given Camera already has `connect_after_init`, image history, routing, and serialization, it's already beyond a minimal interface.

So: **C**, with `reconnect_interval: float | None = None` as the opt-in mechanism.

NiklasNeugebauer:
## Review of current state

### Issues resolved from the original review
- Serialization of `reconnect_interval` in `to_dict()` ✓
- `__del__` removed ✓
- Camera ID added to error log ✓
- Race condition fixed with `_device_connection_lock` mutex ✓

---

### Remaining issues

#### 1. `connect_after_init = True` side effect in `activate()` (line 136)

```python
async def activate(self) -> None:
    ...
    self.connect_after_init = True  # ← still here
```

This silently overwrites a user's explicit `connect_after_init=False` and changes serialization output mid-lifecycle. The reconnect repeater already handles reconnection — this line doesn't add anything useful. The repeater isn't persisted, so after a restart the user must call `activate()` again regardless. **This line should just be removed.**

#### 2. `_device_connection()` is unnecessary indirection

```python
@asynccontextmanager
async def _device_connection(self):
    async with self._device_connection_lock:
        yield
```

This is a one-liner wrapper used in exactly 2 places. It adds a layer of indirection without benefit. Both callers could just use `async with self._device_connection_lock:` directly. The context manager was perhaps a holdover from when the lock was a `Condition` and needed more ceremony — now it's just noise.

#### 3. `.venv/*` in `.gitignore` should be `.venv/`

Minor but mentioned in the review and still not fixed. Every other entry in that file uses the directory pattern (`site/`, `.nicegui/`, etc.).

#### 4. PR description is outdated

The body still describes the `ReconnectingCamera` mixin design (with `class MyReconnectingCamera(ReconnectingCamera, MyCamera)` usage example), but the implementation was refactored to integrate directly into `Camera`. The description should be updated to reflect the current approach (`activate()`/`deactivate()` on base `Camera`).

#### 5. Minor: debug log fires even when repeater already running

```python
if not self._reconnect_repeater:
    self._reconnect_repeater = rosys.on_repeat(...)
if not self._reconnect_repeater.running:
    self._reconnect_repeater.start()
logger.debug('[%s] reconnect repeater started', self.id)  # fires regardless
```

The log message says "started" but fires even on an idempotent second call where nothing happened. Move it into the `if not self._reconnect_repeater.running:` block.

---

**Priority:** Items 1 and 2 are real code issues. Item 4 matters for review clarity. Items 3 and 5 are polish.

NiklasNeugebauer:
This needs more thought. Integration means we cannot nicely wrap the connect behavior and would need a separate `activate` method.

NiklasNeugebauer:

**Verdict: recommend closing in favor of #423** (the newer alternative by the same author). Beyond the architectural argument in #423's description, concrete problems here:

1. `on_startup(self.activate)` changes behavior for *every* camera subclass — all cameras silently gain a reconnect repeater.
2. `activate()` mutates `connect_after_init` as a side effect, which leaks into persistence (`to_dict`) — surprising for a camera constructed with `connect_after_init=False`.
3. `activate()` calls `connect()` *without* holding `_device_connection_lock`, while `deactivate()`/`_try_reconnect` do hold it — activate/deactivate can race.
4. Removing the public `device_connection_lock` `Condition` and the `_device_connection` context manager is a breaking change for out-of-tree subclasses.
5. The branch is 6 months stale (base 2026-01-09, `mergeable_state: unstable`) and would need a full rebase anyway.

The 91-line test file is worth salvaging as inspiration for #423's suite.

NiklasNeugebauer:
closed in favor of #423

PR #387, opened 2026-02-16
Mac field for rtspcamera

### LINKED ISSUE #386: RtspDevice assumes RtspCamera's id to be the mac
The `RtspDevice` requires the mac address to resolve the vendor.  When we initialize the `RtspDevice` within a `RtspCamera` (rtsp_camera.py:74) we set `mac=self.id` (where self is the `RtspCamera`-Instance). Thus, we assume that the field `id` of `RtspDevice` corresponds to the mac (or at least starts with it). Practically, in a project we require two cameras that use main and substream respectively for the same camera (same mac). Thus, there the mac is not a unique identifier.

I suggest to differentiate `id` from `mac` and have both attributes for `RtspCamera`.
I also mention that using `id` as field is considered to be problematic by type checkers as it overwrites the function `id()` - we could improve the variable name when we touch this anyways.

denniswittich:
### Motivation

RtspCamera used id as the MAC address, preventing two cameras from sharing the same MAC (e.g., main stream and substream of the same physical device). Closes [#386](https://github.com/zauberzeug/rosys/issues/386).

### Implementation

  - Add required mac parameter to RtspCamera, making id optional (auto-generated as {mac}-{substream})
  - Pass self.mac instead of self.id to RtspDevice
  - Update RtspCameraProvider to look up cameras by mac field
  - Add backward compatibility: old persisted data without mac infers it from id

### Breaking Change

`RtspCamera` has now a mandatory field `mac` and an optional field `id`. If the `id` is not set in the initializer, it is set to`"{mac}-{substream}" (before, the `id` was identical to the `mac`).

Migration:
- When accessing RTSP cameras by their `id`, one has to compare to "{mac}-{substream}" instead of "{mac}". 
- Note: For old persistence data without `mac`, rosys infers it from `id`, thus nothing to do for RTSP cameras created by the provider! However, when the `id` was modified, one needs to adjust or delete the persistence.

### Progress

### DIFF
--- a/rosys/vision/rtsp_camera/rtsp_camera.py
+++ b/rosys/vision/rtsp_camera/rtsp_camera.py
@@ -53,6 +53,13 @@ class RtspCamera(ConfigurableCamera, TransformableCamera):
         }
 
     @classmethod
+    def args_from_dict(cls, data: dict[str, Any]) -> dict:
+        data = super().args_from_dict(data)
+        if 'mac' not in data:
+            data['mac'] = data['id']
+        return data
+
+    @classmethod
     def from_dict(cls, data: dict[str, Any]) -> Self:
         return cls(**cls.args_from_dict(data))
 
--- a/rosys/vision/rtsp_camera/rtsp_camera.py
+++ b/rosys/vision/rtsp_camera/rtsp_camera.py
@@ -12,7 +12,8 @@ class RtspCamera(ConfigurableCamera, TransformableCamera):
 
     def __init__(self,
                  *,
-                 id: str,  # pylint: disable=redefined-builtin
+                 mac: str,
+                 id: str | None = None,  # pylint: disable=redefined-builtin
                  name: str | None = None,
                  connect_after_init: bool = True,
                  fps: int = 5,
@@ -21,6 +22,9 @@ class RtspCamera(ConfigurableCamera, TransformableCamera):
                  avdec: Literal['h264', 'h265'] = 'h264',
                  ip: str | None = None,
                  **kwargs) -> None:
+        self.mac = mac
+        if id is None:
+            id = f'{mac}-{substream}'
         super().__init__(id=id,
                          name=name,
                          connect_after_init=connect_after_init,
@@ -44,6 +48,7 @@ class RtspCamera(ConfigurableCamera, TransformableCamera):
             name: param.value for name, param in self._parameters.items()
         }
         return super().to_dict() | parameters | {
+            'mac': self.mac,
             'ip': self.ip,
         }
 
@@ -71,7 +76,7 @@ class RtspCamera(ConfigurableCamera, TransformableCamera):
             self.log.error('no IP address provided for camera %s', self.id)
             return
 
-        self.device = RtspDevice(mac=self.id, ip=self.ip,
+        self.device = RtspDevice(mac=self.mac, ip=self.ip,
                                  substream=self.parameters['substream'],
                                  fps=self.parameters['fps'],
                                  on_new_image_data=self._handle_new_image_data,
--- a/rosys/vision/rtsp_camera/rtsp_camera.py
+++ b/rosys/vision/rtsp_camera/rtsp_camera.py
@@ -23,9 +23,7 @@ class RtspCamera(ConfigurableCamera, TransformableCamera):
                  ip: str | None = None,
                  **kwargs) -> None:
         self.mac = mac
-        if id is None:
-            id = f'{mac}-{substream}'
-        super().__init__(id=id,
+        super().__init__(id=id or f'{mac}-{substream}',
                          name=name,
                          connect_after_init=connect_after_init,
                          **kwargs)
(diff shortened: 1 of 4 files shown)


rosys/vision/rtsp_camera/rtsp_camera.py
```diff
@@ -21,6 +22,9 @@ def __init__(self,
                  avdec: Literal['h264', 'h265'] = 'h264',
                  ip: str | None = None,
                  **kwargs) -> None:
+        self.mac = mac
+        if id is None:
+            id = f'{mac}-{substream}'
         super().__init__(id=id,
```
falkoschindler: Maybe a matter of taste, but could be shortened:
```suggestion
        super().__init__(id=id or f'{mac}-{substream}',
```

rosys/vision/rtsp_camera/rtsp_camera_provider.py
```diff
@@ -48,14 +48,15 @@ async def scan_for_cameras(network_interface: str | None = None) -> list[tuple[s
...
+            if camera is None:
                 self.log.debug('found new camera %s', mac)
-                self.add_camera(RtspCamera(id=mac,
+                self.add_camera(RtspCamera(mac=mac,
                                            fps=self.frame_rate,
                                            substream=self.substream,
                                            avdec=self.avdec,
                                            ip=ip))
-            camera = self._cameras[mac]
+                camera = next(c for c in self._cameras.values() if c.mac == mac)
```
falkoschindler: Since we just created the camera, we don't have to search for it. Just keep a reference:
```suggestion
                camera = RtspCamera(mac=mac, fps=self.frame_rate, substream=self.substream, avdec=self.avdec, ip=ip)
                self.add_camera(camera)
```

tests/test_camera_persistence.py
```diff
@@ -88,13 +88,36 @@ async def test_storing_mjpeg_camera_as_dict(rosys_integration, base_camera_param
...
+    camera = RtspCamera(mac='aa:bb:cc:dd:ee:ff', id='my-custom-id', connect_after_init=False)
+    assert camera.id == 'my-custom-id'
+    assert camera.mac == 'aa:bb:cc:dd:ee:ff'
+
+
+def test_rtsp_camera_backward_compat(rosys_integration):
+    old_data = {'id': 'aa:bb:cc:dd:ee:ff', 'substream': 1, 'fps': 5,
+                'bitrate': 4096, 'avdec': 'h264', 'ip': '192.168.1.1',
+                'connect_after_init': False}
+    camera = RtspCamera.from_dict(old_data)
```
falkoschindler: I would inline the dictionary, but expand it to one item per line for readability:
```suggestion
    camera = RtspCamera.from_dict({
        'id': 'aa:bb:cc:dd:ee:ff',
        'substream': 1,
        'fps': 5,
        'bitrate': 4096,
        'avdec': 'h264',
        'ip': '192.168.1.1',
        'connect_after_init': False,
    })
```

tests/test_camera_persistence.py
```diff
@@ -137,7 +160,10 @@ class CalibratableSimulatedCamera(CalibratableCamera, SimulatedCamera):
         pass
 
     for camera_class in [CalibratableRtspCamera, CalibratableUsbCamera, CalibratableMjpegCamera, CalibratableSimulatedCamera]:
-        camera = camera_class(id='test_cam')
+        kwargs: dict = {'id': 'test_cam'}
```
falkoschindler: I don't think the type annotation is necessary:
```suggestion
        kwargs = {'id': 'test_cam'}
```

falkoschindler (changes requested):
Opus 4.6 found two more issues to think about:

---

#### `update_device_list` doesn't handle multiple cameras per MAC (medium)

The stated motivation is supporting main stream + substream from the same physical device (same MAC). However, `update_device_list` uses `next(...if c.mac == mac...)` which only finds the **first** camera with that MAC. If a user has both substream 0 and substream 1 cameras for the same MAC, only the first one found will get its IP updated and connected. The second will be silently skipped.

This may be intentional if auto-discovery only creates one camera per MAC and additional substreams are added manually, but it's worth documenting or handling explicitly.

#### `RtspCameraProvider.backup_to_dict` overrides base class unnecessarily

[rtsp_camera_provider.py:33-37](rosys/vision/rtsp_camera/rtsp_camera_provider.py#L33-L37) duplicates the exact logic of the base `CameraProvider.backup_to_dict`. This isn't introduced by the PR, but now that the id/mac distinction exists, the `restore_from_dict` override (which uses `RtspCamera.from_dict`) becomes more important. The `backup_to_dict` override here should either be removed (it's identical to the base) or have a comment explaining why it exists.

---

And the PR description needs a word about the breaking change which we can copy&paste into the release notes. Thanks!

denniswittich:
> #### `update_device_list` doesn't handle multiple cameras per MAC (medium)
> The stated motivation is supporting main stream + substream from the same physical device (same MAC). However, `update_device_list` uses `next(...if c.mac == mac...)` which only finds the **first** camera with that MAC. If a user has both substream 0 and substream 1 cameras for the same MAC, only the first one found will get its IP updated and connected. The second will be silently skipped.

Interesting.. 
the `c.mac==mac` thing was suggested by Opus 4.6. I stumbled across this and asked it to include the substream because the mac is not unique. However, I then realized, that this basic `RtspCameraProvider` uses a fixed substream, thus, the mac will be a unique identifier - so yes, it was intentional. Long story short, with your suggested modification, this line was removed anyway.

NiklasNeugebauer (approved):

falkoschindler (approved):

PR #388, opened 2026-02-17
Add goodcam to vendor lists

NiklasNeugebauer:
### Motivation
We want to support a new camera manufacturer called GoodCam. This adds the interface both for `RtspCamera` and `MjpegCamera`

### Implementation
- add `GOODCAM` as a new vendor type in both `rosys/vision/mjpeg_camera/vendors.py` and `rosys/vision/rtsp_camera/vendors.py`
- add vendor-specific URL handling for Goodcam in `mac_to_url`
- update `mjpeg_device_factory.py` to return a `MjpegDevice` instance for Goodcam devices (no vendor-specific subclass needed since Goodcam uses standard MJPEG streaming)

### Progress


rosys/vision/mjpeg_camera/generic_mjpeg_device.py
pascalzauberzeug, line 1: I don't get the need of `GenericMjpegDevice`. Why can't we just use `MjpegDevice` since it's not abstract

NiklasNeugebauer: you're right, I had one with settings at first, but there were issues, so i reverted it and missed that the base one has the full implementation already.

pascalzauberzeug (approved):

PR #390, opened 2026-03-12
add to dict for imu and gnss data

felixwrz:
### Motivation
`ImuMeasurement` and `GnssMeasurement` lacked serialization support, making it impossible to save their data to JSON (e.g. for logging or data export). This PR adds `to_dict()` methods to both classes.

### Implementation
Added `to_dict()` to `ImuMeasurement` (`rosys/hardware/imu.py`) and `GnssMeasurement` (`rosys/hardware/gnss/gnss.py`) that serialize all relevant fields to plain Python dicts, which are directly JSON-serializable. 

### Progress

### DIFF
--- a/rosys/hardware/gnss/gnss.py
+++ b/rosys/hardware/gnss/gnss.py
@@ -40,6 +40,22 @@ class GnssMeasurement:
     def age(self) -> float:
         return ((self.gnss_time - rosys.time() + SECONDS_HALF_DAY) % SECONDS_DAY) - SECONDS_HALF_DAY
 
+    def to_dict(self) -> dict:
+        return {
+            'time': self.time,
+            'gnss_time': self.gnss_time,
+            'age': self.age,
+            'lat': self.pose.lat,
+            'lon': self.pose.lon,
+            'heading': self.pose.heading,
+            'quality': self.gps_quality.name.lower(),
+            'num_satellites': self.num_satellites,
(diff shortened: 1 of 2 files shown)


pascalzauberzeug (commented):
We could just use `from_dict` and `to_dict` from `rosys.persistence.converters`

falkoschindler:
The purpose of this PR is not 100% clear. The methods `from_dict` and `to_dict` are usually used for serialization/deserialization in the context of persistent storage. But despite their name, these methods are not suited for this purpose:

1. `GnssMeasurement.to_dict()` includes the computed `age` which changes with `rosys.time()`.
2. `ImuMeasurement.to_dict()` includes computed degrees for some angles but not for all.
3. There are no `from_dict()` counterparts.

Last but not least, there is a missing blank line before `class Gnss`.

If the main goal is to export structured data for a specific application, it might be better to do it in user code where we can care about project-specific requirements.
If, on the other hand, we want to provide complete serialization/deserialization of RoSys dataclasses (my recommendation), we should add `from_dict` counterparts and omit computed properties. (Roll, pitch and yaw might be fine though because a `from_euler` counterpart exists. Quaternions, however, would be numerically more stable.)

felixwrz:
Closing this. We decided to use `rosys.persistence.to_dict()` for serialization instead, so this change is no longer needed.

PR #391, opened 2026-03-20
Introduce FrameProvider protocol, add POSE_UPDATED event to PoseProvider, and rename Odometer prediction API

pascalzauberzeug:
### Motivation

Different projects use different pose estimators (`Odometer`, a Kalman filter, or a camera coordinate frame) that all provide a moving `Frame3d`.
A shared `FrameProvider` protocol lets consuming code accept any of them without coupling to a concrete class.
While at it, the Odometer API is cleaned up: the `prediction` / `prediction_frame` indirection is replaced with private state behind `pose` and `frame` properties.

This expands on #383.

### Implementation

- Add `FrameProvider` protocol to `rosys.geometry` with `FRAME_UPDATED: Event` and `frame: Frame3d`.
- Extend `PoseProvider` with `POSE_UPDATED: Event`.
- Make `Odometer` implement both protocols; rename `prediction` to `_pose`, `prediction_frame` to `_frame`, `PREDICTION_UPDATED` to `POSE_UPDATED`; add `FRAME_UPDATED`.
- Call `_handle_movement()` in `reset()` so subscribers are notified on reset.

### Breaking Changes

| Before | After |
|---|---|
| `odometer.prediction` | `odometer.pose` |
| `odometer.prediction_frame` | `odometer.frame` |
| `odometer.PREDICTION_UPDATED` | `odometer.POSE_UPDATED` |

The frame registry key changes from `rosys.odometer.prediction` to `rosys.odometer`.

### Progress

### DIFF
--- a/rosys/driving/odometer.py
+++ b/rosys/driving/odometer.py
@@ -29,7 +29,7 @@ class Odometer(PoseProvider):
 
         wheels.VELOCITY_MEASURED.subscribe(self.handle_velocities)
         self.prediction: Pose = Pose()
-        self.prediction_frame = Pose3d().as_frame('rosys.odometer.prediction')
+        self.frame = Pose3d().as_frame('rosys.odometer.prediction')
         self.detection: Pose | None = None
         self.current_velocity: Velocity | None = None
         self.last_movement: float = 0
@@ -76,9 +76,9 @@ class Odometer(PoseProvider):
         self._handle_movement()
 
     def _handle_movement(self) -> None:
-        self.prediction_frame.x = self.prediction.x
-        self.prediction_frame.y = self.prediction.y
-        self.prediction_frame.rotation = Rotation.from_euler(0, 0, self.prediction.yaw)
+        self.frame.x = self.prediction.x
+        self.frame.y = self.prediction.y
+        self.frame.rotation = Rotation.from_euler(0, 0, self.prediction.yaw)
         self.PREDICTION_UPDATED.emit()
 
     def prune_history(self, max_age: float = 10.0) -> None:
--- a/rosys/driving/odometer.py
+++ b/rosys/driving/odometer.py
@@ -23,10 +23,10 @@ class Odometer(PoseProvider, FrameProvider):
         """the wheels have turned with non-zero velocity"""
 
         self.POSE_UPDATED = Event[Pose]()
-        """Emitted on movement or detection correction."""
+        """Emitted on movement or detection correction (argument: current ``Pose``)."""
 
         self.FRAME_UPDATED = Event[Frame3d]()
-        """Emitted on movement or detection correction."""
+        """Emitted on movement or detection correction (argument: current ``Frame3d``)."""
 
         self.log = logging.getLogger('rosys.odometer')
 
--- a/rosys/driving/odometer.py
+++ b/rosys/driving/odometer.py
@@ -4,12 +4,12 @@ from copy import deepcopy
 from nicegui import Event
 
 from .. import rosys
-from ..geometry import Pose, Pose3d, PoseStep, Rotation, Velocity
+from ..geometry import FrameProvider, Pose, Pose3d, PoseStep, Rotation, Velocity
 from .pose_provider import PoseProvider
 from .velocity_provider import VelocityProvider
 
 
-class Odometer(PoseProvider):
+class Odometer(PoseProvider, FrameProvider):
     """An odometer collects velocity information from a given wheels module (or any velocity-providing hardware representation).
 
     It can also handle "detections", i.e. absolute pose information with timestamps.
@@ -22,13 +22,16 @@ class Odometer(PoseProvider):
         self.WHEELS_TURNED = Event[[]]()
         """the wheels have turned with non-zero velocity"""
 
-        self.PREDICTION_UPDATED = Event[[]]()
+        self.POSE_UPDATED = Event[[]]()
         """the pose prediction has been updated"""
+
+        self.FRAME_UPDATED = Event[[]]()
+        """the frame has been updated"""
 
         self.log = logging.getLogger('rosys.odometer')
 
         wheels.VELOCITY_MEASURED.subscribe(self.handle_velocities)
-        self.prediction: Pose = Pose()
+        self.pose: Pose = Pose()
         self.frame = Pose3d().as_frame('rosys.odometer.prediction')
         self.detection: Pose | None = None
         self.current_velocity: Velocity | None = None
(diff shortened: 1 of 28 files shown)


rosys/driving/pose_provider.py
```diff
@@ -1,10 +1,10 @@
...
 from ..geometry import Pose
 
 
 class PoseProvider(Protocol):
-
-    @property
-    def pose(self) -> Pose:
-        ...
+    POSE_UPDATED: Event
+    pose: Pose
```
falkoschindler: The new protocol is stricter than before: The class must implement a read-write property to satisfy the protocol. Is this intentional?

rosys/geometry/pose3d.py
```diff
@@ -99,6 +99,12 @@ def ancestors(self) -> Iterator[Frame3d]:
             yield from frame_registry[self.frame_id].ancestors
 
 
+class FrameProvider(Protocol):
+    """Protocol for objects that provide a moving Frame3d reference frame."""
+    FRAME_UPDATED: Event
+    frame: Frame3d
```
falkoschindler: Same here: `frame` needs to be settable. Is this intentional?

rosys/driving/odometer.py
falkoschindler, line 41: Now the `self.odometry_frame` is a very confusing field...

rosys/driving/odometer.py
```diff
@@ -22,14 +22,17 @@ def __init__(self, wheels: VelocityProvider) -> None:
...
 
+        self.FRAME_UPDATED = Event[[]]()
+        """the frame has been updated"""
+
         self.log = logging.getLogger('rosys.odometer')
 
         wheels.VELOCITY_MEASURED.subscribe(self.handle_velocities)
-        self.prediction: Pose = Pose()
-        self.prediction_frame = Pose3d().as_frame('rosys.odometer.prediction')
+        self.pose: Pose = Pose()
```
falkoschindler:
```suggestion
        self.pose = Pose()
```

rosys/driving/odometer.py
```diff
@@ -103,9 +103,10 @@ def _compute_odometry_frame(local_pose: Pose, global_pose: Pose) -> Pose:
...
 
     def reset(self) -> None:
-        self.prediction = Pose()
+        self.pose = Pose()
         self.detection = None
         self.current_velocity = None
         self.last_movement = 0
         self.history.clear()
         self.odometry_frame = Pose()
+        self._handle_movement()
```
falkoschindler, line 122: I'm not sure if this breaking change is always desired...

falkoschindler (commented):
Besides some inline comments, I'm wondering if we could/should provide a legacy API for `prediction`, `prediction_frame` and `PREDICTION_UPDATED`. Three properties could return `pose`, `frame` and `POSE_UPDATED`, respectively. And they could log a deprecation warning.

rosys/driving/pose_provider.py
pascalzauberzeug: That's a good point, they should not be written from outside. I will make them both properties

rosys/driving/odometer.py
pascalzauberzeug: In what scenario do you not want know when your robot moved?

falkoschindler: If you reset your odometer, the robot didn't necessarily _move_. Although the pose _changed_. So it depends what you want to know.

pascalzauberzeug:
@falkoschindler I don't think we should provide a legacy API for this right now. Currently, the Feldfreund is the only mayor plattform in use that needs the Odometer.
In the coming month there will be several, that's why now is the perfect time to just remove it.

rosys/driving/odometer.py
pascalzauberzeug: We decided to keep it here, because the event is named `POSE_UPDATED` and the pose was updated.
The user should take care of any special cases where reset could cause problems

rosys/driving/odometer.py
pascalzauberzeug: We decided to keep it for now

rosys/driving/pose_provider.py
```diff
@@ -1,9 +1,12 @@
 from typing import Protocol
 
+from nicegui import Event
+
 from ..geometry import Pose
 
 
 class PoseProvider(Protocol):
```
falkoschindler, line 8: While at it, we could add a docstring similar to the `FrameProvider`.

rosys/driving/pose_provider.py
```diff
@@ -1,9 +1,12 @@
...
+    POSE_UPDATED: Event
```
falkoschindler: Let's add a docstring.
And should we define the argument type, like `Event[[]]` or `Event[Pose]`?

rosys/geometry/pose3d.py
```diff
@@ -99,6 +99,15 @@ def ancestors(self) -> Iterator[Frame3d]:
             yield from frame_registry[self.frame_id].ancestors
 
 
+class FrameProvider(Protocol):
+    """Protocol for objects that provide a moving Frame3d reference frame."""
+    FRAME_UPDATED: Event
```
falkoschindler: Let's add a docstring.
And should we define the argument type, like `Event[[]]` or `Event[Pose]`?

rosys/driving/odometer.py
```diff
@@ -22,25 +22,34 @@ def __init__(self, wheels: VelocityProvider) -> None:
         self.WHEELS_TURNED = Event[[]]()
         """the wheels have turned with non-zero velocity"""
 
-        self.PREDICTION_UPDATED = Event[[]]()
-        """the pose prediction has been updated"""
+        self.POSE_UPDATED = Event[[]]()
+        """Emitted on movement or detection correction."""
```
falkoschindler: Why changing the docstring? It feels imprecise now, as if the event isn't emitted for zero velocity.

rosys/driving/odometer.py
```diff
@@ -22,25 +22,34 @@ def __init__(self, wheels: VelocityProvider) -> None:
...
+
+        self.FRAME_UPDATED = Event[[]]()
+        """Emitted on movement or detection correction."""
```
falkoschindler: Same concern like above: Why not simply saying "the frame has been updated"?

falkoschindler (commented):
Just some thoughts about documentation...

rosys/driving/odometer.py
pascalzauberzeug: I think the previous docstring wasn't helpful because it just repeated the name of the event.
And it _**is**_ only emitted on movement since it's guarded by `robot_moved`

falkoschindler (approved):

PR #393, opened 2026-03-26
Fix yaw instability at ±π in GeoPose roundtrip and Spline derivative

pascalzauberzeug:
### Motivation

`GeoPose.from_pose(...).to_local()` can flip the yaw from `+π` to `-π` (or vice versa) depending on the position.
Both `from_pose` and `to_local` used a two-point projection + `atan2` to convert between local yaw and geographic heading.
For headings near ±π, the projected point has a near-zero y-offset whose sign is position-dependent, causing `atan2` to return `+π` at one position and `-π` at another.

Additionally, `Spline.gx`/`gy` evaluated the derivative via polynomial coefficients that mix all four control points.
At `t=0` and `t=1`, large intermediate terms cancel to near-zero, and the residual sign varies by Python version and whether coordinates are `float` vs `np.float64`.

### Implementation

**GeoPose**: replace the two-point projection with a direct formula: `heading = direction - yaw` / `yaw = direction - heading`.
No `atan2`, no point projection — exact roundtrip.
Applied to `from_pose`, `to_local`, `pose_to_geo`, and `pose_to_local`.

**Spline**: switch `gx`/`gy` from the expanded polynomial form (`m*t² + 2n*t + o`) to the de Casteljau form using adjacent control-point differences (`(1-t)²·d₀ + 2(1-t)t·d₁ + t²·d₂`).
Mathematically equivalent, but numerically stable because it avoids catastrophic cancellation of O(1) terms.

### Progress

### DIFF
--- a/rosys/geometry/geo.py
+++ b/rosys/geometry/geo.py
@@ -94,9 +94,8 @@ class GeoPose:
     def from_pose(cls, pose: Pose) -> GeoPose:
         """Create a geo pose from a pose in the current geo reference."""
         assert GeoReference.current is not None
-        geo_point1 = GeoReference.current.point_to_geo(pose)
-        geo_point2 = GeoReference.current.point_to_geo(pose.transform_pose(Pose(x=1)))
-        return cls(geo_point1.lat, geo_point1.lon, geo_point1.direction(geo_point2))
+        geo_point = GeoReference.current.point_to_geo(pose)
+        return cls(geo_point.lat, geo_point.lon, GeoReference.current.direction - pose.yaw)
 
     @property
     def point(self) -> GeoPoint:
@@ -113,9 +112,8 @@ class GeoPose:
     def to_local(self) -> Pose:
         """Transform to local Cartesian coordinates relative to the current geo reference."""
         assert GeoReference.current is not None
-        point1 = GeoReference.current.point_to_local(self)
-        point2 = GeoReference.current.point_to_local(self.point.polar(1, self.heading))
-        return Pose(x=point1.x, y=point1.y, yaw=point1.direction(point2))
+        point = GeoReference.current.point_to_local(self)
+        return Pose(x=point.x, y=point.y, yaw=GeoReference.current.direction - self.heading)
 
     def __str__(self) -> str:
         lat_deg, lon_deg, heading_deg = self.degree_tuple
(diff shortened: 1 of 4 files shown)


falkoschindler (changes requested):
After thoroughly reviewing this PR, I've come to the conclusion that we should keep only a part of it and need to think about how to address the problem further downstream.

### Spline derivative (de Casteljau form)

The two forms are mathematically identical. Expanding the new form:

```
(1-t)^2*(b-a) + 2(1-t)t*(c-b) + t^2*(d-c)
= (b-a) + 2t(a-2b+c) + t^2(d-3c+3b-a)
= o + 2*n*t + m*t^2
```

which is the old form. The new form can be computed inline from the existing `a,b,c,d,e,f,g,h` members - the 6 new `_dx`/`_dy` fields are unnecessary.

De Casteljau removes one layer of cancellation: at `t=1`, the old form evaluates `m + 2n + o` (three large terms cancelling to `d-c`), while the new form yields `d-c` directly. But `d-c` itself is a subtraction of two coordinates that may be nearly equal. If the true derivative is near zero at the endpoint, both forms produce a tiny number with an unreliable sign - de Casteljau just arrives there more cleanly.

So the improvement is real but narrow: it helps when `d-c` is not near zero but the polynomial mixing corrupted the result. It doesn't help when `d-c` genuinely is near zero. And in either case, `atan2(+/-epsilon, negative)` still returns +pi or -pi nondeterministically.

### GeoPose roundtrip (direct formula)

The original two-point projection approach was geometrically correct: it accounted for the fact that the local-to-geo angular relationship varies with position on a curved surface. The new formula `heading = direction - yaw` assumes a constant angular offset (flat-earth approximation), which is only exact at the reference origin.

The roundtrip `direction - (direction - yaw) = yaw` is algebraically exact, which is a nice property. But the original code wasn't wrong - +pi and -pi represent the same orientation. The roundtrip inconsistency only matters if downstream consumers don't handle angle wrapping, which brings us back to the real issue.

### Root cause

The fundamental problem is that +pi and -pi are equivalent angles but distinct floating-point values. No amount of numerical improvement to GeoPose or Spline can fully prevent sign flips at this boundary. The robust fix is angle-aware handling in consumers: normalizing differences, using angular distance for comparisons, and unwrapping before interpolation.

### Recommendation

- Accept the Spline de Casteljau change - it's an independent, unambiguous improvement. But inline the computation to avoid extra fields.
- Reconsider the GeoPose change - it trades geometric correctness for roundtrip consistency, and doesn't fully solve the underlying problem either. The +/-2pi issue in consumers should be addressed at the consumer level.

falkoschindler:
We agreed to close this PR and improve the spline gradients on a new one.

PR #394, opened 2026-03-27
Support server-side crop and rotation on image endpoints

pascalzauberzeug:
### Motivation

Camera image endpoints currently only serve the full frame (optionally undistorted/shrunk). Consumers that need a cropped region or a rotated view must fetch the full image and transform client-side, wasting bandwidth and shifting work to the browser.

### Implementation

Accept `crop_x`, `crop_y`, `crop_w`, `crop_h` and `rotation` as query parameters on both the distorted and undistorted camera image routes. When provided, transforms are applied server-side after undistort and before shrink, reusing the existing `process_ndarray_image` helper. The fast path (no processing, direct JPEG encode) is preserved when no crop/rotation is requested.

- `Rectangle` is constructed only when all four crop params are present; partial crop params are silently ignored.
- `rotation` accepts degrees (0, 90, 180, 270) and is validated via `ImageRotation.from_degrees`, which raises on invalid values.
- `ImageRotation` is now exported from `rosys.vision`.

### Progress


NiklasNeugebauer (approved):
Sorry for the slow review. I added input sanitization and error responses of malformatted inputs.

Looks good otherwise. Nice reuse of existing functions.

PR #396, opened 2026-03-31
Switch from Poetry to uv for dependency management

codingpaula:
## Summary

- ⚠️ change in development tooling: RoSys now uses uv instead of Poetry. If you contribute to or develop against a cloned copy of the repository, please install uv and follow the updated CONTRIBUTING.md.
- Convert `pyproject.toml` from Poetry format to PEP 621 (`[project]` metadata, `[dependency-groups]`, `[project.entry-points]`)
- Keep `poetry-dynamic-versioning` as the build backend (no uv-native substitute yet)
- Update CI workflows (`pytest.yml`, `publish.yml`) from `abatilo/actions-poetry` to `astral-sh/setup-uv@v7`
- Update `Dockerfile` to install and use uv instead of poetry
- Absorb `mkdocs_requirements.txt` into a `docs` dependency group
- Update `CONTRIBUTING.md` and `docs/development/README.md`

## Motivation

Align rosys with the rest of the Zauberzeug ecosystem (NiceGUI, feldfreund_devkit) which have already migrated to uv. Benefits include faster dependency resolution, simpler tooling, and a standardized PEP 621 project format.

## Implementation

- Main dependencies use PEP 508 specifiers in `[project] dependencies`
- Dev dependencies use `[dependency-groups] dev`, docs dependencies use `[dependency-groups] docs`
- `poetry-dynamic-versioning` is retained for git-based versioning: the `[build-system]`, `[tool.poetry]`, and `[tool.poetry-dynamic-versioning]` sections are preserved
- CI publish uses version from tag + `uv build` + `uv publish`
- Dockerfile uses `COPY --from=ghcr.io/astral-sh/uv:latest` and `ENV PATH="/rosys/.venv/bin:$PATH"`

## Progress


falkoschindler (commented):
Thanks, @codingpaula! Can't wait to migrate to uv!

There's just one issue I could find:

- **`uv run python -m pip install` will likely fail (publish.yml:47)**

  ```yaml
  if uv run python -m pip install "rosys==${{ env.VERSION }}"; then
  ```
	
  `uv run` creates a project venv that does **not** include pip by default. This verification step will probably fail with `No module named pip`. NiceGUI's `_pypi.yml` uses bare `python -m pip install` for the same purpose — the runner's system Python has pip pre-installed. Suggest reverting to:

  ```yaml
  if python -m pip install "rosys==${{ env.VERSION }}"; then
  ```

Do you agree?

codingpaula:
Sounds reasonable! I guess to work with uv it should be `uv pip install`, but I think testing it with a "bare" python installation makes more sense. I'll revert that line.

falkoschindler (approved):

falkoschindler:
@codingpaula One last thought: You tagged this PR as "breaking change" but didn't explain it any further. Are you sure this is a breaking change? If so, can you, please, add a short explanation for the release notes? Thanks!

codingpaula:
I added a comment to the summary that users contributing or using RoSys as a cloned not installed through PyPI project need to use uv now instead of poetry.

falkoschindler:
@codingpaula Hm, I don't think this is a breaking change. Here's Claude's explanation:

---

I'd say **no, this is not a breaking change** in the conventional (semver) sense.

A "breaking change" typically means the **public API** of the library changed in a backwards-incompatible way — removed/renamed classes, changed function signatures, altered runtime behavior, etc. This PR does none of that.

What it *does* change is the **contributor/developer workflow**: people who clone the repo need `uv` instead of `poetry`. But that only affects the development setup, not anyone who installs RoSys from PyPI via `pip install rosys` or `uv add rosys`. For those users, nothing changes at all.

Paula's framing is understandable — she's being cautious and transparent about the tooling switch. But labeling it "breaking" in a release note could unnecessarily alarm downstream users who'd expect an API migration. A better fit would be something like:

> **Development tooling:** RoSys now uses uv instead of Poetry. If you contribute to or develop against a cloned copy of the repository, please install uv and follow the updated CONTRIBUTING.md.

codingpaula:
Sounds good! I removed the breaking change label and updated the summary.

PR #400, opened 2026-04-15
Auto-detect MJPEG camera auth scheme

NiklasNeugebauer:
## Motivation

OpenIPC cameras fail to stream because `MjpegDevice` hardcodes `httpx.DigestAuth`, but OpenIPC's Majestic server at `/mjpeg` advertises Basic auth — so every request is rejected with 401 even though the credentials are correct.

## Implementation

- Add `MjpegDevice._resolve_auth` that issues an unauthenticated probe via `client.stream('GET', …)` and reads `WWW-Authenticate`
- Return `httpx.DigestAuth` when the server advertises Digest, otherwise `httpx.BasicAuth`
- Store `username`/`password` instead of a pre-built auth object, so each reconnect re-probes and auto-adapts if the camera's auth scheme changes
- Use `client.stream` (not `client.get`) for the probe so the MJPEG body is never buffered in memory

## Test plan

- [x] Stream from an OpenIPC camera (`root`/`Adminadmin`)
- [x] Stream from a camera without credentials to confirm the no-auth path still works

---
🦌 claude-opus-4-6


ftilde:
Seems to fix mjpeg for OpenIPC :+1:

NiklasNeugebauer:

**Verdict: correct problem, correct core idea; small fixes + a test, and a rebase.** Probing `WWW-Authenticate` to choose Digest vs Basic fixes real OpenIPC 401s.

1. **Don't send credentials when the server never asked.** If the unauthenticated probe returns 200 (camera without auth but credentials configured), `scheme == ''` and the code still attaches `BasicAuth` — shipping the password in cleartext to a server that didn't challenge. Return `None` in the no-challenge case. (The third test-plan checkbox — no-credential camera — currently passes for a different reason: `username is None` short-circuits before the probe.)
2. **Probe cost:** the probe opens the *stream* URL and closes it. Cameras with single-client MJPEG limits may briefly see the slot taken; the sequential open/close is probably fine, but handle a 503 on the follow-up gracefully. An alternative with one request in the common case: request unauthenticated, and only on 401 parse the challenge and retry with the right auth.
3. Multiple-scheme headers (`Digest ..., Basic ...`) resolve to the first token — acceptable, but a comment would help.
4. **Rebase needed** (`mergeable: dirty`; `mjpeg_device.py` has changed on main), and note that #425 relocates this whole code path into the capture subprocess — worth coordinating.
5. **Add a test:** `_resolve_auth` is easily unit-tested with `respx` or a stub transport (Digest challenge → `DigestAuth`; Basic challenge → `BasicAuth`; no challenge → `None` after fix (1); probe error → propagates). `tests/test_mjpeg_device.py` is the natural home.

NiklasNeugebauer:
@ftilde I did a small change upon a new review suggestion. otherwise i say we just merge even if axis is untested

tests/test_mjpeg_device.py
```diff
@@ -24,3 +30,240 @@ def test_returns_none_for_unparsable_value():
...
+    auth = device._auth_for_challenge('Digest realm="cam", nonce="abc"', 'user',
+                                      'secret')  # pylint: disable=protected-access
+
+    assert isinstance(auth, httpx.DigestAuth)
+
+
+async def test_auth_for_challenge_returns_basic_auth_for_basic_challenge(monkeypatch: pytest.MonkeyPatch) -> None:
+    device = _make_device(monkeypatch, username='user', password='secret')
+
+    auth = device._auth_for_challenge('Basic realm="cam"', 'user', 'secret')  # pylint: disable=protected-access
```
ftilde: _auth_for_challenge can be a standalone method. This really simplifies the tests. Don't need a monkeypatched device. I hate monkeypatching.

tests/test_mjpeg_device.py
```diff
@@ -24,3 +30,240 @@ def test_returns_none_for_unparsable_value():
...
+    assert isinstance(auth, httpx.DigestAuth)
+
+
+async def test_run_capture_task_retries_with_digest_auth_after_401(monkeypatch: pytest.MonkeyPatch) -> None:
+    """A fresh httpx.DigestAuth never preemptively attaches credentials: even once our own
+    retry loop selects DigestAuth from the first 401, that auth object still runs its own
+    unauthenticated-then-401-then-authenticated cycle internally (it has no cached challenge
+    yet), so the full exchange is 3 requests, not 2. This is unavoidable without reaching into
+    httpx's private DigestAuth challenge cache, and matches this class's pre-existing (and the
+    original PR's) request count for Digest cameras -- only the Basic and no-challenge cases
```
ftilde: What original PR are we talking about here? The comment is really hard to understand without context in general.

tests/test_mjpeg_device.py
```diff
@@ -24,3 +30,240 @@ def test_returns_none_for_unparsable_value():
...
+        assert request.headers.get('authorization') is None
+        return httpx.Response(200, content=_mjpeg_payload(), request=request)
+
+    transport = httpx.MockTransport(handler)
+    _patch_async_client_transport(monkeypatch, transport)
+    device = _make_device(monkeypatch, on_new_image_data=on_new_image_data)
+
+    await asyncio.wait_for(device.run_capture_task(), timeout=1.0)
+
+    assert len(requests) == 1
```
ftilde: These tests seem VERY specific to the current implementation. Why do we care that we use exactly n requests in a specific scenario? Are we not fine in the cases where we require and don't require auth?

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -82,46 +91,56 @@ async def stream() -> AsyncGenerator[tuple[bytes, float | None], None]:
...
                 try:
                     await self._prepare_stream()
-                    async with client.stream('GET', self._url, auth=self._authentication) as response:  # type: ignore
-                        if response.status_code != 200:
-                            self.log.error('could not connect to %s (credentials: %s): %s %s',
-                                           self._url, self._authentication, response.status_code, response.reason_phrase)
+                    auth: httpx.Auth | None = None
+                    while True:
+                        async with client.stream('GET', self._url, auth=auth) as response:
+                            if response.status_code == 401 and auth is None \
```
ftilde: Do we need to do this in the main capture loop? What about a separate loop that checks if we are authenticate before we move on to the actual capturing business? Much simpler to read/understand (if possible).

ftilde (changes requested):

tests/test_mjpeg_device.py
NiklasNeugebauer: *Claude Fable 5 on behalf of Niklas*

Done in 1ec683aa — `_auth_for_challenge` is a module-level function now and its tests are plain assertions without a device or monkeypatching.

tests/test_mjpeg_device.py
NiklasNeugebauer: *Claude Fable 5 on behalf of Niklas*

That docstring explained why the Digest flow takes exactly three requests (httpx.DigestAuth runs its own challenge round-trip internally). Since the tests no longer count requests, the explanation is moot and the docstring is gone (1ec683aa).

tests/test_mjpeg_device.py
NiklasNeugebauer: *Claude Fable 5 on behalf of Niklas*

Agreed that the exact request counts were incidental. The tests now assert behavior only: frames arrive for Digest/Basic/no-auth cameras, credentials are never sent before a 401 challenge (that one is a security property, not an implementation detail), and the task gives up on rejected credentials or non-401 errors (1ec683aa).

rosys/vision/mjpeg_camera/mjpeg_device.py
NiklasNeugebauer: *Claude Fable 5 on behalf of Niklas*

Restructured in 1ec683aa: auth negotiation now lives in a dedicated `_open_stream` context manager that yields a ready-to-consume 200 response, frame extraction moved to a `_frames` generator, so the capture code has no auth logic left. A fully separate pre-check loop would need its own request and re-open the stream afterwards, which an earlier review round ruled out (extra probe connection can briefly take the slot on single-client MJPEG cameras) — this keeps the request pattern identical while separating the concerns.

tests/test_mjpeg_device.py
```diff
@@ -24,3 +29,56 @@ def test_returns_none_for_unparsable_value():
...
+                       *,
+                       username: str | None = None,
+                       password: str | None = None) -> httpx.Response | None:
+    """Open a device stream against a mocked camera and return the negotiated response."""
+    monkeypatch.setattr('rosys.vision.mjpeg_camera.mjpeg_device.on_startup', lambda _handler: None)
+    device = MjpegDevice(mac='7a:7a:21:00:00:00', ip='127.0.0.1',
+                         username=username, password=password,
+                         on_new_image_data=lambda _image, _timestamp: None)
+    async with httpx.AsyncClient(transport=httpx.MockTransport(handler)) as client:
+        async with device._open_stream(client) as response:  # pylint: disable=protected-access
```
ftilde: One more nit, but worth it, I think, so we can get rid of all the monkeypatching: _open_stream can easily be a function which gets url, username and password as parameters.

ftilde (changes requested):

tests/test_mjpeg_device.py
NiklasNeugebauer: the core change from this would be logging. but i guess we can just have another module logger that users can interpret using the url

NiklasNeugebauer: done

ftilde (approved):
Nice!

PR #401, opened 2026-04-20
Fix continuous looping for backward playback of ReplayCameraProvider

jfrieli:
### Motivation

When using backward replays within the ReplayCameraProvider, the timeline does not loop correctly when it reaches the beginning of the replay.

### Implementation

Inspired by the forward replay case, we added a check that integrates a loop by setting the new_replay_time to the end of the replay.

### Progress


NiklasNeugebauer (approved):
Feels wrong to implicitly assume that we want to loop back to the end, but seems like we can do that. Looks good!

PR #403, opened 2026-04-28
Close wrapped coroutine on automation stop

### LINKED ISSUE #336: Tracking stacks are not cleaned up on automation stop
The stacks of `rosys.analysis.track` are not cleaned up when an automation is stopped manually, only when the automation is finished.

So when stop is pressed three times, it will always show:

```
do_something
do_something
do_something
```

```python
import rosys
from nicegui import ui
from rosys.analysis import track

track.ui()
label = ui.label('not started')


@track
async def do_something():
    for i in range(10):
        label.set_text(f'{i}/10')
        await rosys.sleep(1)
    label.set_text('done')


automator = rosys.automation.Automator(steerer=None, default_automation=do_something)
with ui.row():
    rosys.automation.automation_controls(automator)


ui.run()
```

falkoschindler:
### Motivation

Fixes #336.

When `Automator.stop()` is called, `Automation.__await__` returns without ever throwing into or closing the wrapped coroutine. The coroutine is left suspended at its current `await` point, so any `try/finally` or `async with` block above that point never runs — only eventual GC of the `Automation` will (non-deterministically) close the coroutine and trigger cleanup. In the live UI this can be deferred indefinitely.

The reported symptom in #336 is leaked `rosys.analysis.track` stacks, but the bug is general: every `finally` / `async with` inside a manually stopped automation is silently skipped (locks not released, hardware not reset, files not closed, etc.).

### Implementation

Add `self.coro.close()` to `Automation.__await__`'s `finally` block. This eagerly does what GC would eventually do — `coro.close()` injects `GeneratorExit` into the suspended frame so cleanup blocks fire deterministically on every exit path (manual stop, normal completion, or exception).

**Trade-off:** a `finally` block that `await`s something will raise `RuntimeError` ("coroutine ignored GeneratorExit") under `coro.close()` — that's the standard Python semantic. Currently those blocks don't run at all, so this is strictly an improvement. Truly cancellable cleanup with `await` inside `finally` would require throwing a dedicated exception instead, which is out of scope here.

The existing `test_finally_block` regression test was passing on `main` for the wrong reason: refcount-driven deallocation closed the coroutine synchronously at the moment `automator.automation = None` was assigned. In a live NiceGUI app, an extra reference (likely a cycle that only the cyclic GC can break) defers that. The test is strengthened to pin refs to the coroutine and the `Automation` during the loop, defeating refcount-driven cleanup and reproducing the live-app behavior. It now fails on `main` and passes with the fix.

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will…"

### DIFF
--- a/rosys/automation/automation.py
+++ b/rosys/automation/automation.py
@@ -112,6 +112,7 @@ class Automation:
             raise
         finally:
             self._is_waited = False
+            self.coro.close()
             _CURRENT_AUTOMATION.reset(token)
         return None
 
--- a/rosys/automation/automation.py
+++ b/rosys/automation/automation.py
@@ -112,8 +112,8 @@ class Automation:
             raise
         finally:
             self._is_waited = False
-            self.coro.close()
             _CURRENT_AUTOMATION.reset(token)
+            self.coro.close()
         return None
 
     def pause(self) -> None:
--- a/tests/test_automations.py
+++ b/tests/test_automations.py
@@ -71,11 +71,16 @@ async def test_finally_block(automator: Automator):
         finally:
             events.append('tock')
 
-    automator.start(run())
-    await forward(seconds=10)
-    automator.stop(because='test')
-    await forward(seconds=1)
-    assert events == ['tick', 'tick', 'tick', 'tick', 'tock']
+    pinned: list = []  # prevent refcount-driven GC from masking the bug
+    for _ in range(2):
+        coro = run()
+        pinned.append(coro)
+        automator.start(coro)
+        await forward(seconds=10)
+        pinned.append(automator.automation)
+        automator.stop(because='test')
+        await forward(seconds=1)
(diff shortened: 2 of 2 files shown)


pascalzauberzeug (approved):
We should implement an exception to catch that RuntimeError in the future, but for now we can use events to handle that.

```python
import rosys
from nicegui import Event, ui

FINALLY_EVENT = Event[[]]()


async def handle_finally():
    await rosys.sleep(3)
    status.text += 'Finally event handled. '
    FINALLY_EVENT.unsubscribe(handle_finally)


async def do_something():
    try:
        status.text += '🟢 Try... '
        FINALLY_EVENT.subscribe(handle_finally)
        await rosys.sleep(3)
    finally:
        FINALLY_EVENT.emit()
        status.text += '🔴 Finally. '

automator = rosys.automation.Automator(steerer=None, default_automation=do_something)
rosys.automation.automation_controls(automator)

status = ui.label()

ui.run()

```

PR #404, opened 2026-05-05
Make GNSS max_measurement_age configurable at runtime

pascalzauberzeug:
### Motivation

The fixed `MAX_MEASUREMENT_AGE` class constant could not be adjusted without code changes, making it cumbersome to tune for specific setups or experiment with different thresholds.

### Implementation

- Replace the `MAX_MEASUREMENT_AGE` class constant with a `max_measurement_age` constructor parameter (default unchanged at `0.05`s) stored as `self._max_measurement_age`.
- Extend `developer_ui()` for live tuning.

⚠️ Breaking change: If your project changes `MAX_MEASUREMENT_AGE`, you have to add the new parameter instead

### Progress

### DIFF
--- a/rosys/hardware/gnss/gnss_hardware.py
+++ b/rosys/hardware/gnss/gnss_hardware.py
@@ -146,4 +146,4 @@ class GnssHardware(Gnss):
             ui.number('Max Measurement Age', min=0.0, step=0.01, suffix='s') \
                 .bind_value(self, '_max_measurement_age') \
                 .classes('w-32') \
-                .tooltip('Maximum age of measurements to be accepted (default: 0.05s)')
+                .tooltip('Maximum age of measurements to be accepted')
--- a/rosys/hardware/gnss/gnss_hardware.py
+++ b/rosys/hardware/gnss/gnss_hardware.py
@@ -143,7 +143,7 @@ class GnssHardware(Gnss):
     def developer_ui(self) -> None:
         with ui.column():
             super().developer_ui()
-            ui.number('Max Measurement Age', min=0.0, step=0.01, suffix='s') \
+            ui.number('Max Measurement Age', min=0.0, step=0.01, suffix='s', format='%.3f') \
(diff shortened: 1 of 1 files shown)


falkoschindler (commented):
I was about to approve, but Claude has some suggestions worth considering:

- **`developer_ui()` nesting differs from the simulation pattern.** `gnss_simulation.py` calls `super().developer_ui()` _first_ and then opens a sibling `ui.column()`. The new code wraps `super().developer_ui()` inside its own `ui.column()`, which puts the parent's column inside this new one — a nested column rather than a sibling. For visual consistency with the simulation override I'd flip it:

  ```python
  def developer_ui(self) -> None:
      super().developer_ui()
      with ui.column():
          ui.number('Max Measurement Age', min=0.0, step=0.01, suffix='s') \
              .bind_value(self, '_max_measurement_age') \
              .classes('w-32') \
              .tooltip('Maximum age of measurements to be accepted')
  ```

- **Tooltip mentions "default: 0.05s"**, but the default is now per instance (any constructor argument silently makes the tooltip lie). Drop the parenthetical, or build it from the constructor value if you want to keep it.
- **Missing `format`.** Simulation's number inputs use `format='%.3f'` etc. With `step=0.01`, `format='%.2f'` would match (`%.3f` if you want sub-step display). Minor.
- **Removal of `MAX_MEASUREMENT_AGE` class attr is a public-API break** for any subclass / external code that referenced it. A quick grep across the repo shows no other references, so almost certainly fine — just flagging since the constant was uppercased (i.e. looked like a stable knob). Might need a "breaking change" tag and a migration message for the release notes.

pascalzauberzeug:
@falkoschindler 

1. I would keep it that way as long it's only one number. The simulation ui adds a lot more numbers. Images below
2. Default removed
3. Format added
4. Label added


image 

image

falkoschindler (approved):

PR #405, opened 2026-05-08
Add optional frame for projection

codingpaula:
### Motivation

`Calibration.project_from_image` always resolved camera extrinsics to the world frame. When the world frame is e.g. a geo frame with imprecise origin, the result inherits that inaccuracy — even when the caller only cares about positions relative to the robot.

### Implementation

Added an optional `frame: Frame3d | None = None` to `project_from_image`. When omitted, behaviour is unchanged (extrinsics resolve to world). When given, `self.extrinsics.relative_to(frame)` is used instead, so the returned coordinates are expressed in `frame`.

The parameter is forwarded through all three overloads (`Point`, `list[Point]`, `FloatArray`) and into the round-trip `project_to_image` call used by the `reprojection_tolerance` check.

### Progress

### DIFF
--- a/rosys/vision/calibration.py
+++ b/rosys/vision/calibration.py
@@ -253,7 +253,7 @@ class Calibration:
         """Project a point in image coordinates to a plane in world coordinates.
 
         :param image_coordinates: The image coordinates to project.
-        :param target_height: The height of the plane in world coordinates.
+        :param target_height: The height of the plane in ``frame`` or world coordinates.
         :param reprojection_tolerance: The reprojection tolerance as a fraction of the maximum image dimension.
         :param frame: The coordinate frame to project to. If none is given, project to world coordinates.
         :return: The world coordinates of the projected point, or None if projection failed.
@@ -268,7 +268,7 @@ class Calibration:
         """Project points in image coordinates to a plane in world coordinates.
 
         :param image_coordinates: The list of image coordinates to project.
-        :param target_height: The height of the plane in world coordinates.
+        :param target_height: The height of the plane in ``frame`` or world coordinates.
         :param reprojection_tolerance: The reprojection tolerance as a fraction of the maximum image dimension.
         :param frame: The coordinate frame to project to. If none is given, project to world coordinates.
         :return: The world coordinates of the projected points, or None if any projection failed.
@@ -283,7 +283,7 @@ class Calibration:
         """Project points in image coordinates to a plane in world coordinates.
 
         :param image_coordinates: (Nx2) The image coordinates to project.
-        :param target_height: The height of the plane in world coordinates.
+        :param target_height: The height of the plane in ``frame`` or world coordinates.
         :param reprojection_tolerance: The reprojection tolerance as a fraction of the maximum image dimension.
         :param frame: The coordinate frame to project to. If none is given, project to world coordinates.
         :return: (Nx3) The world coordinates of the projected points.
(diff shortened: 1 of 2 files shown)


NiklasNeugebauer:
didnt review fully but sounds good

rosys/vision/calibration.py
```diff
@@ -248,60 +248,69 @@ def project_to_image(self, coordinates: Point3d | list[Point3d] | FloatArray, *,
...
+                                           reprojection_tolerance=reprojection_tolerance,
+                                           frame=frame)[0]
         if not isinstance(image_coordinates, np.ndarray):
             image_array = np.array([point.tuple for point in image_coordinates], dtype=np.float64)
             world_array = self.project_from_image(image_array,
                                                   target_height=target_height,
-                                                  reprojection_tolerance=reprojection_tolerance)
+                                                  reprojection_tolerance=reprojection_tolerance,
+                                                  frame=frame)
             return [Point3d(x=point[0], y=point[1], z=point[2]) if not np.isnan(point).any() else None for point in world_array]
```
codingpaula: @NiklasNeugebauer while discussing this with Claude, the question came up, if the frame should be given to the `Point3d`(s) that are created here. What do you think?

NiklasNeugebauer: Yes, that would make sense.
Since we currently always project to world it should also not break anything current.

falkoschindler (changes requested):
Nice change — the `frame` parameter is forwarded consistently through all overloads and the reprojection round-trip, `frame=None` keeps the old behaviour exactly, and the existing caller stays unaffected. One change request:

**Please fold the two new tests into `test_projection_from_one_frame_into_world_frame`.** They largely re-create setup that already exists there: that test already attaches the extrinsics to a frame, holds both `world_point` and its `frame_point` in one loop, tests `frame=` forwarding on `project_to_image`, and tests `project_from_image` → world. The new `frame=` → local case is the missing symmetric half of that same loop (rename it to drop "into_world_frame"), with roughly:

```python
frame_point_ = cam.calibration.project_from_image(image_point_from_frame, target_height=frame_point.z, frame=cam_frame)
assert np.allclose(frame_point.tuple, frame_point_.tuple, atol=1e-6)
```

That covers the local-frame feature and the `Point`-overload forwarding; adding the `.array` form for one point covers the numpy overload too. Coverage grows and the duplicated fixtures go away.

rosys/vision/calibration.py
falkoschindler: I fixed it in e8a86390.

falkoschindler (approved):
LGTM 👍 Following up on the discussion about whether projected points should carry the target frame, the last three commits round this out nicely:

- **`3a2cdfc`** — folds the frame-projection tests into a single `test_projection_with_frame_argument`, covering both the legacy world-frame path (no `frame`) and the new `frame=cam_frame` path, including the numpy-array overload.
- **`1bcfdec`** — clarifies the `target_height` docstrings: the plane height is now interpreted in `frame` coordinates when a frame is given, otherwise world coordinates.
- **`e8a8639`** — tags the returned `Point3d`s with the target frame via `.in_frame(frame)`, as discussed. With `frame=None` this resolves to world, so existing behavior is unchanged. The test now also asserts `frame_point_.frame_id == cam_frame.id` and that `.resolve()` maps back to the original world point, so the frame contract is locked in.

The change stays symmetric with `project_to_image`, and the default path is provably unchanged (`relative_to(None) == resolve()`). Nice and clean — approving.

PR #406, opened 2026-05-19
add possible patch for client deleted error in 3d scene

codingpaula:
### Motivation

The 3d scene has been causing some problems when we use it with our robots. If two tabs are opened and one closed or the tab is reloaded, we often see "client has been deleted" errors from NiceGUI that completely spam our log which makes debugging something else really hard.

### Implementation

This is a possible patch for the problem in RoSys that Claude and I came up with. The real problem though might lie deeper. We still need to investigate this further.
Here the line `if not await run.cpu_bound(CameraProjector.allclose, texture.args[1], coordinates):` is seen as the culprit and a new mechanism to update the coordinates is implemented.

### Progress

### DIFF
--- a/rosys/vision/camera_objects_.py
+++ b/rosys/vision/camera_objects_.py
@@ -135,5 +135,5 @@ class CameraObjects(Group):
             texture.move(z=z)
             if texture.args[0] != url:
                 texture.set_url(url)
-            if cam_id in changed:
+            if changed is not None and cam_id in changed:
                 texture.set_coordinates(coordinates)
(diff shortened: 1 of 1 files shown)


falkoschindler:
@codingpaula I filed a bug upstream: https://github.com/zauberzeug/nicegui/issues/6058. It shouldn't be hard to fix, so upgrading to NiceGUI 3.13 (once released) should resolve the problem in RoSys without the need for this patch.

codingpaula:
great, I'll leave this open here to be used while debugging robots until the real fix in NiceGUI is ready.

falkoschindler:
With [NiceGUI 3.13](https://github.com/zauberzeug/nicegui/releases/tag/v3.13.0) being released, this problem shouldn't occur anymore.

pascalzauberzeug:
This PR can be closed when https://github.com/zauberzeug/rosys/pull/412 is merged

falkoschindler:
The future is now! 🚀

PR #408, opened 2026-05-20
Rtsp ntp timed image

ftilde:
### Motivation

We want precise timing of camera frames to match them to the robot pose.

### Implementation

Relies on [divinius](https://github.com/zauberzeug/divinus/pull/9) to provide ntp-time information encoded in the rtsp timestamp. The timestamp wraps roughly every 13 hours, but we snap it to the closest rosys time (camera time should never be >13 hours off). Note that it requires an additional camera-specific offset for true timing that requires external calibration and so far has only been done for a specific camera with specific resolution, so this requires some more cleanup.

### Progress

### DIFF
--- a/rosys/vision/rtsp_camera/rtsp_camera.py
+++ b/rosys/vision/rtsp_camera/rtsp_camera.py
@@ -1,16 +1,11 @@
 import logging
-import statistics
-from collections import deque
 from typing import Any, Literal, Self
 
-from ... import rosys
 from ..camera.configurable_camera import ConfigurableCamera
 from ..camera.transformable_camera import TransformableCamera
 from ..image import Image, ImageArray
 from ..image_processing import process_ndarray_image
 from .rtsp_device import RtspDevice
-
-_JITTER_WINDOW = 30
 
 
 class RtspCamera(ConfigurableCamera, TransformableCamera):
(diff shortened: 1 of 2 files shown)


NiklasNeugebauer:

**Verdict: park or close in favor of #422** once the camera firmware carries capture time in a standard form. The engineering inside is careful (tee'd RTP branch, GDP PTS matching, and the wrap-window snapping is mathematically right for a ±13.25 h ambiguity), and the description already flags it needs cleanup. The disqualifying problems as *framework* code:

1. **Vendor-specific semantics applied unconditionally to every RTSP camera.** The code interprets the on-wire RTP timestamp as camera wall-clock (the divinus `pack_ts` behavior). Per RFC 3550 a normal camera's RTP timestamp has a *random* initial offset — for every non-divinus camera (e.g. the Jovision devices this class also serves) the wrap-window snap will produce a confidently wrong `Image.time`, anywhere up to ±6.6 h off, silently. This must be gated per vendor or per camera flag.
2. **`CAMERA_PIPELINE_DELAY_S = 0.0985` is hardcoded for one camera at one resolution.** The resolution-match guard is good, but the constant itself doesn't belong in `rtsp_device.py`; it belongs in per-camera config/calibration data.
3. **Transport change for everyone:** `protocols=tcp latency=0` → `protocols=udp latency=20 buffer-mode=slave` changes streaming behavior (loss tolerance, NAT traversal, latency) for all RTSP users, unrelated to timing. If UDP is required for meaningful timestamps, gate it behind the same flag as (1).
4. **Timestamp discontinuity at stream start:** the first frame(s) fall back to receive time until the RTP branch catches up, then jump to capture-time-minus-pipeline-delay — a ~100 ms+ step between consecutive frames in the same stream. Downstream pose-matching should at least be able to tell which clock a frame came from.
5. Despite the checked checkbox, **no test files were added** (the diff touches only `rtsp_camera.py`/`rtsp_device.py`). The wrap-window math (`k = round(...)`) is pure and eminently unit-testable, including the wrap boundary at 2³²/90 kHz.

PR #413, opened 2026-06-14
Run async cleanup when an automation is stopped

pascalzauberzeug:
### Motivation

#403 made automation cleanup deterministic by adding `coro.close()` on stop. But `close()` injects `GeneratorExit`, and Python forbids `await` while handling it, so any `finally:` / `async with` block that awaits raises `RuntimeError: coroutine ignored GeneratorExit`. This broke `WaypointNavigation.start()`'s `finally: await self.finish()` — wheels-stop / implement-deactivate never ran on stop.

Follow-up to #403 (and the original #336).

### Implementation

Stopping now behaves like `asyncio` cancellation instead of `generator.close()`:

- New `AutomationStopped(BaseException)`. On stop the automator throws it into the coroutine and keeps driving the coroutine to completion, so `try/finally` and `async with` cleanup runs — including `await` during cleanup. It derives from `BaseException` (like `CancelledError`) so user `except Exception` blocks don't swallow the stop. The existing `coro.close()` stays only as a no-op safety net.
- `parallelize` gets the same cancel-and-drain: on stop it injects the exception into every unfinished child and drives each child's cleanup to completion instead of closing it with `GeneratorExit`.

### Progress


falkoschindler (approved):
Approving this. 👍

Full disclosure: the automation driver internals have always been a bit beyond my comprehension — and that was true even before this PR. So in line with our new policy, I'm trusting an approving AI-based review here rather than claiming to have verified every line myself.

The review traced the stop-injection flow through both `automation.py` and `parallelize.py`, confirmed the `AutomationStopped`/`BaseException` design behaves like `asyncio` cancellation (cleanup runs, `except Exception` doesn't swallow it), checked that stop-completion timing is preserved (`is_running` stays `True` until the drain finishes), and ran the suite — all 15 tests in `tests/test_automations.py` pass, including the new async-`finally` regression tests and the production-faithful `ParkOnFuture` variant.

Thanks for the thorough tests and the clear write-up of the #403 trade-off this resolves.

PR #417, opened 2026-06-17
Add ArkVision (ArkCam Basic+) camera support for MJPEG and RTSP

NiklasNeugebauer:
### Motivation

Zauberzeug's Geringhoff project uses Ark Vision Systems ArkCam Basic+ mini cameras.
They need to work in RoSys as both MJPEG and RTSP image sources, and the settings RoSys already models (fps, resolution, mirror, bitrate) should be adjustable through the camera's REST API.

#### New ArkVision support (additive — no change to existing camera behavior)

- Register an `ARKVISION` vendor (MAC prefix `18:fd:cb`) and its stream URLs in `mjpeg_camera/vendors.py` (`http://<ip>:81/`) and `rtsp_camera/vendors.py` (`rtsp://<ip>:8554/h264`, non-standard port, single H.264 stream, no auth).
- Add a shared `ArkVisionRestClient` (`rosys/vision/arkvision_rest_client.py`) holding the REST transport and the sensor-mode-dependent frame-rate mapping. `ArkVisionSettingsInterface` (MJPEG: fps/resolution/mirror, and enabling the HTTP-MJPEG stream via `mJpegStreamType=2`) and `ArkVisionRtspInterface` (H.264: fps/bitrate) subclass it.
- Add `ArkVisionMjpegDevice` and dispatch it from `MjpegDeviceFactory`; select `ArkVisionRtspInterface` for `ARKVISION` in `RtspDevice`.
- Map `fps` to the sensor-mode-dependent `framerate` enum (read from `/imageSensorMode`); map `resolution`/`bitrate` to the nearest supported enum value.
- Extend the `docs/examples/cameras/control.py` settings panel with resolution and mirror controls (MJPEG) and a bitrate control (RTSP).

#### Shared / base-class changes (touch code used by other cameras)

- `MjpegDevice`: add a no-op `_prepare_stream()` hook, awaited right before the stream opens and on every restart, so a vendor can enable its stream on demand — inherited as a no-op by Axis/Motec/GoodCam/OpenIPC.
- `MjpegDevice`/`MjpegCamera`: widen the `fps`/`resolution`/`mirrored` getters to `Optional`, and have the base no-op getters return `None` instead of `0`/`(0,0)`/`False`, so `ConfigurableCamera` keeps the cached/registered value (matches the already-Optional RTSP side; affects the generic MJPEG path used by GoodCam/OpenIPC).
- `RtspDevice`: widen the `_settings_interface` type to include the ArkVision interface, and centralize the `avdec` clamp in `_clamp_avdec` (a passthrough for non-ArkVision vendors).

> [!WARNING]
> **Behavioral change (no public API removed or renamed).** The base `MjpegDevice` settings getters (`get_fps`/`get_resolution`/`get_mirrored`) now return `None` instead of `0`/`(0,0)`/`False`, and their return type widened to `Optional`. These getters are normally not called directly — they are driven through the `ConfigurableCamera` interface (`set_parameters` / `parameters` / the parameter cache), which already treats `None` as "keep the cached value" — so the practical impact is limited to:
> - MJPEG cameras **without** a settings interface (GoodCam, OpenIPC) now report the registered default values for `fps`/`resolution` instead of `0`/`(0, 0)`.
> - Code that calls `MjpegDevice.get_fps()` (etc.) directly and statically types the result as non-`None` may need updating.

Verified end-to-end against a physical ArkCam Basic+ mini: MJPEG and RTSP both produce images, and `fps`/`resolution`/`mirror` (MJPEG) and `fps`/`bitrate` (RTSP) round-trip through the REST interface. `mypy`, `pylint` and pre-commit pass.

> [!NOTE]
> The MAC OUI prefix `18:fd:cb` was observed on a single test unit and may need expanding for other ArkCam SKUs. Achievable `fps` depends on the active sensor mode (`set_fps` snaps to the nearest value within the current mode; it does not change the sensor mode).

### Progress

- [x] Pytests have been added (or are not necessary). — verified locally against the physical camera; automated tests not included as they require the hardware.
- [x] Documentation has been added (or is not necessary). — the `control.py` example now exposes all adjustable ArkVision settings.

---
⬛ claude-opus-4-8[1m]


NiklasNeugebauer:
@pascalzauberzeug @falkoschindler @ftilde does anyone want to have a look at this? mostly api specifics and some architectural changes.

rosys/vision/mjpeg_camera/mjpeg_camera.py
```diff
@@ -99,7 +99,7 @@ async def _set_fps(self, fps: int) -> None:
         assert self.device is not None
         await self.device.set_fps(fps)
 
-    async def _get_fps(self) -> int:
+    async def _get_fps(self) -> int | None:
```
ftilde, line 102: I think we need to change the signatures of the other implementors/derived classes.

ftilde: The same for _get_resolution etc.

rosys/vision/mjpeg_camera/arkvision_mjpeg_device.py
```diff
@@ -0,0 +1,47 @@
...
+
+from .arkvision_settings_interface import ArkVisionSettingsInterface
+from .mjpeg_device import MjpegDevice
+from .vendors import VendorType, mac_to_vendor
+
+
+class ArkVisionMjpegDevice(MjpegDevice):
+    """MJPEG device for Ark Vision ArkCam Basic+ cameras.
+
+    Images come from the MJPEG-over-HTTP stream on port 81, which is enabled on demand via the
```
ftilde: Is the detail about the port and enabling really relevant for the user?

rosys/vision/arkvision_rest_client.py
```diff
@@ -0,0 +1,90 @@
...
+
+def nearest_index(values: list[int], target: int) -> int:
+    """Index of the entry in `values` closest to `target`; ties resolve to the lower index."""
+    return min(range(len(values)), key=lambda i: (abs(values[i] - target), i))
+
+
+class ArkVisionRestClient:
+    """Base for ArkCam Basic+ REST settings interfaces (``http://<ip>/rest/...``).
+
+    The REST server has no authentication: a GET reads a settings group, a POST writes (partial)
```
ftilde: Seems mostly not relevant for users. The framerate note could live at that function if really relevant.

rosys/vision/mjpeg_camera/mjpeg_camera.py
ftilde: Welp, apparently, it's fine in python since we are not widening the set of types. Nevermind then.

rosys/vision/mjpeg_camera/mjpeg_device_factory.py
```diff
@@ -25,6 +26,11 @@ def create(mac: str, ip: str, *,
                                     username=username, password=password,
                                     control_port=control_port, on_new_image_data=on_new_image_data)
 
+        if mac_to_vendor(mac) == VendorType.ARKVISION:
```
ftilde: Not really the "fault" of this PR, but shouldn't we precompute mac_to_vendor before the equality-checks? Maybe we can just quick-fix this here.

ftilde (dismissed):
Approved overall. I don't like some of the verbose comments, but maybe that's just me.

I haven't tested it on hardware of course, but I guess you did. :D

rosys/vision/mjpeg_camera/mjpeg_device_factory.py
NiklasNeugebauer: good insight. Will add that.

rosys/vision/mjpeg_camera/arkvision_mjpeg_device.py
NiklasNeugebauer: I'll have another look thriugh the docstrings and comments. Thanks

ftilde (approved):

PR #418, opened 2026-06-17
Make OpenIPC camera settings (fps/bitrate/resolution/mirror) configurable

NiklasNeugebauer:
### Motivation

RoSys can *stream* from OpenIPC cameras but couldn't *configure* them: the per-vendor settings interface that exists for Jovision/Motec/Axis was missing for OpenIPC, so `fps`/`bitrate`/`resolution`/`mirror` changes were silently dropped. OpenIPC cameras run the [divinus](https://github.com/zauberzeug/divinus) streamer, which exposes a full HTTP control API — this wires RoSys's existing camera parameters to it.

### Implementation

- Add `OpenIpcSettingsInterface` (`rosys/vision/openipc_settings_interface.py`): an `httpx` client for the divinus HTTP API, mapping RTSP/H.26x params to `/api/mp4` (fps, bitrate), MJPEG params to `/api/mjpeg` (fps, resolution), and mirroring to `/api/isp`. The RTSP getters/setters reuse the existing `JovisionInterface` contract, so `RtspDevice` needs only a vendor branch.
- Wire OpenIPC into `RtspDevice` (`VendorType.OPENIPC` → `OpenIpcSettingsInterface`).
- Add `OpenIpcMjpegDevice(MjpegDevice)` and route OpenIPC MJPEG cameras to it in `MjpegDeviceFactory` (GOODCAM keeps the plain device).
- Fix a pre-existing crash in `RtspDevice.shutdown` (a concurrent restart could null `_capture_process` between the None-check and the `.returncode` read) — surfaced because every settings change restarts the pipeline.
- Since `control.py` builds its UI from `get_capabilities()`, the existing camera settings panel now drives a real OpenIPC camera with no UI changes.

Companion to zauberzeug/divinus#49, which documents the API this drives.

### Verified on hardware (192.168.1.221)

- `fps` and `bitrate` apply live via `/api/mp4`; the divinus HTTP API needs no auth (the default `none` is correct).
- Bitrate is **kbps** on both sides (confirmed from the divinus source); passed through unchanged.
- `mirror`/`flip` (`/api/isp`) are config-only in divinus — applied at channel init, so they take a divinus restart and do not change live (left as-is).
- divinus ignores the RTSP `/stream=N` path (single H.26x stream, no substream), and the test camera streams **H.265** — `avdec` (the RoSys-side decoder) must be set to match.

### Notes / possible follow-ups (out of scope)

- `avdec` stays a manual decoder selection that must match the camera's encoder; auto-syncing it to `/api/mp4?h265=` is a possible future enhancement.
- `exposure`/`gain` (`/api/exposure`, `/api/gain`) are not yet exposed as RoSys parameters.

### Progress

---
⬛ claude-opus-4-8 (1M context)

### DIFF
--- a/rosys/vision/mjpeg_camera/openipc_mjpeg_device.py
+++ b/rosys/vision/mjpeg_camera/openipc_mjpeg_device.py
@@ -1,0 +1,44 @@
+from collections.abc import Awaitable, Callable
+
+from ..openipc_settings_interface import OpenIpcSettingsInterface
+from .mjpeg_device import MjpegDevice
+from .vendors import VendorType, mac_to_vendor
+
+
+class OpenIpcMjpegDevice(MjpegDevice):
+    """MJPEG device for OpenIPC cameras running the divinus streamer.
+
+    Extends :class:`MjpegDevice` with an :class:`OpenIpcSettingsInterface` so fps,
+    resolution and mirroring are applied on the device via the divinus HTTP API
+    instead of the no-op defaults of the base class.
+    """
+
+    def __init__(self, mac: str, ip: str, *,
+                 username: str | None = None,
+                 password: str | None = None,
+                 on_new_image_data: Callable[[bytes, float], Awaitable | None]) -> None:
+        super().__init__(mac, ip, username=username, password=password, on_new_image_data=on_new_image_data)
+
+        vendor = mac_to_vendor(mac)
+        if vendor != VendorType.OPENIPC:
+            raise ValueError(f'OpenIpcMjpegDevice can only be used with OPENIPC devices. Got {vendor} for mac="{mac}"')
+
+        self.settings_interface = OpenIpcSettingsInterface(ip, username=username, password=password)
+
+    async def set_fps(self, fps: int) -> None:
+        await self.settings_interface.set_mjpeg_fps(fps)
+
+    async def get_fps(self) -> int:
+        return await self.settings_interface.get_mjpeg_fps() or 0
+
+    async def set_resolution(self, width: int, height: int) -> None:
+        await self.settings_interface.set_resolution(width, height)
+
+    async def get_resolution(self) -> tuple[int, int]:
+        return await self.settings_interface.get_resolution()
+
+    async def set_mirrored(self, mirrored: bool) -> None:
+        await self.settings_interface.set_mirrored(mirrored)
+
+    async def get_mirrored(self) -> bool:
+        return await self.settings_interface.get_mirrored() or False
--- a/rosys/vision/rtsp_camera/rtsp_device.py
+++ b/rosys/vision/rtsp_camera/rtsp_device.py
@@ -71,16 +71,17 @@ class RtspDevice:
         return url
 
     async def shutdown(self) -> None:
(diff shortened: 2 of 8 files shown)


NiklasNeugebauer:
@pascalzauberzeug @falkoschindler @ftilde does anyone want to have a look thorugh this? mostly camera/ api specific code

rosys/vision/rtsp_camera/rtsp_device.py
```diff
@@ -42,9 +43,11 @@ def __init__(self, mac: str, ip: str, *,
 
         vendor_type = mac_to_vendor(mac)
 
-        self._settings_interface: JovisionInterface | None = None
+        self._settings_interface: JovisionInterface | OpenIpcSettingsInterface | None = None
```
ftilde: We could also define an interface for that, but maybe it's overkill for two implementations.

rosys/vision/mjpeg_camera/openipc_mjpeg_device.py
```diff
@@ -0,0 +1,44 @@
+from collections.abc import Awaitable, Callable
+
+from ..openipc_settings_interface import OpenIpcSettingsInterface
+from .mjpeg_device import MjpegDevice
+from .vendors import VendorType, mac_to_vendor
+
+
+class OpenIpcMjpegDevice(MjpegDevice):
```
ftilde: It's a bit confusing that the settings interface between Mjpeg and RTSP is completely different (one uses a parent/child, the other a member). We could unify that, but maybe that out of scope for this PR...

ftilde (dismissed):
With this PR we assume that the cameras run divinus, rather than the default streamer, but I don't see a way how we can distinguish them, and I think we should apply it anyway. If a users tries to apply a setting to a camera without divinus, that call will fail visibly (with a stack trace).

rosys/vision/mjpeg_camera/openipc_mjpeg_device.py
NiklasNeugebauer: I agree, thats historical baggage. Lets file a general Issue for that

NiklasNeugebauer: we can also add the interface to that issue

NiklasNeugebauer: https://github.com/zauberzeug/rosys/issues/419 filed

NiklasNeugebauer:
@ftilde We could rename the vendor to `Zauberzeug` to make clear its not stock openipc
Maybe we should also discuss the current fleet and possible update strategy separately.

ftilde:
> @ftilde We could rename the vendor to `Zauberzeug` to make clear its not stock openipc Maybe we should also discuss the current fleet and possible update strategy separately.

Sounds like a good idea. Or `OpenIPC-Zauberzeug`. Deploying the current code on robots with old firmware doesn't (really) hurt as I wrote above. We get a stacktrace about not being able to set parameters, but that basically serves as a reminder to update/change the firmware.

ftilde (approved):

PR #420, opened 2026-06-22
Add GmslCamera for GMSL2/FPD-Link cameras on NVIDIA Jetson

NiklasNeugebauer:
### Motivation

We need to integrate GMSL2/FPD-Link cameras (a global-shutter camera on a Jetson GMSL carrier board) into RoSys. The GMSL link, driver and `nvarguscamerasrc` access have already been validated on the target device; this PR adds the RoSys camera class so those frames are available to the UI and detection pipeline. The same camera also needs configurable long exposure for an upcoming long-exposure (material-flow streak) approach.

### Implementation

Frames are captured through NVIDIA's Argus stack (`nvarguscamerasrc`), so the hardware ISP performs debayering, white balance and tone mapping — exposure and gain either run on the ISP's auto algorithms or are pinned to fixed values (a long manual exposure + low fps yields the streak mode, with ISP correction left on). The pipeline is run as a `gst-launch-1.0 ... ! gdppay ! fdsink` subprocess whose frames are read off a pipe, reusing the exact mechanism already proven in `RtspDevice`.

- Add `rosys/vision/gmsl_camera/` package:
  - `GmslCamera` (`ConfigurableCamera` + `TransformableCamera` + `CalibratableCamera`) with `auto_exposure`/`exposure` (absolute seconds), `auto_gain`/`gain`, `fps`, `width`, `height` parameters, identified by Argus `sensor_id`
  - `GmslDevice` driving the Argus GStreamer pipeline; parameter changes rebuild the pipeline (debounced so a batch of setters triggers a single restart)
  - `GmslCameraProvider` — cameras are registered explicitly by `sensor_id`; there is no auto-discovery (see below)
  - pure `build_argus_command(...)` so pipeline construction is unit-testable
- The capture loop auto-reconnects (with backoff) when the Argus pipeline exits, since the Jetson Argus stack intermittently fails to create a session and the long GMSL coax can drop
- Refactor: extract the GStreamer Data Protocol reader (`GDPPacket`, `GDPPayloadType`, caps parsing) into a shared `rosys/vision/gstreamer.py` and update `rtsp_device.py` to import it (single source of truth, no behavior change)
- `connect()` bails with a warning when `gst-launch-1.0` is absent, and pipeline-spawn / already-exited processes are handled without leaking tracebacks (so the example runs cleanly off-Jetson)
- Export `GmslCamera`/`GmslCameraProvider` from `rosys.vision`
- Add `docs/examples/cameras/gmsl.py` (minimal), `gmsl_control.py` (interactive single-camera demo) and `jetson_cameras.py` (multiple cameras via the provider), plus a docs section
- Integrate `GmslCameraProvider` into the multi-camera `control.py` example alongside the RTSP/MJPEG/USB/simulated providers, showing it is a first-class `CameraProvider`

**No auto-discovery (deliberate).** Whether a GMSL port has a camera connected cannot be told from Video4Linux/sysfs — the device tree exposes every defined port as a `/dev/video` node regardless. The only reliable signals are an Argus capture attempt or the deserializer link-lock register over I2C (what the vendor driver uses), both too costly/board-specific for a generic scan. So cameras are registered explicitly; `sensor_id` maps 1:1 to `/dev/video<sensor_id>` and the device-tree module index.

### Test Plan

Automated (run locally, green): `uv run pytest tests/test_gmsl_camera.py` — command building (auto vs pinned exposure in ns, pinned gain, resolution/framerate), GDP caps parsing, `to_dict`/`from_dict` round-trip, non-operable provider. Existing `tests/test_cameras.py` / `tests/test_camera_persistence.py` still pass after the RTSP refactor.

Device validation on the target Jetson (hardware-gated test `test_gmsl_camera_capture` skips without hardware; set `GMSL_TEST_SENSOR_ID`):
- [x] **Live capture works** — produces a clean 1920×1200 ISP color image (good auto white balance) through the wide-angle lens; verified live in a browser, and two cameras render side-by-side via the provider.
- [x] **Exposure reaches the long-exposure target** — `nvarguscamerasrc` reports `1920×1200 @ 100fps, Exposure Range 15000–250000000 ns`; a pinned 0.25 s exposure captures and is visibly brighter than auto. ISP backend suffices, **no V4L2 fallback needed** (V4L2 raw also confirmed: `exposure` `15–250001` µs).

### Progress

---
⬛ claude-opus-4-8[1m]


NiklasNeugebauer:

**Verdict: approve once the draft questions are settled — the most polished PR in the queue.** Clean package layout, the `GDPPacket`/caps-parsing extraction into `rosys/vision/gstreamer.py` (shared with `RtspDevice`) is a genuine improvement, the self-healing capture loop matches the #423 pattern, and the tests are thoughtful for a hardware-bound feature (command construction, caps parsing, dict round-trip, provider guard, simulated capture; 8 passed, 1 skipped on the branch). Full suite green.

1. **Parameter changes eat the 2 s crash back-off.** `request_restart()` terminates the process, and the capture loop *always* sleeps `RECONNECT_DELAY = 2.0` between sessions — so a single exposure tweak costs ≥2 s of blackout plus terminate time. Distinguish intentional restarts (immediate rebuild) from crash back-off.
2. **Debounce is same-tick only.** `await asyncio.sleep(0)` batches synchronous setter bursts, but `_apply_all_parameters()` awaits each setter; with a slow terminate in flight the "task not done" guard mostly saves you, but it's timing-dependent. A short debounce window (e.g. 50 ms) would make batching deterministic.
3. `is_connected` is `self._capture_task is not None` — always `True` after construction, even if Argus fails forever. Same `is_connected`-semantics question raised on #423; worth aligning the contract across devices in one place.
4. Timestamps are receive-time (`rosys.time()`), while #408/#422 invest heavily in capture-time. Argus exposes sensor timestamps; worth a TODO so GMSL doesn't become the odd one out.
5. The `gstreamer.py` extraction will conflict with #422 (removes GDP usage from rtsp) and #408 (extends `GDPPacket` with PTS) — sequencing again.
6. The long-exposure warning (`exposure * fps > 1`) warns but proceeds — consider clamping fps automatically, since streak mode is the stated use case.

PR #421, opened 2026-06-22
Use the camera's capture timestamp for MJPEG frames

NiklasNeugebauer:
### Motivation

MJPEG `Image.time` is currently set to the client *receive* time (`rosys.time()`), which includes network and buffering delay and drifts from when the frame was actually captured. Cameras that stamp each frame at capture time expose that instant in the multipart part header, so we can report the true capture time instead — important for latency-sensitive and time-synchronised use.

### Implementation

- Add `parse_capture_timestamp()` to read the `X-Timestamp: <sec>.<usec>` field (absolute Unix epoch) from the MJPEG multipart part header that precedes each JPEG (the bytes before the `\xff\xd8` SOI marker).
- Use the parsed capture time as the image `timestamp`, falling back to `rosys.time()` when the field is absent or unparsable — so cameras without the header behave exactly as before.
- Yield the JPEG as `bytes` instead of a `memoryview` into the shared receive buffer. This also removes a latent aliasing hazard: the buffer is compacted in place on the next loop iteration while the consumer may still hold the view across an `await`.
- Add unit tests for the header parser (present, case-insensitive, absent, unparsable, last-wins).

### Progress

---
⬛ claude-opus-4-8[1m]

### DIFF
--- a/rosys/vision/mjpeg_camera/mjpeg_device.py
+++ b/rosys/vision/mjpeg_camera/mjpeg_device.py
@@ -11,6 +11,26 @@ from ..image_processing import remove_exif
 from .vendors import mac_to_url
 
 
+def parse_capture_timestamp(part_header: bytes) -> float | None:
+    """Extract the absolute capture instant (Unix epoch seconds) from the ``X-Timestamp``
+    field of an MJPEG multipart part header.
+
+    Cameras that stamp each frame at capture time emit an ``X-Timestamp: <sec>.<usec>``
+    line in the part header preceding the JPEG. Returns ``None`` when the field is absent
+    or unparsable, so the caller can fall back to the receive time.
+    """
+    marker = b'x-timestamp:'
+    index = part_header.lower().rfind(marker)
+    if index == -1:
+        return None
+    line_end = part_header.find(b'\r\n', index)
+    raw = part_header[index + len(marker):] if line_end == -1 else part_header[index + len(marker):line_end]
+    try:
+        return float(raw.strip())
+    except ValueError:
+        return None
+
+
 class MjpegDevice:
 
     def __init__(self, mac: str, ip: str, *,
@@ -50,7 +70,7 @@ class MjpegDevice:
     async def run_capture_task(self) -> None:
         self.log.debug('Starting capture task for %s', self._url)
 
-        async def stream() -> AsyncGenerator[bytes, None]:
+        async def stream() -> AsyncGenerator[tuple[bytes, float | None], None]:
             async with httpx.AsyncClient() as client:
                 assert self._url is not None
                 try:
@@ -84,8 +104,10 @@ class MjpegDevice:
                                 if start == -1:
                                     continue
 
+                                # the bytes before the SOI marker are this frame's multipart part header
+                                capture_time = parse_capture_timestamp(bytes(buffer_view[:start]))
                                 end += 2
-                                yield buffer_view[start:end]
+                                yield bytes(buffer_view[start:end]), capture_time
                                 buffer_view[:buffer_end - end] = buffer_view[end:buffer_end]
                                 buffer_end -= end
 
@@ -96,11 +118,11 @@ class MjpegDevice:
                     self.log.warning('Connection to %s failed. Was something disconnected?\n%s', self._url, e)
                     raise e
 
-        async for image in stream():
+        async for image, capture_time in stream():
             if not image:
                 continue
             try:
-                timestamp = rosys.time()
+                timestamp = capture_time if capture_time is not None else rosys.time()
                 result = self._on_new_image_data(remove_exif(image), timestamp)
                 if isinstance(result, Awaitable):
                     await result
--- a/tests/test_mjpeg_device.py
+++ b/tests/test_mjpeg_device.py
@@ -1,0 +1,26 @@
+from rosys.vision.mjpeg_camera.mjpeg_device import parse_capture_timestamp
+
+
+def test_parses_x_timestamp():
+    header = b'\r\n--boundary\r\nContent-Type: image/jpeg\r\nContent-Length: 1234\r\nX-Timestamp: 1718900000.123456\r\n\r\n'
+    assert parse_capture_timestamp(header) == 1718900000.123456
+
+
+def test_parses_x_timestamp_case_insensitively():
+    header = b'--boundary\r\nx-timestamp:1718900000.5\r\n\r\n'
+    assert parse_capture_timestamp(header) == 1718900000.5
+
+
+def test_returns_none_without_header():
+    header = b'\r\n--boundary\r\nContent-Type: image/jpeg\r\nContent-Length: 1234\r\n\r\n'
+    assert parse_capture_timestamp(header) is None
+
+
+def test_returns_none_for_unparsable_value():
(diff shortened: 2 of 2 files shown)


evnchn:
*Posted by Claude Code on @evnchn's behalf.*

**TL;DR:**

Took this for a spin as a second pair of eyes — stood up a local MJPEG server with `X-Timestamp` headers I controlled and drove the real `MjpegDevice.run_capture_task()` against it. **Behaves exactly as described:** capture time when present, host `rosys.time()` fallback when absent or malformed. LGTM.

One shared heads-up and one bit of credit:

- **Heads-up (shared with #422):** `image.time` is now in the camera's clock domain, while `get_recent_images()` compares it to host-wall-clock `rosys.time()` — worth a note on clock sync. Collapsed below; real but narrower than it first looks.
- **Credit:** the `memoryview` → `bytes` change is a genuine latent-bug fix independent of the feature — nice catch.

<details>
<summary><b>What I ran — end-to-end loopback reproducer + output (no camera)</b></summary>

`multipart/x-mixed-replace` server on `127.0.0.1`, four frames with controlled part headers, driving this branch's actual capture loop (the real `stream()` + `parse_capture_timestamp` integration, not just the pure parser):

```
captured 4 frames

scenario                         image.time        check
explicit capture time            1000000000.500    OK   == 1000000000.5
no X-Timestamp header            1782180938.503    OK   ~ host wall clock
malformed X-Timestamp            1782180938.555    OK   ~ host wall clock (fallback)
camera clock 1h SLOW (skew)      1782177338.451    OK   == host_now - 3600
```

The last row shows the camera value flows through **unbounded** (`image.time = host_now − 3600`) — the input to the clock-domain note below. Env: this branch, Python 3.13, host wall-clock mode (plain script, not pytest).

```python
"""Tier B — loopback MJPEG server + the real MjpegDevice.run_capture_task() on this branch."""
import asyncio, logging, time
from aiohttp import web
from rosys.vision.mjpeg_camera.mjpeg_device import MjpegDevice

FAKE_JPEG = b'\xff\xd8' + b'JPEGDATA' * 8 + b'\xff\xd9'  # SOI..EOI, payload free of ffd8/ffd9
FRAMES = [
    ('explicit capture time',       '1000000000.5'),
    ('no X-Timestamp header',       None),
    ('malformed X-Timestamp',       'not-a-number'),
    ('camera clock 1h SLOW (skew)', None),  # filled at runtime = host_now - 3600
]

async def mjpeg_handler(request):
    boundary = 'frameboundary'
    resp = web.StreamResponse(headers={'Content-Type': f'multipart/x-mixed-replace; boundary={boundary}'})
    await resp.prepare(request)
    host_now = time.time()
    for label, ts in FRAMES:
        if label.startswith('camera clock'):
            ts = f'{host_now - 3600:.3f}'
        header = f'--{boundary}\r\nContent-Type: image/jpeg\r\n'
        if ts is not None:
            header += f'X-Timestamp: {ts}\r\n'
        header += f'Content-Length: {len(FAKE_JPEG)}\r\n\r\n'
        await resp.write(header.encode() + FAKE_JPEG + b'\r\n')
        await asyncio.sleep(0.05)  # force each frame into its own chunk
    return resp

async def main():
    logging.basicConfig(level=logging.ERROR)
    app = web.Application(); app.router.add_get('/stream', mjpeg_handler)
    runner = web.AppRunner(app); await runner.setup()
    site = web.TCPSite(runner, '127.0.0.1', 0); await site.start()
    port = site._server.sockets[0].getsockname()[1]

    captured = []
    dev = MjpegDevice.__new__(MjpegDevice)            # bypass mac->url + on_startup machinery
    dev._url = f'http://127.0.0.1:{port}/stream'
    dev._authentication = None
    dev.log = logging.getLogger('test.mjpeg')
    dev._capture_task = None
    dev._on_new_image_data = lambda data, ts: captured.append((data, ts))

    host_before = time.time()
    await dev.run_capture_task()                      # returns when the server closes the stream
    host_after = time.time()
    await runner.cleanup()

    expectations = [('explicit capture time', 1000000000.5), ('no X-Timestamp header', 'wallclock'),
                    ('malformed X-Timestamp', 'wallclock'), ('camera clock 1h SLOW (skew)', host_before - 3600)]
    for (label, exp), (_, ts) in zip(expectations, captured):
        if exp == 'wallclock':
            ok = host_before - 2 <= ts <= host_after + 2
        else:
            ok = abs(ts - exp) < 0.5
        print(f'{label:32} {ts:18.3f}  {"OK" if ok else "FAIL"}')

asyncio.run(main())
```
</details>

<details>
<summary><b>Clock-domain heads-up (shared with #422) — full reproducer + the walk-back</b></summary>

`Camera.get_recent_images()` filters `i.time > current_time - timespan` with `current_time` defaulting to host `rosys.time()` (`rosys/vision/camera/camera.py:142-144`). With `image.time` now on the camera clock, an un-synced camera shifts the recency window:

```
get_recent_images(current_time=HOST_NOW, timespan=10s)
  camera clock 1h SLOW, frame JUST arrived     -> DROPPED (looks stale)
  camera clock 1h FAST, frame is genuinely OLD -> KEPT  (looks fresh forever)
```

I checked whether this reaches the odometer and it does **not** in core (`Odometer.handle_detection` takes a `Pose`, not an `Image` — no `image.time → Pose.time` link), so the confirmed surface is just `get_recent_images`. Suggestion: a line documenting the NTP-sync assumption, or bounding the skew. Not a blocker.

```python
"""Tier A — clock-domain demo; drives the REAL Camera.get_recent_images() with skewed image.time."""
from collections import deque
import numpy as np
from rosys.vision.camera.camera import Camera
from rosys.vision.image import Image

class FakeCamera(Camera):  # skip nicegui route registration; inherit the real method under test
    def __init__(self):
        self.id = 'demo'
        self.images = deque(maxlen=16)

def img(t):
    return Image.from_array(np.zeros((8, 8, 3), np.uint8), camera_id='demo', time=t)

HOST_NOW, TIMESPAN = 1_000_000_000.0, 10.0
rows = [
    ('synced clock,   2s ago   (genuinely fresh)', HOST_NOW - 2,    'keep'),
    ('synced clock,  30s ago   (genuinely stale)', HOST_NOW - 30,   'drop'),
    ('camera clock 1h SLOW, frame JUST arrived',   HOST_NOW - 3600, 'keep'),
    ('camera clock 1h FAST, frame is genuinely OLD', HOST_NOW + 3600, 'drop'),
]
cam = FakeCamera()
for _, t, _ in rows:
    cam.images.append(img(t))
kept = {round(i.time, 3) for i in cam.get_recent_images(current_time=HOST_NOW, timespan=TIMESPAN)}
for label, t, should in rows:
    actual = 'keep' if round(t, 3) in kept else 'drop'
    print(f'{label:48} {t:14.1f} should={should} actual={actual} {"" if actual==should else "<<< SKEW"}')
```
</details>

Nothing blocking from my side — already approved by @ftilde and CI-green; just adding the empirical confirmation + the shared heads-up.

NiklasNeugebauer:
Maybe we should consider some general documentation about camera timestamping (rtsp has the same mechanism)

ftilde (approved):
I've tested it on a robot and think it's good as is. Maybe we can just merge it already and add the docs with the rtsp-PR?

PR #422, opened 2026-06-22
Use the camera's capture timestamp for RTSP frames

NiklasNeugebauer:
### Motivation

RTSP `Image.time` is currently set to the client *receive* time (`rosys.time()`), which includes network and decode delay and drifts from when the frame was actually captured. Cameras that timestamp frames at capture time carry that instant per frame in the WebRTC "Absolute Capture Time" RTP header extension, so we can report the true capture time — important for latency-sensitive and time-synchronised use.

### Implementation

- Add `capture_time.py` with the WebRTC abs-capture-time codec (Q32.32 NTP ↔ Unix epoch) and `read_abs_capture_time()` to read the one-byte RTP header extension from a mapped `GstRtp.RTPBuffer`.
- Read the per-frame capture instant in `RtspDevice` and use it as the image `timestamp`, falling back to `rosys.time()` when the extension is absent — cameras without it behave exactly as before.
- The absolute capture time only exists at the RTP layer (lost after depay), so replace the `gst-launch-1.0 ... ! gdppay ! fdsink` subprocess with an **in-process** GStreamer pipeline: a pad probe on the depayloader sink pad reads the extension and maps it by GStreamer buffer PTS; the `appsink`'s decoded RGB frames are pulled in an asyncio executor and matched to their capture time by PTS.
- Load the GStreamer Python bindings (PyGObject `Gst`/`GstRtp`) lazily and cache the result, so the bindings are only required for RTSP and a deployment without them reports the missing dependency once rather than on every reconnect.
- Preserve the full public surface of `RtspDevice` (`url`, `shutdown`, `restart_gstreamer`, `set/get_fps`, `set/get_substream`, `set/get_bitrate`, `get/set_avdec`, `is_connected`, `authorized`).
- Add `gi` to mypy's `ignore_missing_imports`; add unit tests for the abs-capture-time codec and extension reader.

### New runtime requirement

RTSP capture now needs the **GStreamer Python bindings** (PyGObject with `Gst` and `GstRtp` typelibs) in-process, where it previously only needed the `gst-launch-1.0` binary. The import is lazy and cached, so non-RTSP users are unaffected.

### Work In Progress

- Not yet verified against a physical camera + GStreamer (no hardware in this environment). The codec and parsing are unit-tested; the pipeline/threading/lifecycle logic has been reviewed but needs a live RTSP smoke test (frame flow, capture-time vs. fallback, reconnect, shutdown) before marking ready.

### Progress

---
⬛ claude-opus-4-8[1m]

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,6 +123,8 @@ follow_imports = "skip"
 module = [
     "cairosvg",
     "coloredlogs",
+    "gi",
+    "gi.*",
     "imgsize",
     "networkx",
     "objgraph",
--- a/rosys/vision/rtsp_camera/capture_time.py
+++ b/rosys/vision/rtsp_camera/capture_time.py
@@ -1,0 +1,46 @@
+"""Recover the absolute capture instant of a frame from an RTP header extension.
+
+Cameras that stamp frames at capture time carry the instant in the WebRTC
+"Absolute Capture Time" RTP header extension: a 64-bit Q32.32 NTP timestamp
+(seconds since 1900-01-01, big-endian) on the first packet of each frame.
+Unlike the RTP timestamp it is self-anchored -- each frame carries a freshly
+clock-corrected absolute time -- so it needs neither an RTCP Sender Report
+anchor nor extrapolation, while the RTP timestamp stays a clean media clock.
+
+See http://www.webrtc.org/experiments/rtp-hdrext/abs-capture-time.
+"""
+from __future__ import annotations
+
+from typing import Any
+
+ABS_CAPTURE_TIME_URI = 'http://www.webrtc.org/experiments/rtp-hdrext/abs-capture-time'
+
+# Seconds between the 1900 (NTP) and 1970 (Unix) epochs.
+_NTP_UNIX_EPOCH_OFFSET = 2_208_988_800
+
+# Default one-byte header-extension id for abs-capture-time. The id is properly negotiated
+# via the SDP `a=extmap` line; until that is parsed the extension is read at this fixed id.
+DEFAULT_ABS_CAPTURE_TIME_EXT_ID = 10
+
+
+def decode_abs_capture_time(payload: bytes) -> float:
+    """Decode the extension payload (>=8 bytes; trailing optional fields ignored) to Unix epoch seconds."""
+    return int.from_bytes(payload[:8], 'big') / (1 << 32) - _NTP_UNIX_EPOCH_OFFSET
+
+
+def encode_abs_capture_time(epoch: float) -> bytes:
+    """Encode a Unix epoch (seconds) as the extension's 8-byte Q32.32 NTP payload (inverse of decode)."""
+    q32_32 = round((epoch + _NTP_UNIX_EPOCH_OFFSET) * (1 << 32)) & ((1 << 64) - 1)
+    return q32_32.to_bytes(8, 'big')
+
+
+def read_abs_capture_time(rtp_buffer: Any, ext_id: int = DEFAULT_ABS_CAPTURE_TIME_EXT_ID) -> float | None:
+    """Read the abs-capture-time epoch (s) from a mapped ``GstRtp.RTPBuffer`` one-byte header extension.
+
+    Returns ``None`` when the extension is absent, so the caller can fall back to the receive time.
+    ``rtp_buffer`` is kept untyped to avoid importing the GStreamer bindings at module import time.
+    """
+    ok, data = rtp_buffer.get_extension_onebyte_header(ext_id, 0)
+    if not ok or not data:
+        return None
+    return decode_abs_capture_time(bytes(data))
--- a/rosys/vision/rtsp_camera/rtsp_device.py
+++ b/rosys/vision/rtsp_camera/rtsp_device.py
@@ -1,25 +1,41 @@
 from __future__ import annotations
 
 import asyncio
+import functools
 import logging
-import re
-import shlex
-import signal
-import struct
-import subprocess
-from asyncio.subprocess import Process
-from collections.abc import AsyncGenerator, Awaitable, Callable
-from dataclasses import dataclass
-from enum import Enum
-from typing import Literal
+import threading
+from collections.abc import Awaitable, Callable
+from typing import Any, Literal
 
 import numpy as np
 from nicegui import background_tasks
 
 from ... import rosys
 from ...vision.image import ImageArray
+from .capture_time import DEFAULT_ABS_CAPTURE_TIME_EXT_ID, read_abs_capture_time
 from .jovision_rtsp_interface import JovisionInterface
 from .vendors import VendorType, mac_to_url, mac_to_vendor
+
+
+@functools.lru_cache(maxsize=1)
+def _load_gstreamer() -> Any:
+    """Import and initialize the GStreamer Python bindings (PyGObject with ``Gst`` and ``GstRtp``).
+
+    Returns the ``(Gst, GstRtp)`` modules, or ``None`` if the bindings are missing. The result is
+    cached, so a deployment without them does not re-run the import or log the error on every
+    reconnect attempt; it is reported once.
+    """
+    try:
+        import gi  # pylint: disable=import-outside-toplevel  # noqa: PLC0415
+        gi.require_version('Gst', '1.0')
+        gi.require_version('GstRtp', '1.0')
+        from gi.repository import Gst, GstRtp  # pylint: disable=import-outside-toplevel  # noqa: PLC0415
+        Gst.init(None)
+        return (Gst, GstRtp)
+    except (ImportError, ValueError) as e:
+        logging.getLogger('rosys.vision.rtsp_camera.rtsp_device').error(
+            'RTSP capture requires the GStreamer Python bindings (PyGObject with Gst and GstRtp): %s', e)
+        return None
 
 
 class RtspDevice:
@@ -35,9 +51,11 @@ class RtspDevice:
         self._substream = substream
         self._on_new_image_data = on_new_image_data
         self._avdec = avdec
+        self._ext_id = DEFAULT_ABS_CAPTURE_TIME_EXT_ID
 
         self._capture_task: asyncio.Task | None = None
-        self._capture_process: Process | None = None
+        self._pipeline: Any = None
+        self._gst: Any = None
         self._authorized: bool = True
 
         vendor_type = mac_to_vendor(mac)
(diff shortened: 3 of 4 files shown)


evnchn:
*Posted by Claude Code on @evnchn's behalf.*

**TL;DR:**

I took your "needs a live smoke test" note as an invite to poke at this — ran the decoder against real `GstRtp.RTPBuffer`s end to end. **The decode + fallback logic is correct.** Two things worth a look before it ships:

1. **The abs-capture-time extmap id is hardcoded to `10`.** A camera that negotiates a *different* id via SDP `a=extmap` is silently treated as having no capture time (falls back to receive-time, no log). Reproduced below.
2. **Heads-up shared with #421:** `image.time` now lives in the camera's clock domain, and `get_recent_images()` compares it against host-wall-clock `rosys.time()`. Real but narrow — I checked, and the scarier "wrong odometer pose" version does *not* hold in core.

Not a blocker from my side — just second-pair-of-eyes evidence to fold in. Everything below is runnable.

<details>
<summary><b>What I ran — full reproducer + output (and what it does NOT prove)</b></summary>

Per your own research path, the read is just `RTPBuffer.get_extension_onebyte_header(ext_id, 0)`, so a transport-free in-process RTP buffer exercises the exact path your pad probe uses — no camera, no RTSP, no sockets. This drives **this branch's actual `capture_time.py`** against real GStreamer RTP buffers built with `add_extension_onebyte_header`.

```
1. encode/decode NTP Q32.32 round-trip (your codec):
  [OK] epoch 1000000000.5        -> 1000000000.5
  [OK] epoch 1700000000.123456   -> 1700000000.123456
  [OK] epoch 0.0                 -> 0.0

2. real RTPBuffer, extension at id=10 (matches the hardcode):
  [OK] capture time recovered    -> 1700000123.456

3. real RTPBuffer, extension at id=11 but read at hardcoded id=10:
  [OK] -> None  (silent fallback to receive time)   <-- the finding

4. real RTPBuffer with NO extension:
  [OK] -> None  (no regression; caller uses rosys.time())
```

**Honest caveat:** this proves the **decode + id-handling** logic, not that production `rtspsrc` surfaces the extension on the depay sink pad. It *narrows* the live smoke test you flagged, it doesn't replace it. Env: GStreamer 1.28.2 + PyGObject + `GstRtp` on macOS.

```python
"""Tier C — drives this branch's capture_time.py against real GstRtp RTP buffers.
Run under a python with PyGObject/GstRtp; loads capture_time.py from the branch checkout."""
import importlib.util
import sys
from pathlib import Path

import gi
gi.require_version('Gst', '1.0')
gi.require_version('GstRtp', '1.0')
from gi.repository import Gst, GstRtp

CT_PATH = Path.home() / 'rosys-review/rosys/vision/rtsp_camera/capture_time.py'
spec = importlib.util.spec_from_file_location('capture_time', CT_PATH)
ct = importlib.util.module_from_spec(spec); spec.loader.exec_module(ct)
Gst.init(None)

PR_EXT_ID = ct.DEFAULT_ABS_CAPTURE_TIME_EXT_ID  # 10, hardcoded by the PR
fails = 0

def check(label, cond, detail=''):
    global fails
    if not cond:
        fails += 1
    print(f'  [{"OK" if cond else "FAIL"}] {label}{("  " + detail) if detail else ""}')

print('\n1. encode/decode NTP Q32.32 round-trip (the PR codec, no GStreamer):')
for epoch in (1_000_000_000.5, 1_700_000_000.123456, 0.0):
    back = ct.decode_abs_capture_time(ct.encode_abs_capture_time(epoch))
    check(f'epoch {epoch}', abs(back - epoch) < 1e-3, f'-> {back}')

def make_rtp_with_ext(ext_id, epoch):
    buf = GstRtp.RTPBuffer.new_allocate(0, 0, 0)
    ok, rtp = GstRtp.RTPBuffer.map(buf, Gst.MapFlags.READ | Gst.MapFlags.WRITE)
    assert ok
    rtp.add_extension_onebyte_header(ext_id, ct.encode_abs_capture_time(epoch))
    return rtp, buf

print(f'\n2. real GstRtp.RTPBuffer, extension at id={PR_EXT_ID} (matches PR hardcode):')
EPOCH = 1_700_000_123.456
rtp, _ = make_rtp_with_ext(PR_EXT_ID, EPOCH)
got = ct.read_abs_capture_time(rtp, ext_id=PR_EXT_ID)
check('feature works: capture time recovered', got is not None and abs(got - EPOCH) < 1e-3, f'-> {got}')
rtp.unmap()

print(f'\n3. real RTP buffer, extension at id=11 but PR reads hardcoded id={PR_EXT_ID}:')
rtp, _ = make_rtp_with_ext(11, EPOCH)
got = ct.read_abs_capture_time(rtp, ext_id=PR_EXT_ID)
check('id-mismatch -> None (silent fallback to receive time)', got is None,
      'a camera negotiating a different extmap id is silently treated as having no capture time')
rtp.unmap()

print('\n4. real RTP buffer with NO extension (camera without the feature):')
buf = GstRtp.RTPBuffer.new_allocate(0, 0, 0)
ok, rtp = GstRtp.RTPBuffer.map(buf, Gst.MapFlags.READ | Gst.MapFlags.WRITE)
got = ct.read_abs_capture_time(rtp, ext_id=PR_EXT_ID)
check('absent -> None (no regression; caller uses rosys.time())', got is None)
rtp.unmap()

print('ALL CHECKS PASSED' if fails == 0 else f'{fails} CHECK(S) FAILED')
sys.exit(1 if fails else 0)
```
</details>

<details>
<summary><b>Finding: hardcoded <code>ext_id=10</code> → silent fallback on id mismatch</b></summary>

Minimal repro (scenario 3 above, distilled):

```python
from gi.repository import Gst, GstRtp
import capture_time as ct  # this branch's rosys/vision/rtsp_camera/capture_time.py
Gst.init(None)

buf = GstRtp.RTPBuffer.new_allocate(0, 0, 0)
ok, rtp = GstRtp.RTPBuffer.map(buf, Gst.MapFlags.READ | Gst.MapFlags.WRITE)
rtp.add_extension_onebyte_header(11, ct.encode_abs_capture_time(1_700_000_123.456))  # camera uses id 11

print(ct.read_abs_capture_time(rtp, ext_id=10))  # -> None  (capture time silently lost)
```

Your own docstring already calls this out (`DEFAULT_ABS_CAPTURE_TIME_EXT_ID`, *"properly negotiated via the SDP `a=extmap` line; until that is parsed the extension is read at this fixed id"*). Flagging it because the degradation is *silent* — a camera on a different id looks identical to a camera without the feature. Three ways to handle, in rough order of effort: (a) parse the `a=extmap` from the SDP; (b) log-once when the hardcoded id misses but other extensions are present; (c) note it as a known limitation in the PR body.
</details>

<details>
<summary><b>Clock-domain heads-up (shared with #421) — full reproducer + the walk-back</b></summary>

`Camera.get_recent_images()` filters `i.time > current_time - timespan` with `current_time` defaulting to host-wall-clock `rosys.time()` (`rosys/vision/camera/camera.py:142-144`, `rosys.py:107`). Once `image.time` is the camera's capture clock, an un-NTP-synced camera shifts which frames count as "recent":

```
get_recent_images(current_time=HOST_NOW, timespan=10s)
  camera clock 1h SLOW, frame JUST arrived     -> DROPPED (looks stale)
  camera clock 1h FAST, frame is genuinely OLD -> KEPT  (looks fresh forever)
```

**What I checked and walked back:** I initially worried this corrupts the odometer, but `Odometer.handle_detection(detection: Pose)` takes a localization `Pose`, not a vision `Image` — there is no `image.time → Pose.time` link in core, so that path is *not* affected unless app-level localization wires it up. (Independently re-verified, incl. a fresh-lineage cross-check.) So the confirmed surface is `get_recent_images` skew. Suggestion: document the NTP-sync assumption, or bound/warn on large `|capture_time − rosys.time()|`.

```python
"""Tier A — clock-domain demo; drives the REAL Camera.get_recent_images() with skewed image.time."""
from collections import deque
import numpy as np
from rosys.vision.camera.camera import Camera
from rosys.vision.image import Image

class FakeCamera(Camera):  # skip nicegui route registration; inherit the real method under test
    def __init__(self):
        self.id = 'demo'
        self.images = deque(maxlen=16)

def img(t):
    return Image.from_array(np.zeros((8, 8, 3), np.uint8), camera_id='demo', time=t)

HOST_NOW, TIMESPAN = 1_000_000_000.0, 10.0
rows = [
    ('synced clock,   2s ago   (genuinely fresh)', HOST_NOW - 2,    'keep'),
    ('synced clock,   5s ago   (genuinely fresh)', HOST_NOW - 5,    'keep'),
    ('synced clock,  30s ago   (genuinely stale)', HOST_NOW - 30,   'drop'),
    ('camera clock 1h SLOW, frame JUST arrived',   HOST_NOW - 3600, 'keep'),
    ('camera clock 1h FAST, frame is genuinely OLD', HOST_NOW + 3600, 'drop'),
]
cam = FakeCamera()
for _, t, _ in rows:
    cam.images.append(img(t))
kept = {round(i.time, 3) for i in cam.get_recent_images(current_time=HOST_NOW, timespan=TIMESPAN)}
for label, t, should in rows:
    actual = 'keep' if round(t, 3) in kept else 'drop'
    print(f'{label:48} {t:14.1f} should={should} actual={actual} {"" if actual==should else "<<< SKEW"}')
```
</details>

<details>
<summary><b>Minor</b></summary>

- Per-frame fallback can intermix capture-clock and host-clock stamps within one stream (when a PTS lookup misses) — a consistent per-stream fallback would avoid a mixed-domain `image.time` series.
- `asyncio.get_event_loop()` inside the running coroutine → `get_running_loop()` (soft-deprecated on 3.12/3.13).
</details>

Defer entirely to your cadence on the live-camera test.

NiklasNeugebauer:
Still unsure if the GI integration works/ is needed/ is helpful generally. I will look at this again in a couple of days.

NiklasNeugebauer:

**Verdict: right long-term direction (standards-based abs-capture-time), but it's a real reliability tradeoff that needs team alignment, and it currently conflicts with main.** The codec is clean and all 5 codec tests pass on the branch (the fake-`RTPBuffer` trick to test without GStreamer is nice).

1. **Strategic: this PR and #408 solve the same problem in opposite ways** — #408 reads a vendor-specific NTP-in-RTP-timestamp (divinus) with a hardcoded per-resolution calibration constant, this one reads the standard WebRTC `abs-capture-time` RTP header extension. Pick one; I'd pick this one (works with any compliant camera, no calibration constant in framework code) and add the extension on the camera firmware side. #408 could then be parked/closed.
2. **Process-isolation regression:** moving from a `gst-launch` subprocess to in-process PyGObject means a decoder crash (avdec on a corrupted stream) now takes down the whole robot process instead of one camera. Notably, #425/#428 move MJPEG *out* of process partly for isolation while this moves RTSP *into* the process — the team should explicitly own this asymmetry, or plan to run the Gst pipeline in the same kind of capture subprocess.
3. **PyGObject/GstRtp is a new runtime dependency that is declared nowhere** — the diff only adds a mypy ignore for `gi`. No `pyproject` extra, no Dockerfile change, no docs. On a deployment without the bindings, RTSP silently stops working with a single log line. At minimum document it and fail loudly in `RtspCamera.connect()`.
4. **`DEFAULT_ABS_CAPTURE_TIME_EXT_ID = 10` is a silent interop trap.** The id is negotiated per-stream via SDP `a=extmap`; a camera using any other id silently degrades to receive-time stamps with no signal. Until SDP parsing exists, make the id a per-camera constructor parameter and log once when the extension is expected but never seen.
5. **Clock-domain mixing:** frames with the extension get camera-NTP time, frames without fall back to `rosys.time()` — with an unsynced camera clock these can differ by seconds and `Image.time` will jump between clock domains frame-to-frame. Suggest a sanity clamp (warn if |capture − receive| exceeds a threshold) and not mixing within a session once the extension has been seen.
6. **Executor usage:** the pull loop parks a default-executor thread per camera indefinitely (`run_in_executor(None, sink.try_pull_sample, 1 s)` in a loop). N cameras = N mostly-blocked threads competing with everything else on the default executor. A dedicated thread per device (or rosys's io pool) is safer.
7. `shutdown()` calls `pipeline.set_state(NULL)` on the event loop thread — Gst state changes can block while streaming threads join; push it to an executor.
8. Buffer-PTS matching between the depay pad probe and the appsink relies on PTS preservation through decode/convert — true for these pipelines, but a one-line comment on the B-frame/PTS assumption would help future readers. The bounded `capture_by_pts` map with a lock is correct.
9. **`mergeable: dirty`** — real conflicts with current main; needs a rebase, and see the sequencing note in the #423 review.

PR #423, opened 2026-06-25
Reconnect cameras automatically at the device level

NiklasNeugebauer:
### Motivation

Cameras lose their connection due to network glitches, a bad cable or a power hiccup.
Today a dropped connection ends the device's capture task for good, so the only way back is a full `CameraProvider` scan — which is heavy and, with `auto_scan=False`, may also surface cameras we deliberately don't want connected (common in test/development setups).

This is an alternative to #385: instead of wrapping reconnection around the `Camera` base class (which forced a separate `activate()`/`deactivate()` API and raised serialization and race-condition concerns), reconnection lives **inside each device**, where the connection already lives. Providers are no longer involved in reconnection at all.

### Implementation

Every camera device keeps itself connected for as long as the camera is connected, retrying every `reconnect_interval` seconds (default `3.0`). All three real devices now share one self-healing loop pattern (`_run_capture_loop`): run one capture/streaming session, and when it ends or fails, wait `reconnect_interval` and reconnect, until `shutdown()` cancels the task.

- **`RtspDevice` / `MjpegDevice`** — each capture session is one streaming session. Both stop retrying when the camera rejects our credentials (RTSP on an `Unauthorized` gstreamer error, MJPEG on an HTTP `401`) instead of hammering the endpoint forever.
- **`UsbDevice`** — the capture task is the same self-healing loop: `_run_capture_session` reads frames until the capture is lost (failed reads past `MAX_READ_FAILURES` / closed handle, e.g. an unplugged cable), then the loop waits `reconnect_interval` and re-opens the device — re-resolving the `/dev/video*` node from the camera uid (which can change on re-plug) and restoring resolution/fps. This replaces the previous `on_repeat` polling with manual `_next_reconnect_time` rate-limiting, and lets successful reads run at the camera's native rate instead of being capped by the poll interval.
- **`SimulatedDevice`** — gains `disconnect()` to simulate a cable drop and reconnects itself after `reconnect_interval`; a `simulate_failing` flag drops the connection randomly over time. `shutdown()` stops the loop.

**Connection-state contract (uniform across all devices and cameras):**

- `is_connected` — the camera/device is **actually streaming/capturing right now**.
- `is_active` — the self-healing loop is alive (streaming **or** waiting to reconnect); `connect()` idempotency, the providers, and parameter caching all key off this.
- `is_reconnecting` — `is_active and not is_connected`.

This replaces the earlier, inconsistent meaning where "connected" variously meant "loop alive" (RTSP/MJPEG) or "actually connected" (USB/simulated), so a dead RTSP camera no longer shows as connected forever in a UI.

**Supporting changes:**

- Keep `connect()`/`disconnect()` semantics intact — no new public method. `connect()` starts a self-healing connection (tearing down any stale device first); `disconnect()` stops the retries and tears it down.
- Add a `reconnect_interval` parameter to every camera, thread it to the device, and persist it in `to_dict()` (restored via the base passthrough). `SimulatedCamera` also takes `simulate_failing`.
- **Drop the `SimulatedCameraProvider`'s disconnect/rescan simulation** — the failure now lives in the device, so providers are not needed to recover a connection; `simulate_failing` is forwarded to the cameras the provider creates. The old `simulate_device_failure` attribute is kept as a deprecated alias (emits `DeprecationWarning`) for one release.
- Register `rosys.on_shutdown(self.disconnect)` **once per camera** (in the `Camera` base) rather than per device, so devices — which are recreated on every reconnect cycle — no longer leak shutdown handlers, while cameras created without a provider are still torn down cleanly.
- Guard `RtspDevice._run_gstreamer` with a `finally` that terminates and clears the gstreamer process, preventing an orphaned process and a reconnect wedge when a frame callback raises.

### Tests

`tests/test_camera_reconnect.py` covers the behavior without hardware:
- `MjpegDevice` against a local "flaky" HTTP server that drops the connection after one frame — reconnects and keeps receiving frames, stops once `shutdown()` is called, and gives up (no further attempts) after an HTTP `401`.
- `RtspDevice` reconnect loop (gstreamer session stubbed) reconnects repeatedly, stops after `shutdown()`, and keeps `is_active` true while `is_connected` is false during back-off.
- `UsbDevice` against a fake capture — re-opens the device after a simulated cable drop, resumes frames, and stops reopening after `shutdown()`.
- `SimulatedDevice` resumes image generation after `disconnect()` once `reconnect_interval` has elapsed; `camera.disconnect()` stops its loop and no further images arrive.
- `reconnect_interval` (and `simulate_failing`) is persisted/threaded through to the device for all camera types; the deprecated `simulate_device_failure` alias warns; and devices do not leak shutdown handlers across repeated connect/disconnect cycles.

### Progress

---
⬛ claude-opus-4-8[1m]


NiklasNeugebauer:

**Verdict: this is the right architecture — prefer it over #385 and close #385.** Reconnection lives where the connection lives; providers stay out of it; no new public `activate()/deactivate()` surface; per-device `reconnect_interval` is persisted (tests cover this). All 11 new tests pass and the full suite is green on the branch. I also verified end-to-end with a real `MjpegCamera` against a local fake server that drops the stream every ~1 s: the device autonomously opened 5 sessions in 9 s with `reconnect_interval=1.0` — reconnection works without any provider involvement.

Issues to address before undrafting:

1. **`rosys.on_shutdown(self.shutdown)` in device constructors leaks device instances.** `on_shutdown` appends the bound method to a module-level list that is never pruned, and devices are recreated on every camera `connect()` cycle. Over a long-running deployment with flaky cameras that list grows without bound, and every dead device stays referenced (its logger, callbacks, and for `UsbDevice` a `__del__` that will now never run before process exit). Ironic given #427 exists precisely for this pattern. Options: register once per class, unregister in `shutdown()`, or make `on_shutdown` weak-capable via #427's `_weaken`.
2. **`is_connected` semantics are now inconsistent and partly misleading.** For `RtspDevice`/`MjpegDevice`, "connected" now means "the reconnect loop task is alive" — `True` even while the camera is unreachable and the loop is backing off. `SimulatedDevice.is_connected` and `UsbDevice.is_connected` report the *actual* connection state. Anything downstream that surfaces `camera.is_connected` in a UI will show a dead RTSP camera as connected forever. Suggest a uniform contract: `is_connected` = actually streaming; expose the loop via something like `is_active`/`is_reconnecting`.
3. **MJPEG never gives up on 401.** RTSP breaks the loop on `Unauthorized`; the MJPEG loop logs the 401 and retries every 3 s forever, hammering the camera (and lockout-prone auth endpoints) with bad credentials. Mirror the RTSP behavior, or add exponential backoff.
4. `SimulatedCameraProvider.simulate_device_failure` → `simulate_failing` is a breaking attribute rename; consider deprecating the old name for one release in case anyone toggles it at runtime.
5. Coordination: conflicts are guaranteed with #425/#428 (mjpeg capture loop moves into a subprocess) and #422 (rtsp `_run_gstreamer` rewrite). Since this PR's loop pattern is what `GmslDevice` (#420) already ships, a natural sequence is: #423 (pattern) → #425/#428 (mjpeg process) → #422 (rtsp in-process), resolving as you go.

NiklasNeugebauer:

Some observations that span the current PR queue:

1. **Landing order matters more than any single review point.** Suggested: #427 (independent) → #423 (reconnect pattern; supersedes #385) → #400 (rebased, folded into the mjpeg path) → #425 + #428 together (mjpeg subprocess + memfd) → #420 → #422 (rtsp timing; supersedes #408) — with #385, #408 and #276 closed or parked.
2. **Define `is_connected` semantics once** ("actually streaming" vs. "reconnect loop alive") and apply them to Rtsp/Mjpeg/Usb/Simulated/Gmsl uniformly; three open PRs currently redefine it in three different ways.
3. **Process-isolation policy:** MJPEG is moving *out* of process (#425/#428) while RTSP is moving *into* the process (#422). Both have good local reasons; pick a target architecture — a shared capture-subprocess mechanism would serve both and could host #422's RTP pad probe.
4. Independent of these PRs: `tests/test_recorder_replay.py::test_record_images` fails on current `main` in a plain Linux container (the JPEG never lands in `recordings_dir` within the forwarded window; possibly TurboJPEG-absence/timing related) — worth a separate look.

rosys/vision/camera/camera.py
```diff
@@ -100,9 +101,19 @@ def from_dict(cls, data: dict[str, Any]) -> Self:
...
     @property
     def is_connected(self) -> bool:
-        """To be interpreted as "ready to capture images"."""
+        """Whether the camera is currently streaming or capturing images."""
         return False
 
+    @property
+    def is_active(self) -> bool:
+        """Whether the camera keeps itself connected (streaming or waiting to reconnect)."""
+        return self.is_connected
```
ftilde, line 124: Is that the right default? When we don't pass `connect_after_init` this seems wrong.

Do we even want a default?

rosys/vision/usb_camera/usb_device.py
```diff
@@ -83,29 +108,74 @@ def create_capture(index: int) -> cv2.VideoCapture | None:
...
+            self._capture = None
 
-    async def release_capture(self) -> None:
-        await rosys.run.io_bound(self._capture.release)
+    async def shutdown(self) -> None:
+        self._should_run = False
+        self._capture_task.stop()
+        await self._release()
+
+    def _reapply_settings(self) -> None:
```
ftilde: Is there a reason we are only applying these settings and not exposure/auto-exposure?

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -53,19 +55,55 @@ def __init__(self, mac: str, ip: str, *,
...
 
     @property
     def is_connected(self) -> bool:
+        """Whether the MJPEG stream is currently delivering frames."""
+        return self._streaming
+
+    @property
+    def is_active(self) -> bool:
+        """Whether the self-healing capture loop is alive (streaming or waiting to reconnect)."""
         return (self._capture_task is not None) and (not self._capture_task.done())
```
ftilde: When auth fails, `_capture_task` will be None/ended. Do we want that to count as active or not? `update_device_list` of the provider will recreate the device (periodically), which contradicts the intent of not trying to reconnect when auth fails. I'm not sure what we want here.

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -53,19 +55,55 @@ def __init__(self, mac: str, ip: str, *,
         if url is None:
             raise ValueError(f'could not determine URL for {mac}')
         self._url = url
+        self.reconnect_interval = reconnect_interval
+        self._should_run: bool = True
```
ftilde: I find it very hard to keep track of which combination of `_should_run`, `_streaming` and `_authorized` is valid. Can we combine the state into an enum where each state is actually valid? Should reduce errors.

rosys/vision/camera/configurable_camera.py
```diff
@@ -48,7 +48,7 @@ def _register_parameter(self, name: str, getter: Callable[[], Any], setter: Call
         self._parameters[name] = Parameter(info=info, getter=getter, setter=setter, value=default_value)
 
     async def _apply_parameters(self, new_values: dict[str, Any], force_set: bool = False) -> None:
-        if not self.is_connected:
+        if not self.is_active:
```
ftilde: I *think* this should still be `is_connected`. Otherwise we cannot write parameters, right?

ftilde: If that's the case, we should also document it, i.e. that we can only write params when the cam is connected. that should be the case for network-cams, at least.

rosys/vision/camera/camera.py
```diff
@@ -53,6 +53,7 @@ def __init__(self,
 
         if connect_after_init:
             rosys.on_startup(self.connect)
+        rosys.on_shutdown(self.disconnect)
```
ftilde, line 60: Claude notes that we leak cameras here. Not sure if we care.

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -53,19 +55,55 @@ def __init__(self, mac: str, ip: str, *,
...
+    @property
+    def is_active(self) -> bool:
+        """Whether the self-healing capture loop is alive (streaming or waiting to reconnect)."""
         return (self._capture_task is not None) and (not self._capture_task.done())
 
+    @property
+    def authorized(self) -> bool:
+        return self._authorized
+
     def start_capture_task(self) -> None:
```
ftilde: The semantics are somewhat confusing here. `start_capture_task` actually spawns a `run_capture_loop` which then calls `run_capture_task`.

Maybe `...loop` should the new `...task` (since it's the actual task), and `run_capture_task` can be something like `stream_until_disconnect` or something?

rosys/vision/usb_camera/usb_device.py
```diff
@@ -83,29 +108,74 @@ def create_capture(index: int) -> cv2.VideoCapture | None:
         return capture
 
     async def _capture_image(self) -> None:
-        if not self._capture.isOpened():
+        if not self._should_run:
+            return
+        if self._capture is None or not self._capture.isOpened():
+            await self._reconnect()
             return
```
ftilde: Why do we return here instead of trying to capture an image?

ftilde: Hm, I guess it's fine since we try to capture with 100Hz.

ftilde (changes requested):

rosys/vision/camera/camera.py
NiklasNeugebauer: aha, might be a good point

rosys/vision/camera/configurable_camera.py
NiklasNeugebauer: yeah, i am kind of on the fence about it, but i think i tend more towards is_connected too. Its obviously not clear why we are not streaming right now, but usually its because of a connection issue.

rosys/vision/mjpeg_camera/mjpeg_device.py
NiklasNeugebauer: hm good idea

rosys/vision/mjpeg_camera/mjpeg_device.py
NiklasNeugebauer: I think we want to say that its not active

NiklasNeugebauer: But i'll have to reconsider the provider logic. Didn't think about that

rosys/vision/camera/camera.py
NiklasNeugebauer: maybe we don't need one. But it should be right if we cannot differentiate the too. if we don't connect after init then we also never activate (automatically).

rosys/vision/mjpeg_camera/mjpeg_device.py
NiklasNeugebauer: yep, i like that. though the same thing. especially since the loop is actually a task.

rosys/vision/usb_camera/usb_device.py
NiklasNeugebauer: yes, i'd say dont bother

NiklasNeugebauer:
I'll look into the changes. thanks

rosys/vision/mjpeg_camera/mjpeg_device.py
NiklasNeugebauer: Done in 8a42d2af. Renamed the outer loop to `_run_capture_task` (it's the actual task, as you say) and the inner single-session method to `_connect_and_stream_images`. I also propagated the `_start_capture_task` / `_run_capture_task` naming to the USB and RTSP devices so all three match, and made the MJPEG starter private (`_start_capture_task`) since nothing calls it from outside.

rosys/vision/mjpeg_camera/mjpeg_device.py
NiklasNeugebauer: @agent-NiklasNeugebauer can you implement that?

rosys/vision/camera/camera.py
```diff
@@ -16,6 +17,39 @@
...
+    unreachable camera stops hammering, and is jittered so cameras that went down together do not
+    retry in lock-step.
+
+    :param reconnect_interval: the configured wait after a stream that was working ended
+    :param failed_attempts: how many attempts in a row have failed to deliver a single frame
+    """
+    interval = max(reconnect_interval, MIN_RECONNECT_INTERVAL)
+    interval *= 2 ** min(max(failed_attempts - 1, 0), 16)
+    interval *= random.uniform(1 - RECONNECT_JITTER, 1 + RECONNECT_JITTER)
+    return min(max(interval, MIN_RECONNECT_INTERVAL), MAX_RECONNECT_INTERVAL)
```
agent-NiklasNeugebauer: **[RV-3cd3fd1c]** suggestion · line-by-line

`retry_delay()` applies jitter *after* clamping to the configured interval, so the returned wait can fall below `reconnect_interval` (a configured 3 s yields 2.4-3.6 s). The `reconnect_interval` docstring promises the device "waits before reopening a stream" — if the configured value is meant as a floor, jitter should only ever extend it (`interval * uniform(1, 1 + RECONNECT_JITTER)`); if it is a target, say so on the property.

Also, the `min(..., 16)` exponent cap is unexplained and unreachable in practice: with `MIN_RECONNECT_INTERVAL = 0.1` the `MAX_RECONNECT_INTERVAL` clamp bites at ~9 consecutive failures, so the cap only guards against float overflow. Either drop it or say why 16.

rosys/vision/usb_camera/usb_device.py
```diff
@@ -82,30 +96,132 @@ def create_capture(index: int) -> cv2.VideoCapture | None:
...
+        Runs until `shutdown()` cancels the task.
+        """
+        try:
+            while self._should_run:
+                streamed = False
+                try:
+                    streamed = await self._run_capture_session()
+                except Exception:
+                    self.log.exception('[%s] capture session failed', self.uid)
+                self._failed_attempts = 0 if streamed else self._failed_attempts + 1
```
agent-NiklasNeugebauer: **[RV-a8a395ec]** suggestion · line-by-line

`self._failed_attempts = 0 if streamed else self._failed_attempts + 1` resets the back-off whenever a session delivered *at least one* frame (`_run_capture_session()` returns `opened`, which is `True` as soon as a capture was opened — not that frames flowed). A camera that opens, yields nothing, and drops therefore keeps `_failed_attempts` at 0 and is retried at the base interval forever, which is exactly the hammering the back-off exists to prevent.

Consider resetting only after a session that actually ran for a while, or counting delivered frames rather than "was a capture open". The same shape exists in `rtsp_device.py:186` and `mjpeg_device.py:170`.

rosys/vision/rtsp_camera/rtsp_device.py
```diff
@@ -85,39 +141,120 @@ async def shutdown(self) -> None:
...
-    def _start_gstreamer_task(self) -> None:
-        self.log.debug('[%s] Starting gstreamer task', self._mac)
+    def _start_capture_task(self) -> None:
+        self.log.debug('[%s] Starting capture loop', self._mac)
         if self._capture_task is not None and not self._capture_task.done():
-            self.log.warning('[%s] capture task already running', self._mac)
+            self.log.warning('[%s] capture loop already running', self._mac)
             return
-        self._capture_task = background_tasks.create(self._run_gstreamer(), name=f'capture {self._mac}')
+        self._should_run = True
```
agent-NiklasNeugebauer: **[RV-38e34258]** suggestion · line-by-line

`_start_capture_task()` sets `_should_run = True` after the "already running" guard, but nothing coordinates it with `shutdown()`. `shutdown()` sets `_should_run = False`, then awaits the task's cancellation for up to 5 s; a `_start_capture_task()` that lands in that window passes the guard (the old task is cancelled/done) and flips `_should_run` back to `True`, resurrecting a device the owner is tearing down. `RtspCamera.set_parameters` guards this with `_device_connection()`, but `restart_gstreamer()` is public and carries no such lock.

`usb_device.py:103` has the same shape.

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -1,19 +1,35 @@
...
 from ..image_processing import remove_exif
 from .vendors import mac_to_url
 
 log = logging.getLogger('rosys.vision.mjpeg_camera.mjpeg_device')
 
 
+class CaptureState(enum.Enum):
+    """State of the self-healing capture loop.
+
+    Modelling the loop as a single state (instead of separate ``should_run``/``streaming``/``authorized``
```
agent-NiklasNeugebauer: **[RV-715eaaf4]** suggestion · comment-check

`Modelling the loop as a single state (instead of separate ``should_run``/``streaming``/``authorized`` flags) keeps only the valid combinations representable.` contrasts the code with flags that do not exist in this file — it documents the change, not the state machine. A reader who never saw the previous shape gains nothing.

The first line plus the per-member comments already carry the information; drop the parenthetical.

rosys/vision/camera/camera.py
```diff
@@ -100,22 +148,33 @@ def from_dict(cls, data: dict[str, Any]) -> Self:
...
         return False
 
+    @property
+    def is_active(self) -> bool:
+        """Whether a connection is wanted, i.e. whether the camera keeps itself connected."""
+        return self.is_connected
+
+    @property
+    def is_reconnecting(self) -> bool:
+        """Whether the camera is active but not currently streaming (waiting to reconnect)."""
```
agent-NiklasNeugebauer: **[RV-15dd3b2c]** nit · comment-check

`"""Whether the camera is active but not currently streaming (waiting to reconnect)."""` restates `return self.is_active and not self.is_connected` in words. The name and the one-line body already say it.

rosys/vision/usb_camera/usb_device.py
```diff
@@ -8,46 +7,75 @@
...
     return image.tobytes()
 
 
 class UsbDevice:
-
-    def __init__(self, video_id: int, capture: cv2.VideoCapture, *,
-                 on_new_image_data: Callable[[np.ndarray | bytes, float], Awaitable | None]) -> None:
-        self._video_id: int = video_id
-        self._capture: cv2.VideoCapture = capture
+    MAX_READ_FAILURES = 10
```
agent-NiklasNeugebauer, line 28: **[RV-9e83d123]** suggestion · test-coverage

`MAX_READ_FAILURES` and the path it guards are untested: no test drives `_run_capture_session()` through 10 consecutive `read()` failures to assert the capture is released and the session ends (the unplugged-cable case this constant exists for). `tests/test_camera_reconnect.py` patches `find_device_node` and a fake capture, so a test that returns `(False, None)` from `read` is cheap to add.

rosys/vision/usb_camera/usb_device.py
```diff
@@ -8,46 +7,75 @@
...
-        for node in device_nodes
-        if node.startswith('/dev/video')
-    ]
-    return min(video_ids) if video_ids else None
+def find_device_node(camera_uid: str) -> str | None:
+    """Find the lowest-numbered ``/dev/video*`` node of the camera with the given uid."""
+    video_nodes = [node.strip() for node in device_nodes_from_uid(camera_uid) if node.startswith('/dev/video')]
+    if not video_nodes:
+        return None
+    return min(video_nodes, key=lambda node: int(node.removeprefix('/dev/video')))
```
agent-NiklasNeugebauer, line 20: **[RV-aa995a09]** suggestion · test-coverage

`find_device_node()` is patched out in every test that touches USB (`test_camera_reconnect.py:364,403,1034,1133,1151,1162`), so its own logic — filtering to `/dev/video*` and picking the *lowest-numbered* node — has no coverage. It is a pure function over `device_nodes_from_uid`; a unit test with a fake uid mapping (multiple nodes, unsorted, non-video entries, empty) would pin the behaviour the reconnect loop depends on.

rosys/vision/usb_camera/usb_camera.py
```diff
@@ -29,12 +29,13 @@ def __init__(self,
...
                          connect_after_init=connect_after_init,
                          **kwargs)
+        self.log = logging.getLogger(f'rosys.vision.usb_camera.{self.id}')
         self._pending_operations = 0
         self.device: UsbDevice | None = None
         self.detect: bool = False
         self.color: str | None = None
 
-        self._register_parameter('auto_exposure', self.get_exposure, self.set_exposure, auto_exposure)
+        self._register_parameter('auto_exposure', self.get_auto_exposure, self.set_auto_exposure, auto_exposure)
```
agent-NiklasNeugebauer: **[RV-7f279363]** suggestion · test-coverage

The five `_register_parameter` calls feed `to_dict()` (`usb_camera.py:44`) and `_apply_all_parameters` on reconnect, but no test round-trips a `UsbCamera` through `to_dict()`/`from_dict()` to assert the parameter values survive. `test_usb_camera_persists_reconnect_interval` covers only `reconnect_interval`; a full round-trip would also pin that `reconnect_interval` is not silently swallowed by the parameter dict merge.

rosys/vision/camera/camera.py
```diff
@@ -16,6 +17,39 @@
...
+
+It also caps the wait after a camera refused the stream, so a camera that starts answering again is
+picked up within this long at the latest.
+"""
+
+RECONNECT_JITTER = 0.2
+"""Relative spread of the wait between two connection attempts."""
+
+
+def retry_delay(reconnect_interval: float, failed_attempts: int = 0) -> float:
```
agent-NiklasNeugebauer: **[RV-e09d04d3]** suggestion · architecture

`retry_delay()`, `clamp_reconnect_interval()` and the three `*_RECONNECT_INTERVAL` constants live in `camera/camera.py`, and all three device modules import them from there (`from ..camera.camera import ...`). The abstract `Camera` itself only uses `clamp_reconnect_interval`; the rest is device-side back-off policy that now makes the camera base module a policy hub the devices depend on.

A small `vision/reconnect.py` (or `camera/reconnect.py`) holding the policy would let devices depend on the policy without depending on the camera base class, and would give the back-off a home of its own to test.

rosys/vision/usb_camera/usb_device.py
```diff
@@ -82,30 +96,132 @@ def create_capture(index: int) -> cv2.VideoCapture | None:
...
             return
-        read_result = await rosys.run.io_bound(self._capture.read)
-        if read_result is None:
-            return
-        capture_success, frame = read_result
+        self._should_run = True
+        self._capture_task = rosys.background_tasks.create(
+            self._run_capture_task(), name=f'usb capture {self.uid}')
+
+    async def _run_capture_task(self) -> None:
```
agent-NiklasNeugebauer: **[RV-1d8f9460]** suggestion · architecture

`_run_capture_task()` is now the third near-identical copy of the same loop (`rtsp_device.py:172`, `mjpeg_device.py:~152`): guard on a run flag, run one session, update `_failed_attempts` from whether it streamed, log a reason, sleep `_retry_interval`, and clear `_capture_task` in `finally`. The differences are the session call, the log wording, and one extra wait rule each.

That is a lot of duplicated liveness logic for three devices that must behave identically. A shared driver (a mixin or a helper taking `run_session()` and `retry_interval()`) would keep the three in step; today a fix to one has to be remembered three times — which the `_failed_attempts` reset issue already demonstrates.

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -1,19 +1,35 @@
...
-from ...rosys import on_startup
+from ..camera.camera import MAX_RECONNECT_INTERVAL, clamp_reconnect_interval, retry_delay
+from ..http import new_async_client
 from ..image_processing import remove_exif
 from .vendors import mac_to_url
 
 log = logging.getLogger('rosys.vision.mjpeg_camera.mjpeg_device')
 
 
+class CaptureState(enum.Enum):
```
agent-NiklasNeugebauer: **[RV-5f4d079b]** suggestion · architecture

`MjpegDevice` models its loop as a `CaptureState` enum, while `RtspDevice` and `UsbDevice` use `_should_run` plus `_authorized`/`_capture` checks for the same three questions (`is_connected`, `is_active`, refused/unauthorized). The enum is the better model — but having two models across sibling devices means each device's `is_connected` has to be read separately to know what it means, and the shared retry loop cannot be factored out while they disagree.

Pick one and apply it to all three, ideally together with the duplication above.

rosys/vision/mjpeg_camera/arkvision_mjpeg_device.py
```diff
@@ -8,18 +8,25 @@
...
             raise ValueError(
                 f'ArkVisionMjpegDevice can only be used with ARKVISION devices. Got {vendor} for mac="{mac}"')
-        self.settings_interface = ArkVisionSettingsInterface(ip)
         super().__init__(mac, ip, index=index, username=username, password=password,
-                         on_new_image_data=on_new_image_data)
+                         on_new_image_data=on_new_image_data, on_connect=on_connect,
+                         reconnect_interval=reconnect_interval)
+
+    @property
+    def settings_interface(self) -> ArkVisionSettingsInterface:
```
agent-NiklasNeugebauer, line 27: **[RV-4b40a4dc]** suggestion · architecture — _disputed: The settings interface is a thin stateless REST wrapper; constructing it is a field assignment, and rebuilding per call is what keeps it correct across address changes without invalidation logic._

`settings_interface` constructs a fresh `ArkVisionSettingsInterface(self.ip)` on every access, and every `get_*`/`set_*` on this class goes through it — so a single UI panel read builds several. Caching it per address (invalidated when `ip` changes, as `RtspDevice._bind_settings_interface()` already does) would keep one object and give the interface somewhere to hold a client or session.

rosys/vision/usb_camera/usb_device.py
agent-NiklasNeugebauer: **[RV-cfdfb191]** suggestion · architecture

All three devices call `_start_capture_task()` from `__init__` (`usb_device.py:55`, `rtsp_device.py:63`, `mjpeg_device.py:82`), so merely constructing a device spawns a background task and starts talking to hardware. That makes the constructor unusable outside a running event loop, gives the owner no window to wire callbacks before frames arrive, and forces every test that wants a device object to also shut it down (`test_devices_do_not_leak_shutdown_handlers` does exactly this dance).

An explicit `start()` that `Camera.connect()` calls would keep construction pure and make the lifecycle symmetric with `shutdown()`.

rosys/vision/camera/camera.py
agent-NiklasNeugebauer: **[RV-185170dc]** suggestion · architecture

`MIN_RECONNECT_INTERVAL`, `MAX_RECONNECT_INTERVAL` and `RECONNECT_JITTER` are module-level constants, so the back-off ceiling is global: a fleet on a flaky link cannot raise the 30 s cap, and a camera that must be retried aggressively cannot lower the floor, without patching the module. `reconnect_interval` is per-camera but only sets the *base*.

If the cap is meant to be tunable, it belongs next to `reconnect_interval` on the camera/device; if not, the docstrings should say it is deliberately fleet-wide.

rosys/vision/simulated_camera/simulated_device.py
```diff
@@ -41,6 +81,14 @@ async def _create_image(self) -> None:
         if isinstance(result, Awaitable):
             await result
 
+    async def _invoke_on_connect(self) -> None:
```
agent-NiklasNeugebauer, line 76: **[RV-9921748d]** suggestion · architecture

`SimulatedDevice` reconnects by comparing `rosys.time() - self._disconnect_time` against `reconnect_interval` inside `on_repeat(self._step)`, while the hardware devices run a task with `retry_delay()` back-off and jitter. So the simulation twin has neither growing back-off nor jitter, and its `is_active` means "repeater running" rather than "capture loop alive".

That is the one divergence that matters here: reconnect behaviour is the feature under development, and it cannot be developed against simulation if simulation implements a different policy. Routing the twin through the same `retry_delay()` would keep the pair honest.

rosys/vision/mjpeg_camera/arkvision_mjpeg_device.py
agent-NiklasNeugebauer, line 27: **[RV-7ddae27a]** suggestion · architecture

`assert self.ip is not None, 'cannot reach the camera settings without an address'` makes a property raise `AssertionError` for an ordinary, expected state — an MJPEG camera legitimately has no address until discovery finds it (`test_mjpeg_camera_is_active_without_known_address`). Callers such as `_prepare_stream()` reach this from inside the capture loop, where the exception is swallowed by the generic `except Exception` and logged as a traceback.

Either return `None` and let callers skip, or raise a domain error the loop handles deliberately. Asserts also vanish under `python -O`.

rosys/vision/simulated_camera/simulated_camera_provider.py
```diff
@@ -35,38 +45,11 @@ def restore_from_dict(self, data: dict[str, dict]) -> None:
...
-        """
-        for camera in self._cameras.values():
-            if not camera.is_connected and self.simulate_device_failure:
-                # return camera with fixed probability
-                if random.random() < 0.5:
-                    yield camera.id
-            else:
-                yield camera.id
-
     async def update_device_list(self) -> None:
```
agent-NiklasNeugebauer, line 35: **[RV-779375ef]** suggestion · architecture

`update_device_list()` is now an empty coroutine, yet `__init__` still registers it via `rosys.on_repeat(self.update_device_list, self.SCAN_INTERVAL)` when `auto_scan` is true (the default) — a repeating no-op, plus a `scan_for_cameras()` that no longer exists on a provider whose base class advertises discovery. If simulated cameras are always added explicitly, skip the registration (or drop `auto_scan`) rather than scheduling nothing.

While here: `backup_to_dict()` (line 37) builds a `cameras` dict and then returns `{}`, so simulated cameras never persist and `restore_from_dict()` can never see them.

rosys/vision/camera/camera.py
```diff
@@ -53,6 +90,16 @@ def __init__(self,
 
         if connect_after_init:
             rosys.on_startup(self.connect)
+        rosys.on_shutdown(self.disconnect)
```
agent-NiklasNeugebauer, line 60: **[RV-1ead5fe1]** suggestion · architecture

`Camera.__init__` registers `rosys.on_startup(self.connect)` and `rosys.on_shutdown(self.disconnect)` while `connect()`/`disconnect()` are empty `# noqa: B027` stubs on the base class. The base therefore commits every subclass to a lifecycle it does not implement, and any subclass that forgets to override gets silent no-ops at startup and shutdown instead of an error.

Making them `@abc.abstractmethod` on this already-`abc.ABC` class would move that to construction time; the `noqa` suppressions are a hint that the empty bodies exist only to satisfy the registration.

rosys/vision/camera/camera.py
NiklasNeugebauer: yep, thats true. the jitter should happen before clamping

rosys/vision/usb_camera/usb_device.py
NiklasNeugebauer: fair, that would indeed happen for rtsp cameras... we want something cheap to measure if a frame ever went through.

rosys/vision/rtsp_camera/rtsp_device.py
NiklasNeugebauer: sounds reasonable

rosys/vision/mjpeg_camera/mjpeg_device.py
NiklasNeugebauer: yes lets minimize this as much as possible. maybe we dont need it at all.

rosys/vision/camera/camera.py
NiklasNeugebauer: yes away

rosys/vision/usb_camera/usb_device.py
NiklasNeugebauer: sure

rosys/vision/usb_camera/usb_device.py
NiklasNeugebauer: sounds good.

rosys/vision/camera/camera.py
NiklasNeugebauer: great

rosys/vision/usb_camera/usb_device.py
NiklasNeugebauer: i like it, was also thinking that. lets come up with that last once others are resolved since we have to experiment with it a little.

rosys/vision/mjpeg_camera/mjpeg_device.py
NiklasNeugebauer: i like the enum

rosys/vision/usb_camera/usb_device.py
NiklasNeugebauer: not sure about this. really we want the device to represent an attempted connection. having to call a start kind of removes that concept.

rosys/vision/camera/camera.py
NiklasNeugebauer: we could consider making those static somewhere but i dont see a good place. i think we can keep it fleet wide and lets not document it.

rosys/vision/simulated_camera/simulated_device.py
NiklasNeugebauer: i agree but not a priority. maybe we can look at it with the consolidation of the loops mentioned earlier (1d8f9460)

rosys/vision/mjpeg_camera/arkvision_mjpeg_device.py
NiklasNeugebauer: raise an error.

rosys/vision/simulated_camera/simulated_camera_provider.py
NiklasNeugebauer: both sound reasonable. lets clean up.

rosys/vision/camera/camera.py
NiklasNeugebauer: fair, but shouldnt we still to the registration? otherwise all subclasses have to implement that.

rosys/vision/camera/camera.py
agent-NiklasNeugebauer: Fixed in 0a5c3e59 — jitter now only ever stretches the wait (`uniform(1.0, 1.0 + RECONNECT_JITTER)`), so 3 s configured means 3.0–3.6 s and never 2.4 s.

The exponent cap went with it: `MAX_DOUBLINGS` is now derived from the two bounds (`ceil(log2(MAX/MIN))` = 9) instead of the guessed 16, so it says what it actually guards. New test `test_a_configured_interval_is_never_undercut`. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/rtsp_camera/rtsp_device.py
agent-NiklasNeugebauer: Fixed in a41eccfe — both `RtspDevice` and `UsbDevice` now set a `_shutting_down` flag for the duration of the tear-down, and `_start_capture_task()` refuses while it is set.

Test `test_a_device_being_torn_down_is_not_restarted` gathers `shutdown()` with a `_start_capture_task()` landing in its await window; it fails on the old code. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/mjpeg_camera/mjpeg_device.py
agent-NiklasNeugebauer: Removed in c38d991f — you were right that it was not needed at all. The class docstring is now just "State of the self-healing capture loop."; the per-member comments carry the rest. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/camera/camera.py
agent-NiklasNeugebauer: Gone in a9396878. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/usb_camera/usb_device.py
agent-NiklasNeugebauer: Fixed in 4f9d25a1 — and the cheap measure was already there, just unused: `_run_capture_session()` now sets `streamed = True` on the line where it hands a frame to `_on_new_image_data`, exactly as `rtsp_device.py:325` and `mjpeg_device.py:271` already did. The only change is *where* the flag is set, so there is no new bookkeeping.

Worth noting the direction was the opposite of my finding: RTSP and MJPEG were already correct, USB was the only one counting an opened capture as success. Two tests: `test_usb_device_backs_off_while_no_frame_arrives` (fails on the old code) and `test_usb_device_releases_a_capture_that_only_fails_to_read` for the `MAX_READ_FAILURES` path. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/usb_camera/usb_device.py
agent-NiklasNeugebauer: Added in 4f9d25a1 — `test_usb_device_releases_a_capture_that_only_fails_to_read`. `FakeCapture` gained a `failing_reads` flag and a `reads` counter, so the test asserts the capture is released at exactly `MAX_READ_FAILURES` reads and that the device then tries to reopen. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/usb_camera/usb_device.py
agent-NiklasNeugebauer: Added in 741e5bf9 — `test_find_device_node_picks_the_lowest_numbered_video_node`, parametrised over unsorted nodes (`video10` must not beat `video1`), a single node, a mixed set with `/dev/media0`, only non-video nodes, and the empty set. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/camera/camera.py
agent-NiklasNeugebauer: Done in a9396878 — new `rosys/vision/reconnect.py` holds the three bounds plus `retry_delay()` and `clamp_reconnect_interval()`; `camera.py` imports only what it uses (`MIN_RECONNECT_INTERVAL`, `clamp_reconnect_interval`), and the three devices no longer import from the camera base module. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/mjpeg_camera/mjpeg_device.py
agent-NiklasNeugebauer: Noted — kept as is for now, since the enum only pays off once it is the model all three devices share, and you want the loop consolidation done last. See my reply on the consolidation comment: I'd bring the enum along with it rather than converting RTSP and USB separately first. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/mjpeg_camera/arkvision_mjpeg_device.py
agent-NiklasNeugebauer: Done in c38d991f — new `CameraAddressUnknown` in `mjpeg_device.py`, raised by the `settings_interface` property of all three vendor devices (ArkVision, Motec, OpenIPC-Zauberzeug — they all had the same assert).

The capture loop catches it next to the existing `httpx.HTTPError` branch and logs the one-line reason it already has for an unreachable camera, so a camera waiting for discovery no longer produces a traceback. Two tests: the property raising it, and the loop reporting it without `exc_info`. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/simulated_camera/simulated_camera_provider.py
agent-NiklasNeugebauer: Both done in 6a52f185.

The no-op scan is gone along with `auto_scan` and `SCAN_INTERVAL` — nothing to discover, so nothing to schedule. Four call sites in the tests dropped the argument.

`backup_to_dict` is deleted rather than fixed: the base `CameraProvider.backup_to_dict` already builds exactly the dict the override was assembling before it returned `{}`, so simulated cameras now actually persist and the `restore_from_dict` below it can see them. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/camera/camera.py
agent-NiklasNeugebauer: Agreed, leaving them fleet-wide and undocumented. They moved to `rosys/vision/reconnect.py` in a9396878, which at least gives them a place that is about the policy rather than about the camera class — if a per-camera override is ever wanted, that is where it goes. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/simulated_camera/simulated_device.py
agent-NiklasNeugebauer: Agreed on both counts — not done here. Left for the loop consolidation (RV-1d8f9460), which is where the twin would pick up `retry_delay()` for free rather than growing a second implementation of it. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/usb_camera/usb_device.py
agent-NiklasNeugebauer: You're right, and I'm withdrawing this one — not done.

"A device *is* an attempted connection" is the invariant the rest of the design leans on: `is_active` is defined as "the capture loop is alive", `connect()` creates the device precisely so that an unreachable camera still keeps trying, and the docs say "`connect()` always creates that device, even when the camera cannot be reached at all". A device that exists but has not been started would be a fourth state — not connected, not reconnecting, not shut down — and every `is_active` reader would have to care.

The two costs I listed are real but small, and both have better answers than a lifecycle change:

- *constructor needs a running loop* — true, but a device without a loop is useless anyway; there is nothing to test on it.
- *no window to wire callbacks* — `on_new_image_data` and `on_connect` are constructor arguments, so there is no window to miss.

The test dance I pointed at is just that a started thing must be stopped, which is correct rather than awkward. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/camera/camera.py
agent-NiklasNeugebauer: Yes — keep the registration. My finding conflated two things, and only the smaller one is worth acting on.

The registration is right where it is: it is what makes `connect_after_init` and `disconnect()`-on-shutdown a property of *every* camera rather than something each subclass remembers. Moving it out would be exactly the "subclasses all have to implement that" you're describing.

What I was actually reaching for is `@abc.abstractmethod`, which is orthogonal: the base would keep registering, and a subclass with a device would be forced to say what connecting means instead of silently getting a no-op. But it does not survive contact with `ReplayCamera` (`record_replay/replay_camera.py:17`), a `Camera` with no device at all — it would have to grow two empty overrides to satisfy the ABC, which is worse than the empty base it inherits today.

So: left as is, and I put the reason into the `connect()` docstring in a9396878 so the next reader does not re-derive it —

```
A camera without a device of its own, such as one replaying recorded images, needs no
implementation, so this is not abstract.
```

The `# noqa: B027` suppressions stay; that is flake8-bugbear complaining about exactly this pattern, and here it is the intended one. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/usb_camera/usb_device.py
agent-NiklasNeugebauer: Agreed — left for last, nothing done here.

One thing the fixes in this round sharpened, as input for when you get to it. The USB back-off bug (4f9d25a1) was a one-line difference in the shared shape: RTSP and MJPEG set `streamed` when a frame is handed over, USB set it when a capture opened. Three copies, one of them silently wrong for months, and no test could have caught it generically because there was nothing shared to test.

The tear-down race (a41eccfe) went the other way: the same fix had to be applied twice, to `rtsp_device.py` and `usb_device.py` separately, and MJPEG needs neither because its `CaptureState` already encodes `STOPPED`. That is the argument for doing the enum (RV-5f4d079b) *as part of* the consolidation rather than before it — the state model is what decides how much of the loop can be shared, and the MJPEG one is the one that made a guard unnecessary rather than duplicated.

Happy to prototype it in a separate PR against this branch once the rest has settled, so the experimenting does not sit in this diff. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/mjpeg_camera/arkvision_mjpeg_device.py
agent-NiklasNeugebauer: Withdrawing this one myself — the dispute was right.

`ArkVisionSettingsInterface`'s whole state is `ip` plus a logger (`arkvision_rest_client.py:34`), so constructing it is two assignments. And the benefit I claimed, "somewhere to hold a client or session", is refuted by the class itself: every `_get`/`_post` opens its own `async with new_async_client()`, so a cached interface would still build one client per request. Caching would only add ip-invalidation logic for nothing. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/usb_camera/usb_device.py
agent-NiklasNeugebauer: One correction to my own finding, now that it is fixed: I wrote that "the same shape exists in `rtsp_device.py:186` and `mjpeg_device.py:170`". That is wrong — both of those already set `streamed` on a delivered frame. USB was the only one counting an opened capture. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Fixes%20implemented%20by%20the%20agent%20from%20the%20reviewer%27s%20replies%3B%20the%20reviewer%20decided%20each%20item.)

rosys/vision/usb_camera/usb_device.py
agent-NiklasNeugebauer: Opened as #466 — targets this branch, so it reads as a diff on top of the review fixes.

It came out smaller than feared: **-81 lines net**, and every device file shrank. `CaptureDevice` owns the loop, back-off, task handle and lifecycle; a device supplies `_run_session()` and, optionally, `_tear_down_session()` / `_retry_reason()`.

Two things fell out for free, which is why I folded them in rather than leaving them:

- **RV-5f4d079b** — the enum was the natural state model for the shared loop, so RTSP and USB adopt it rather than being converted separately.
- **RV-9921748d** — `SimulatedDevice` joins the same loop, so the twin now has the growing, jittered back-off it lacked. `test_every_device_backs_off_the_same_way` holds all four to it, which is the test that could not exist before.

One deliberate API change: `shutdown()` is async everywhere. The two synchronous ones cancelled the task and returned without waiting, which is why the tests needed a `shutdown_device()` helper for the two shapes — that helper is gone.

Details and the one open design question in the PR description. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Refactor%20designed%20and%20implemented%20by%20the%20agent%3B%20the%20reviewer%20set%20the%20direction%20and%20the%20scope.)

NiklasNeugebauer:
Superseded by a stack of three PRs, so the core can be reviewed on its own:

1. #470 — vision I/O helpers (shared SSL context, request timeouts, udev scan lock)
2. #471 — RTSP and MJPEG reconnection (the `CaptureDevice` loop, `is_connected`/`is_active` contract, providers only discover and rebind)
3. #472 — USB and simulated reconnection, docs

How the still-open threads here map onto the new code:

- **`is_active` default in `Camera`** (@ftilde): still `return self.is_connected`, now documented as "whether a connection is wanted". For a camera without a device of its own that collapses to "connected". Open for discussion on #471 if you think the base should not have a default at all.
- **USB: why only some settings are reapplied on reconnect**: the device now calls the camera's `_apply_all_parameters` via `on_connect` after every reopen, so auto-exposure and exposure are reapplied too (`test_usb_camera_reapplies_parameters_after_reconnect`, #472).
- **MJPEG auth failure and the provider recreating the device**: a refused stream no longer ends the loop. The device stays `is_active`, backs off to one attempt every 30 s and recovers when the answer changes; providers no longer connect or recreate devices, they only rebind a changed IP (#471).
- **Leaking cameras through `on_shutdown`**: `rosys.on_shutdown` on `main` holds bound methods weakly, and `Camera` registers its `disconnect` once instead of once per device (`test_devices_do_not_leak_shutdown_handlers`, #472).
- The three agent findings (USB back-off reset, `CaptureState` docstring, `reconnect.py` module) are in the new branches.

---
⬛ claude-fable-5-1

PR #425, opened 2026-06-29
Mjpeg device process

ftilde:
### Motivation

MJPEG streams load the main process twice: the HTTP stream is read and split on the event loop, and each frame goes through `cpu_bound` for decoding and comes back as a 6 MB pickled array.

### Implementation

- Add `MjpegStreamWorker`: a spawned process per stream session that reads, splits and decodes the stream.
- Add `stream_channel.py`: message types plus a `MessageSender`/`MessageReceiver` pair.
  On Linux frames cross as memfd file descriptors over a socket pair and the parent `mmap`s them zero-copy; elsewhere they are pickled over a plain pipe.
  `open_channel()` probes `memfd_create` when a channel is opened (via libc when `os.memfd_create` is missing, as in uv's Python builds).
- Latest-frame-wins in the reader thread: an unconsumed frame is replaced by the next one before it reaches the event loop, so a stalled loop retains at most one frame per camera.
- `MjpegDevice` delivers decoded `ImageArray`s; `MjpegCamera` applies rotation and crop in-process like `RtspCamera`.
- A stream that opens and then stalls is logged at WARNING; a dead worker reports its exit code.
- `decode_jpeg_image` always returns three channels (the PIL fallback returned 2-D for grayscale).
- Axis default fps is 10, matching `MjpegCamera`, so Axis cameras no longer reopen the stream after the first frame.
- `SPAWN_CONTEXT` lives in `rosys.helpers.spawning`, shared with path planning; `ImageDataHandler` lives in `capture_device.py`.
- README and AGENTS.md note that scripts creating cameras at import time need an `if __name__ == '__main__':` guard.
- Tests: channel round trips for both transports, in-process `_run_worker` tests with a stub sender, end-to-end camera test through the worker, reconnect tests against a real worker.

Trade-off: each session start spawns a process (≈1 s, mostly `import rosys`), which also delays reconnects by that much.

Idle CPU-load on the main process with one front and two main cameras drops from ~80% to ~20% (without memfd-Transport ~35%). Load for spawned processes is ~50% (main cameras) and <15% (front/back cameras).

### Progress

---
⬛ claude-fable-5-1


NiklasNeugebauer:

**Verdict: request changes.** The architecture (own the MJPEG stream + JPEG decode in a spawned subprocess, ship decoded frames to the parent, forward `set_*`/`get_*` over a control pipe) is sound, and I verified it end-to-end against a local fake OpenIPC MJPEG server: connect works, frames arrive in the parent, control calls round-trip, `disconnect()` is clean and fast (~0.4 s). But there is one behavioral regression and several robustness gaps.

**Blocking**

1. **Zombie-connected state after the stream ends (reproduced).** `MjpegCamera.is_connected` is now `self._capture.is_running`, i.e. `process.is_alive()`. When the MJPEG stream ends or errors, the child's capture task terminates ("Capture task stopped") but `_main()` keeps waiting on the `_stop` event — the process stays alive forever. Repro (fake server closes the stream after 1 s): `is_connected == True`, child alive, no new frames, indefinitely. On `main`, `is_connected` flips to `False` when the capture task dies, which is exactly what `CameraProvider` rescans key off to reconnect. With this PR, a single dropped stream permanently kills the camera while it *reports connected*. Fix options: set `_stop` when the device's capture task finishes (exit the child), or have `is_running` reflect capture-task liveness via the control channel. (If #423 lands inside the child, the loop self-heals, but `is_connected` semantics still need to be defined for the "retrying" state.)

**Non-blocking but important**

2. **Blocking `Connection.send` inside the child's event loop.** `_handle_image_data` runs synchronously in the capture task and `send()`s a decoded RGB frame (~6 MB at 1080p) into a 64 KB OS pipe. If the parent's reader thread stalls, the send blocks the child's *entire* event loop — including control-command handling — and frames back up instead of being dropped (no latest-frame-wins). Consider a latest-only handoff or sending from an executor.
3. **`MjpegCapture.call()` has no timeout.** `_call_sync` does a blocking `recv()` while holding `_control_lock`. A hung child permanently occupies one default-executor thread *and* the lock, so every subsequent `call()` piles up another blocked executor thread. Add a timeout (`Connection.poll(timeout)`) and fail the call.
4. **Reconnect leaks the previous parent-side pipe ends.** `connect()` only checks `is_connected`; after a child death it overwrites `self._capture` without `shutdown()` on the old handle — `_image_reader`/`_control_parent` are only closed by GC finalizers. Cheap fix: `if self._capture is not None: await self._capture.shutdown()` first.
5. **Child-side failures are invisible.** If `MjpegDeviceFactory.create` or initial parameter application raises in `_main`, the parent sees only `is_connected == False`. The child's logging is unconfigured (spawn = fresh interpreter), so diagnostics rely on Python's last-resort stderr handler. Consider forwarding a startup status/error over the control pipe and configuring logging in `run()`.
6. **`spawn` re-imports `__main__`.** Any user script constructing an `MjpegCamera` at module level without an `if __name__ == '__main__'` guard now breaks in surprising ways (I hit this myself in the repro — the child re-executed the script). Same caveat as the path planner, but cameras are far more commonly constructed in user scripts. Deserves a docs note; also note the ~1.5–2 s spawn+import latency per camera (observed 1.75 s to first frame).
7. **Data volume tradeoff should be stated.** Decoded RGB over the pipe is *more* bytes than the JPEG it replaces (6 MB vs ~100–500 KB per frame). The win is moving decode off the main process; the pipe copy is the new cost — which is exactly what #428 removes. I'd land the two together (benchmarks in the #428 review).
8. **Public behavior change:** `MjpegDevice` now starts its capture task immediately when constructed inside a running loop (previously always deferred to `on_startup`). Fine for the subprocess, but it changes semantics for any downstream code constructing devices directly — changelog-worthy.
9. **No tests** for `MjpegCapture`/`MjpegCaptureProcess` (the existing 18 mjpeg/camera tests pass on this branch, but none touch the new path). The lifecycle is very testable: a fake HTTP MJPEG server + a real spawned capture runs in ~3 s and would have caught (1).
10. Heads-up: overlaps heavily with #423 (both rework the mjpeg capture loop) and #400 (auth resolution would move into the child). Worth deciding the landing order explicitly.

ftilde:
Note that the implementation was completely redesigned after changes on main.

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -102,73 +88,60 @@ async def _prepare_stream(self) -> None:
...
+            if isinstance(callback_result, Awaitable):
+                await callback_result
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            self.log.error('Error processing image: %s', e)
+
+    async def _tear_down_session(self) -> None:
+        if self._worker is None:
+            return
+        await self._worker.shutdown()
+        self._worker = None
```
felixwrz, line 144: **MAJOR** — `_tear_down_session` reads `self._worker`, awaits its shutdown and then clears `self._worker`. When `restart_capture()` (the `ip` setter) cancels a session, the new session assigns its own worker during that await, and the old session's `self._worker = None` then wipes the *new* worker: the new session fails on its next `receive()` and worker B's process, reader thread and camera connection live on unreaped, even past `device.shutdown()`. Please bind the worker to a local in `_run_session` and shut that one down in its `finally`, or guard with `if self._worker is worker` the way `RtspDevice` checks `_capture_process` identity (rtsp_device.py:106, :221).

<details><summary>Sequence and repro</summary>

1. Device streams from camera A; the provider sets `device.ip = B`. `_stop_capture_task` cancels the old task, `_start_capture_task` starts a new one.
2. Old task's `finally` → `_tear_down_session` reads worker A and suspends in `worker.shutdown()` (`run_in_executor(join)`).
3. New task runs `_run_session` and assigns `self._worker = MjpegStreamWorker(B)`.
4. Old task resumes and executes `self._worker = None`.
5. New session's `await self._worker.receive()` raises `AttributeError` → logged as "capture session failed" with a traceback, retried every `reconnect_interval`; its own `_tear_down_session` sees `None` and returns, so worker B is never terminated.

Reproduced deterministically with a fake worker whose `shutdown()` yields once. A regression test: start a stream, set `device.ip` while connected, then assert the device reaches STREAMING again and only one worker process is alive after `shutdown()`.
</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -102,73 +88,60 @@ async def _prepare_stream(self) -> None:
...
-            async with open_stream(client, url, self._username, self._password) as response:
-                if response is None:
-                    self._set_state(CaptureState.REFUSED)
+        await self._prepare_stream()
+        self._worker = MjpegStreamWorker(self._mac, url, self._username, self._password)
+        try:
+            while True:
+                message = await self._worker.receive()
+                if isinstance(message, StreamOpened):
+                    await self._enter_streaming()
```
felixwrz, line 104: **MAJOR** — `StreamOpened` calls `_enter_streaming()` without checking `_keeps_running()`, unlike the other branches. `shutdown()` sets STOPPED, then awaits `_tear_down_session()` before cancelling the task, so a `StreamOpened` that is already queued flips the state back to STREAMING, fires `on_connect` (`_apply_all_parameters`) on a camera being torn down, and turns the worker's intentional death into `StreamEnded(FAILED)` → `RuntimeError` → an ERROR traceback plus a scheduled retry; the device ends in CONNECTING instead of STOPPED. The root cause is that the parent cannot tell "I terminated the worker" from "the worker died". I'd have `MjpegStreamWorker.shutdown()` record the intentional close and map an EOF after it to `EndReason.ENDED` (as `RtspDevice` does via the SIGTERM return code), which lets every `StreamEnded` be handled uniformly and removes the `_keeps_running()` special case two lines below. The minimal fix is a `_keeps_running()` guard before `_enter_streaming()`.

<details><summary>Sequence and a side effect</summary>

1. `camera.disconnect()` → `device.shutdown()` sets `_state = STOPPED`, awaits `_tear_down_session()` (yields in the executor join).
2. The capture task pops the queued `StreamOpened`; `_set_state(STREAMING)` only checks task identity, which passes, so STOPPED is overwritten.
3. `on_connect` runs `MjpegCamera._apply_all_parameters` against a disconnecting camera.
4. The reader thread posts the EOF sentinel `StreamEnded(FAILED, 'the stream worker exited')`; `if self._keeps_running()` is now true (state is STREAMING), `_end_session` raises `RuntimeError`, `_describe_session_error` logs an ERROR traceback and the loop schedules a retry until `_await_capture_task` cancels it.

Reproduced by a verifier. Side effect of the same gap: after every *normal* stream end the reader thread appends a second, phantom `StreamEnded(FAILED)` behind the real `StreamEnded(ENDED)`; it is only never read because the session returns first.
</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/mjpeg_stream_worker.py
```diff
@@ -0,0 +1,176 @@
...
+            self.log.warning('stream worker did not end; killing it')
+            self._process.kill()
+            await self._loop.run_in_executor(None, self._process.join, 1.0)
+        self._receiver.close()
+
+    def _read_messages(self) -> None:
+        while True:
+            try:
+                message = self._receiver.receive()
+            except (EOFError, OSError):
```
felixwrz, line 62: **MAJOR** — `_read_messages` only survives `EOFError`/`OSError`. Any other exception from `receive()` propagates out of the daemon thread (only `threading.excepthook` sees it), no sentinel is posted, `_message_arrived` is never set again and `MjpegStreamWorker.receive()` awaits forever, leaving the device in STREAMING with `is_connected == True`, no frames and no reconnect until `shutdown()`. Please catch `Exception` here, log it, and still post the `None` sentinel so the device reconnects.

<details><summary>A realistic trigger with the memfd channel</summary>

On a long-running robot whose parent process hits its fd limit, the kernel drops the SCM_RIGHTS descriptor (`MSG_CTRUNC`). `multiprocessing.reduction.recvfds` then gets `msg=b'\x01', ancdata=[]`, skips its `if not msg and not ancdata: raise EOFError` branch and raises `RuntimeError('received 0 items of ancdata')` — not an `OSError`. This is the "connected but frozen" state from the earlier review, reached through a different door. A receive timeout in `_run_session` (e.g. twice the read timeout) would be a cheap second floor against any kind of stall.
</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/mjpeg_stream_worker.py
```diff
@@ -0,0 +1,176 @@
...
+                return
+            send(StreamOpened())
+            for jpeg, capture_time in _split_frames(response.iter_bytes()):
+                array = decode_jpeg_image(remove_exif(jpeg))
+                if array is not None:
+                    send(Frame(array=array, capture_time=capture_time))
+        send(StreamEnded(reason=EndReason.ENDED))
+    except (BrokenPipeError, OSError):
+        return  # the parent is gone
+    except httpx.HTTPError as e:
```
felixwrz, line 173: **MAJOR** — `main` had `except httpx.ReadTimeout: log.warning('Connection to %s timed out')` for a stream that opens and then stalls. The PR drops it, and `ReadTimeout` (a subclass of `HTTPError`) now lands here as `EndReason.UNREACHABLE`, so the device logs `cannot reach the camera: timed out` at INFO only. A camera that accepted the connection and then hung its firmware is invisible to anyone reading at WARNING, and the text says the opposite of what happened. Please catch `httpx.ReadTimeout` before this clause and report it as its own reason (e.g. `STALLED`) that the device logs at WARNING, or at least map it to `ENDED` with a WARNING as before.

<details><summary>Repro</summary>

Verified with a server that sends the 200 multipart header and then nothing: the sync client with `DEFAULT_TIMEOUT = 5.0` raises `ReadTimeout` from `iter_bytes()` after 5.02 s (MRO: `ReadTimeout → TimeoutException → TransportError → RequestError → HTTPError`). On `main` this logged one WARNING plus the INFO retry line; on this branch only `[mac] cannot reach the camera: timed out; retrying in 3.0 s`. No test covers the old warning, which is why its loss went unnoticed.
</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/mjpeg_stream_worker.py
```diff
@@ -0,0 +1,176 @@
...
+    """Handle of the process that reads and decodes one MJPEG stream session."""
+
+    def __init__(self, name: str, url: str, username: str | None, password: str | None) -> None:
+        self._loop = asyncio.get_running_loop()
+        self.log = logging.getLogger(f'rosys.vision.mjpeg_camera.mjpeg_stream_worker.{name}')
+        self._messages: deque[Message] = deque()
+        self._message_arrived = asyncio.Event()
+
+        self._receiver, sender = open_channel()
+        self._process = SPAWN_CONTEXT.Process(target=_run_worker, args=(url, username, password, sender),
```
felixwrz, line 36: **Question** — Was a long-lived worker per device considered, one that receives `open(url, auth)` / `close` commands like `PlannerProcess`, so a reconnect is a socket reopen instead of a spawn? The description names the ~1 s per session, but because this module lives inside `rosys.vision.mjpeg_camera`, each child re-imports the whole `rosys` package (~0.65 s measured on an M-series Mac, several seconds on a Jetson), and that price is paid on every network flap, URL change and read timeout. If the per-session spawn stays, a lean entry module that imports only httpx, numpy and the decoder would remove most of the cost.

<details><summary>Measurements and two consequences the description does not mention</summary>

- `import rosys.vision.mjpeg_camera.mjpeg_stream_worker` costs 630–740 ms here (`rosys.analysis` 379 ms, `nicegui` 208 ms, `scipy` 188 ms) and logs `Restore failed: rosys.config.json not found` from persistence; a bare spawn is ~20 ms, so ~97 % of the session start is the import.
- **Axis cameras spawn twice at startup:** `AxisSettings(fps=6, …)` (axis_mjpeg_device.py:39) differs from `MjpegCamera(fps=10)` (mjpeg_camera.py:22), so `_apply_all_parameters` changes the URL after the first frame → "stream settings changed; reopening" → 3 s sleep → second spawn. Aligning the defaults avoids that regardless of the design.
- An import-time failure in the child (a missing system library, say) surfaces only as `the stream worker exited`, retried every 3 s with no diagnostic.
- `spawn` re-imports the user's `__main__`; scripts that construct an `MjpegCamera` at module level without an `if __name__ == '__main__'` guard now break, and cameras are constructed in user scripts far more often than the path planner is. Worth a docs note either way.
- The tests acknowledge the cost: `forward_until` attempts went from 20 to 200 "to cover the second or more every MJPEG session needs to spawn its stream worker".
</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/mjpeg_stream_worker.py
```diff
@@ -0,0 +1,176 @@
...
+                detail = f'{response.status_code} {response.reason_phrase} (auth: {auth_scheme})'
+                send(StreamEnded(reason=EndReason.REFUSED, detail=detail))
+                return
+            send(StreamOpened())
+            for jpeg, capture_time in _split_frames(response.iter_bytes()):
+                array = decode_jpeg_image(remove_exif(jpeg))
+                if array is not None:
+                    send(Frame(array=array, capture_time=capture_time))
+        send(StreamEnded(reason=EndReason.ENDED))
+    except (BrokenPipeError, OSError):
```
felixwrz, line 171: **CLEANUP** — `except (BrokenPipeError, OSError): return` wraps the whole worker body, and `BrokenPipeError` is already an `OSError`. Any real `OSError` — `_memfd_create` failing with `EMFILE`, `os.write` failing under memory pressure for a large frame, `_ssl_context()` failing to load the CA bundle in the fresh child — exits silently, and the parent only ever sees the fabricated `the stream worker exited`, retried forever. httpx wraps socket and SSL errors in `httpx.HTTPError`, not `OSError`, so this guard can be narrowed to the `send` calls (`BrokenPipeError`, `ConnectionResetError`) and every other `OSError` left to the `except Exception` clause that reports `type(e).__name__: e`.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/stream_channel.py
```diff
@@ -0,0 +1,173 @@
...
+    def close(self) -> None:
+        self._connection.close()
+
+
+def open_memfd_channel() -> tuple[MessageReceiver, MessageSender]:
+    reader, writer = SPAWN_CONTEXT.Pipe(duplex=True)  # a socket pair, which can carry file descriptors
+    return MemfdReceiver(reader), MemfdSender(writer)
+
+
+open_channel = open_memfd_channel if memfd_is_available() else open_pickled_channel
```
felixwrz, line 172: **CLEANUP** — This line runs the memfd probe as an import-time side effect: `import rosys` → `rosys.vision` → `mjpeg_camera` → `mjpeg_device` → `stream_channel` (verified in `sys.modules`), so every rosys-importing process — the app, each `cpu_bound` pool worker, each spawned stream worker, pytest — creates and closes a memfd at import, and the transport is frozen in a module global that tests cannot inject. Consider a `functools.cache`d `open_channel()` called from `MjpegStreamWorker.__init__` instead. Related, in `_memfd_create` (line 106): the ctypes fallback does `ctypes.CDLL(None)` plus the `memfd_create` dlsym lookup on every call, and `MemfdSender.send` calls it once per frame; on uv's python-build-standalone interpreters, where `os.memfd_create` is genuinely absent (it is here, 3.13.11), that is the production hot path at stream fps. Resolving `libc.memfd_create` once at module level with `argtypes`/`restype` fixes that.

<details><summary>Also</summary>

`test_the_platform_channel_is_chosen` re-evaluates the same expression as this line and compares it with itself, so it cannot fail; a lazy `open_channel()` would let the test patch `memfd_is_available` and assert the choice instead.
</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

tests/test_camera_reconnect.py
```diff
@@ -693,30 +695,38 @@ async def stream(self) -> None:
...
+    await server.start()
+    device = AxisMjpegDevice(AXIS_MAC, f'127.0.0.1:{server.port}', on_new_image_data=lambda data, timestamp: None)
+    device.reconnect_interval = 0.2
+    try:
+        await forward_until(lambda: device.is_connected, message='expected the stream to be opened')
+        assert server.connections == 1
+        await device.set_fps(12)
+        await forward_until(lambda: server.connections == 2,
+                            message='expected the stream to reopen after the settings changed')
+        assert 'fps=12' in (device.url or '')
```
felixwrz, line 708: **CLEANUP** — This assertion holds the instant `set_fps(12)` returns: `AxisMjpegDevice.set_fps` only writes `axis_settings.fps`, and `url` is recomputed from the settings on every read, so it says nothing about what the worker requested. `FlakyMjpegServer._handle` reads the request line (line 201) and discards it. A regression where the worker is constructed from a stale URL, or the session keeps the previous worker, still yields a second connection and passes. The test on `main` asserted `'fps=6' in opened_urls[0] and 'fps=12' in opened_urls[1]`; could the server record the request paths so the test can assert `fps=12` on the second one?

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/stream_channel.py
```diff
@@ -0,0 +1,173 @@
...
+from multiprocessing import reduction
+from multiprocessing.connection import Connection
+from typing import Protocol
+
+import numpy as np
+
+from ..image import ImageArray
+
+# spawn, not fork (which is broken for Python), regardless of the global start method (see path planning, #19)
+SPAWN_CONTEXT = multiprocessing.get_context('spawn')
```
felixwrz, line 18: **CLEANUP** — Same definition (and the same comment) as `SPAWN_CONTEXT` in `rosys/pathplanning/planner_process.py:16`; one shared home for it would keep the two from drifting.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -1,40 +1,25 @@
...
 from ..capture_device import CaptureDevice, CaptureState
-from ..http import new_async_client
-from ..image_processing import remove_exif
+from ..image import ImageArray
+from .mjpeg_stream_worker import MjpegStreamWorker
+from .stream_channel import EndReason, Frame, StreamEnded, StreamOpened
 from .vendors import mac_to_url
 
-log = logging.getLogger('rosys.vision.mjpeg_camera.mjpeg_device')
+ImageDataHandler = Callable[[ImageArray, float], Awaitable | None]
```
felixwrz, line 13: **CLEANUP** — `rtsp_device.py:30` spells out the identical `Callable[[ImageArray, float], Awaitable | None]` inline. Since both devices now deliver decoded arrays, this alias belongs in `capture_device.py` next to the base class, used by both.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/mjpeg_device.py
```diff
@@ -102,73 +88,60 @@ async def _prepare_stream(self) -> None:
...
+                self.log.debug('capture session ended')
+            case EndReason.REFUSED:
+                self.log.error('camera at %s refused the stream: %s', url, message.detail)
+                self._set_state(CaptureState.REFUSED)
+            case EndReason.UNREACHABLE:
+                raise CameraUnreachable(message.detail)
+            case EndReason.FAILED:
+                raise RuntimeError(message.detail)
+
+    async def _deliver(self, frame: Frame) -> None:
```
felixwrz, line 131: **CLEANUP** — The call-then-await-if-awaitable dance here is what `rosys.helpers.invoke(handler, *args)` already does; `_deliver` could reduce to the timestamp fallback, `await invoke(self._on_new_image_data, frame.array, timestamp)` and the error log.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/stream_channel.py
```diff
@@ -0,0 +1,173 @@
...
+class MessageReceiver(Protocol):
+
+    def receive(self) -> Message:
+        ...
+
+    def close(self) -> None:
+        ...
+
+
+class PickledSender:
```
felixwrz, line 66: **CLEANUP** — `PickledSender`/`PickledReceiver` are pure pass-throughs to `Connection.send`/`recv`/`close`. If the `MessageReceiver` protocol named its method `recv` instead of `receive`, the raw `Connection` objects would satisfy both protocols and the two classes could go.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/mjpeg_camera.py
```diff
@@ -93,17 +91,9 @@ async def _tear_down_device(self) -> None:
...
-            if image_array is not None:
-                image = Image.from_array(image_array, camera_id=self.id, time=timestamp)
-        else:
-            image = await rosys.run.cpu_bound(Image.from_jpeg_bytes, image_bytes, camera_id=self.id, time=timestamp)
-
-        if image is not None:
-            self._add_image(image)
+    async def _handle_new_image_data(self, image_array: ImageArray, timestamp: float) -> None:
+        transformed_image_array = process_ndarray_image(image_array, self.rotation, self.crop)
+        self._add_image(Image.from_array(transformed_image_array, camera_id=self.id, time=timestamp))
```
felixwrz, line 96: **Note** — The removed `Image.from_jpeg_bytes` path expanded 2-D (grayscale) decodes to three channels (image.py:106); this line hands the worker's raw `decode_jpeg_image` result to `Image.from_array`, whose `assert len(array.shape) == 3` (image.py:121) fails on every frame for a grayscale MJPEG camera on a host without libturbojpeg (the PIL fallback returns `(h, w)`), logging "Error processing image" at stream fps while `is_connected` stays True. Rare in our fleet, but a silent regression vs `main`; normalizing to three channels right after `decode_jpeg_image` in `_run_worker` (which also makes the memfd header shape uniform) or a sentence in the description would cover it.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

evnchn:
**`latest-frame-wins` bounds the deque, not the frames waiting on it.**

[`_read_messages`](https://github.com/zauberzeug/rosys/blob/c216b03e568601a3f1c490b457295fd08a1bf348/rosys/vision/mjpeg_camera/mjpeg_stream_worker.py#L58-L69) runs in its own thread and `MemfdReceiver.receive()` maps each frame *before* handing it to `call_soon_threadsafe`. The collapse in `_handle_incoming_message` then runs on the event loop — so with a responsive loop you do get latest-frame-wins, and with a blocked one every pending callback pins a live 6.22 MB mmap.

- event loop blocked 4 s, 1080p: reader pulled **60 frames, Shmem +356 MiB**, then collapsed 60 → 1 once it drained
- the transports differ in how much they let queue up: receiver never reading, **pickled completes 0 of 60 sends** (a 6.22 MB pickle does not fit the pipe buffer), **memfd completes 60 of 60**. [`MemfdSender.send`](https://github.com/zauberzeug/rosys/blob/c216b03e568601a3f1c490b457295fd08a1bf348/rosys/vision/mjpeg_camera/stream_channel.py#L130-L140) puts ~100 bytes and an fd on the wire, so the socket buffer no longer limits retained pixel memory — the fd does, and each holds a whole decoded frame in tmpfs

Not a leak: the references release as callbacks drain. But there is no application-level bound on frames already mapped and waiting, once per camera, on exactly the loop this PR exists to unblock. Two separable guarantees if you want them — coalescing before `call_soon_threadsafe` bounds what the parent retains; stopping the child running ahead of a stalled loop needs something flowing the other way. Spans both files, so no suggestion block.

I'd want that bounded before memfd becomes the Linux default, but the mechanism is yours to pick.

Measured on Linux in Docker against verbatim copies of the sender, receiver and `_read_messages`, since memfd is Linux-only and I'm on macOS. I did not drive the real worker through `FlakyMjpegServer`, so these are transport-level numbers, not end-to-end ones.

Context for why I went looking: #428 was closed as landing here. The stalled-parent resource-budget concern in its point 4 still applies at `c216b03`, though the retention mechanism has moved — child-side dup'd fds then, parent-side mapped frames now. Of the rest — 1 is structurally gone (`send_handle`/SCM_RIGHTS needs no live sender, and `receive()` maps inside the reader's `try`, so a failure is caught by its `except (EOFError, OSError)`; once the sends complete, a child exiting no longer prevents delivery); 2 is moot; 3 is fixed for the `uint8` contract by carrying `shape`; 5 has channel tests, though same-process rather than the spawned round trip asked for; 6 was praise, not a finding; 7 is addressed by `CaptureState.STREAMING`.

felixwrz (changes requested):
**Summary** — Moving the MJPEG read/split/decode into a spawned worker and shipping frames over memfd is the right direction, and the CPU numbers in the description back it up. I read the diff against `CaptureDevice` and `RtspDevice`, ran the gates locally and verified the findings below against the sources (several were reproduced). The redesign resolves most of the earlier review; the remaining points are inline at their lines.

**MAJOR** (inline)
- `_tear_down_session` clears the shared `_worker` after an await, so an `ip` change orphans the new session's worker process.
- The `StreamOpened` branch skips `_keeps_running()`, so a shutdown racing the first frame flips STOPPED back to STREAMING; the worker should report its intentional end.
- The reader thread only survives `EOFError`/`OSError`; any other exception leaves the device waiting in STREAMING forever.
- The `ReadTimeout` WARNING from `main` is gone; a stalled stream now reads like a connect failure at INFO.

**Questions** (inline)
- Was a long-lived worker per device (open/close commands, like `PlannerProcess`) considered? Each session currently pays a full `import rosys` on spawn, and Axis cameras spawn twice at startup.

**CLEANUP** (inline)
- Narrow the `OSError` net in `_run_worker` to the send calls.
- Resolve libc once and make the channel choice lazy instead of an import-time side effect of `import rosys`.
- The URL-change test no longer asserts which URL the second connection used.
- A few definitions duplicate existing ones (`SPAWN_CONTEXT`, `ImageDataHandler`, `_deliver` vs `rosys.helpers.invoke`, the pickled pass-throughs).

**Note** (inline)
- Grayscale JPEGs via the PIL fallback are no longer expanded to three channels; rare, but a silent regression worth a sentence in the description.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20read%20of%20the%20diff%20and%20its%20verification%20agents%2C%20triage%20and%20severities%20by%20the%20reviewer%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's read of the diff and its verification agents, triage and severities by the reviewer, wording by the model.")

rosys/vision/mjpeg_camera/mjpeg_device.py
ftilde: fixed

rosys/vision/mjpeg_camera/mjpeg_device.py
ftilde: fixed

rosys/vision/mjpeg_camera/mjpeg_stream_worker.py
ftilde: fixed

rosys/vision/mjpeg_camera/mjpeg_stream_worker.py
ftilde: fixed

rosys/vision/mjpeg_camera/mjpeg_stream_worker.py
ftilde: Workarounds in 43199b07 9d0668db and 9be5b6f3. We should probably improve the import times of rosys...

rosys/vision/mjpeg_camera/mjpeg_stream_worker.py
ftilde: fixed in fd4d6aa1

rosys/vision/mjpeg_camera/stream_channel.py
ftilde: fixed in fa1bfcaf4e6ea92735441ddbfc5f61eaeec84126

tests/test_camera_reconnect.py
ftilde: fixed in 7163fd95

rosys/vision/mjpeg_camera/stream_channel.py
ftilde: fixed in 4f1b35fc

rosys/vision/mjpeg_camera/mjpeg_device.py
ftilde: fixed in 3d2204d1

rosys/vision/mjpeg_camera/mjpeg_device.py
ftilde: fixed in da383cb6

rosys/vision/mjpeg_camera/stream_channel.py
ftilde: fixed in 8db88463

rosys/vision/mjpeg_camera/mjpeg_camera.py
ftilde: fixed in 13638ef2, i.e. fixes decode_jpeg_image directly. Its type annotation was actually incorrect before :scream:

ftilde:
Problems that @evnchn identified fixed in e44c9419, 6b958856, 70d8fdef, 984532ff

ftilde:
@NiklasNeugebauer Note: If you think the spawn delay is a problem, we could switch to forkserver on linux. But before we do that, we should probably try to reduce the rosys import time.

NiklasNeugebauer (approved):

NiklasNeugebauer:
> @NiklasNeugebauer Note: If you think the spawn delay is a problem, we could switch to forkserver on linux. But before we do that, we should probably try to reduce the rosys import time.

I don't think the spawn delay is really a problem since we don't really reconnect all that often.

PR #427, opened 2026-06-30
Derive the repeater lifetime from the handler type

ftilde:
### Motivation

So far `rosys.on_repeat` implicitly kept the object alive forever via the reference stored in the task, so an object that registered a repeated method and was later discarded leaked one timer per instance (one of the drivers behind #436)., For some objects it is unclear whether or not they are constructed in UI contexts.
An example of this problem is `CameraProjector` which is (sometimes) constructed in a ui context, uses a repeater and is thus leaked. We also cannot use `ui.timer` due to the "sometimes".

This PR derives the lifetime of the repetition from the handler, so a repeated bound method no longer keeps its object alive — the repetition ends when the object is discarded:

```python
class Camera:
    def __init__(self) -> None:
        rosys.on_repeat(self.step, 0.1)  # ends when this camera is dropped

    def step(self) -> None:
        ...


cameras: dict[str, Camera] = {}
cameras['front'] = Camera()  # stored, so it repeats
del cameras['front']         # dropped, so the repetition ends without leaking
```

For page-scoped periodic UI updates, prefer NiceGUI's `ui.timer`, which is tied to the client and stops when the page is closed.

### Implementation

After the discussion in #429, the `weak` parameter from earlier iterations of this PR was dropped in favor of deriving the repeater's lifetime from the handler type:

- **Bound methods** are held weakly (via `weakref.WeakMethod`): the repetition ends when the object is garbage-collected.
- **Plain functions** (including transparently decorated ones) are held strongly: they repeat until shutdown.
- **Lambdas, capturing closures, partials and other callables** are rejected with a `TypeError` that explains how to restructure the handler, because no single lifetime rule fits them.

A check after startup warns when a repeated method's object was never stored (pointing to `ui.timer` for page-scoped UI); if the object is collected later, the repetition ends silently with a debug log.
The `Repeater` bookkeeping was also cleaned up: the task registry prunes itself via done-callbacks, and a stopped repeater can be started again.
The README documents the lifetime rule and the `ui.timer` boundary.

**Breaking changes:**

- `on_repeat(obj.method, ...)` no longer keeps `obj` alive — store the object (or the returned `Repeater`) to define the lifetime of the repetition.
- Lambdas, capturing closures and partials are no longer accepted as handlers.

Note that the repetition only ends when the object is actually collected: since RoSys disables automatic garbage collection, objects in reference cycles are only collected under memory pressure and keep repeating until then.

### Progress

- [x] I chose a meaningful title.
- [x] Documentation has been added.


falkoschindler (commented):
Thanks for the PR, @ftilde! Let me share a few findings after a first AI-assisted review pass:

1. **Weak teardown is unreliable in production because automatic GC is disabled — rosys.py (333), camera_projector.py (32)**

A weak repeater only stops once its target object is actually freed and `WeakMethod` clears. But rosys.py (333) does `gc.disable()`, and the only `gc.collect()` — in `_garbage_collection`, rosys.py (280-283) — runs solely under memory pressure (free < 300 MB). So objects are reclaimed _only by refcounting_. Any target object that sits in a reference cycle — which is exactly what the UI-context objects this feature is aimed at tend to do (NiceGUI element ↔ parent links, `bind_*`/event closures capturing `self`) — will not be freed by refcounting, so its repeater keeps ticking indefinitely, which is the very leak the PR set out to fix. The new tests pass only because they call `gc.collect()` explicitly (`test_weak_repeater_stops_when_object_is_collected`, `test_collected_object_does_not_start_orphan_task_after_startup`), which masks this gap — production code never calls it on the timescales that matter. Worth either documenting the limitation on the `weak=` parameter (works by refcounting; cyclic owners won't be collected promptly under `gc.disable()`) or confirming the intended targets like `CameraProjector` are cycle-free. `CameraProjector` itself looks acyclic, so it is fine; the risk is the generalized feature being weaker than the docstring/tests imply.

2. **Stop log line logs the wrapper's default repr instead of the handler name — rosys.py (211)**

`log.info('%s must be stopped', self.handler)` formats `self.handler`, which for a weak repeater is a `_WeakHandler` with no custom `__repr__`. The error branches two lines down, rosys.py (219) and (223), use `self.handler.__qualname__`. On stopping a weak repeater the log shows `<rosys.rosys._WeakHandler object at 0x…>` instead of e.g. `CameraProjector.step`, degrading diagnosability. Trivial fix: use `self.handler.__qualname__` here too (it is preserved on `_WeakHandler`).

3. **`SimulatedDevice` leaks its repeater on disconnect — a real instance of the bug this PR targets — simulated_device.py (27), simulated_camera.py (51)**

`SimulatedDevice.__init__` does `self._repeater = rosys.on_repeat(self._create_image, …)` with the default strong reference, but `SimulatedCamera.disconnect()` just sets `self.device = None` and never stops the repeater. So the orphaned device stays alive and keeps ticking, emitting images into `_add_image` forever — one leaked repeater per disconnect/reconnect cycle. Compare `UsbDevice` in usb_device.py (50), which gets this right: its `disconnect()` calls `self._capture_task.stop()`. Preferred fix is the explicit `.stop()` (mirrors `UsbDevice`, doesn't lean on refcounting/GC timing under `gc.disable()` per finding 1); `weak=True` would also stop the leak but is the implicit version of a fix the codebase already does explicitly. Audited all 28 `on_repeat` call sites: the rest are long-lived singletons (hardware modules, camera providers, system services) where `weak=True` would be wrong, and `UsbDevice` already handles teardown explicitly — this is the only genuine analog.

I haven't looked into the tests yet and will do that later once the rest is stable.

ftilde:
Thanks for the review @falkoschindler. 2 is fixed. For 3 the issue is far more widespread (but now fixed, see recent commits). UsbDevice in fact never worked as intended since `stop()` was called in `__del__()`, which was never called due to the repeater.
I'm not sure what we can do about 1. The PR at least doesn't make the issue worse. At some point we _have_ to collect garbage. Maybe we can discuss in person.

falkoschindler (changes requested):
Thanks for tackling this — the `_WeakHandler` mechanism itself is sound (async handlers still work through `invoke`, the done-callback/stop/restart interplay is correctly guarded, and the new tests pin the tricky lifetime semantics nicely). But the blanket `weak=True` conversion and the `__qualname__` changes introduce a few regressions, two of them silent, so I'd like to see these addressed before merging.

**Requesting changes for:**

1. **`rosys/persistence/backup_schedule.py:27` — silent loss of backups.** The repeater was the *only* strong reference to `BackupSchedule`; it isn't `Persistable` and has no subscriptions. A fire-and-forget `BackupSchedule()` statement — exactly the usage our docs suggest ("you can use the `BackupSchedule` module") — is now refcount-collected right after `__init__`, so daily backups never run, with no warning or log. Verified on the branch: the handler is dead immediately after bare instantiation, even with gc disabled. Please keep this one strong (or give it a self-anchor).

2. **`rosys/analysis/legacy/network_monitor.py:32` — same mechanism for `NetworkMonitor`.** It's a pure side-effect module (nothing reads `interfaces`), so a bare `NetworkMonitor()` dies instantly and network monitoring silently stops existing. Verified empirically. Keep strong or anchor.

3. **`rosys/rosys.py:182` — the graceful fallback crashes.** For `weak=True` with a `functools.partial` or callable object, `_weaken` correctly returns `None`, but the warning then dereferences `handler.__qualname__`, which those don't have (on all of 3.11–3.13) → `AttributeError` from `Repeater.__init__` instead of warn-and-degrade. `getattr(handler, '__qualname__', repr(handler))` fixes it.

4. **`rosys/rosys.py:211` — shutdown log regression for strong handlers.** `log.info('%s must be stopped', self.handler)` → `self.handler.__qualname__` breaks partials/callable objects at shutdown: line 211 raises, the `except` block re-raises on line 219's identical access, and the task ends with an error dump instead of the clean `break`. (Lines 219/223 predate this PR, but the same `getattr` treatment would fix all three.)

5. **`rosys/rosys.py:169` — unguarded `weakref.WeakMethod`.** Bound methods of `__slots__` classes without `__weakref__` (e.g. `@dataclass(slots=True)`, our house style for value types) raise `TypeError` here, so `weak=True` crashes instead of taking the warn-and-fallback path. A `try/except TypeError` reusing the fallback makes the behavior uniform. (No converted call site triggers this — latent public-API edge.)

**Non-blocking, but worth a look:**

6. **`rosys/vision/simulated_camera/simulated_device.py:27` — the leak fix doesn't cover `remove_camera`.** `SimulatedCamera` ↔ `SimulatedDevice` form a reference cycle (`device._on_new_image` is the camera's bound `_add_image`), and RoSys runs with `gc.disable()`, so dropping a camera without `disconnect()` (which is all `CameraProvider.remove_camera` does — `del self._cameras[id]`) leaves the device and its repeater running until an emergency low-memory `gc.collect()`. Verified with a runtime demo. The explicit-`disconnect()` path *is* fixed now, which is progress — but `remove_camera` should probably disconnect, or the cycle be broken.

7. **`weak=True` is a provable no-op at five sites** where a sibling registration pins the object for the process lifetime: `app_controls_.py:62` (`on_shutdown(self.clear)` + three strong `Event.subscribe` calls), `path_planner.py:53` (`on_startup`/`on_shutdown`), and the usb/rtsp/mjpeg providers (`on_shutdown(self.shutdown)`; `shutdown_handlers` is never pruned, `startup_handlers` isn't cleared after `startup()`). Suggest dropping the flag there so it doesn't suggest collectability that can't happen — or weakening those registrations in a follow-up.

8. **`rosys/core.py:14`** — `Callable[..., Repeater]` erases all static checking on the alias (missing interval, swapped args, and `week=True` all pass mypy 1.19.1 silently — tested). A small callback `Protocol` (`def __call__(self, handler: Callable, interval: float, *, weak: bool = False) -> Repeater: ...`) keeps full checking and is assignment-compatible with the real function.

Happy to re-review once 1–5 are addressed; the core of the PR is good and I'd like to see it land.

---

*Disclaimer: this review was researched and written by Claude (Fable 5); findings 1–6 were verified by executing the code on this branch.*

NiklasNeugebauer:

**Verdict: approve with two fixes.** The core mechanism is correct and well-tested (nice test file). I verified the motivating leak independently: 50 page-widget objects registering `on_repeat(self.update, ...)` with `weak=False` stay alive forever; with `weak=True` all 50 are collected and their repeater tasks self-terminate and are discarded from `Repeater.tasks`. Full suite green on the branch (165 passed).

1. **Bug: `weak=True` crashes on `functools.partial` and callable instances.** The warning path does `handler.__qualname__`, which neither has:
   `Repeater(functools.partial(obj.method), 1.0, weak=True)` → `AttributeError: 'functools.partial' object has no attribute '__qualname__'` (verified). Same for a callable object. Use `getattr(handler, '__qualname__', repr(handler))`. The same hard `.__qualname__` access was added to the "must be stopped" log line in `_repeat` — same fallback needed there, or a `partial` handler (even with `weak=False`) trips it at shutdown.
   Optional follow-up: a `partial` wrapping a bound method could actually be weakened by unwrapping `.func`/`.args`.
2. **The blanket migration of ~20 internal call sites to `weak=True` is a semantic change for downstream apps** and deserves loud release notes: any object an app constructs *without retaining a reference* (a provider, `Odometer`, hardware wrapper, the persistence backup owner, RobotBrain heartbeat owner…) will now be garbage-collected and its background loop silently stops. Previously `on_repeat` itself kept these alive. Two asks:
   - When a weak repeater ends because its object was collected, log it (e.g. `log.info('%s stopped: object was garbage-collected', qualname)`). Right now the loop just exits — a silently stopped heartbeat or backup loop is painful to debug.
   - Double-check the safety-relevant sites (RobotBrain `send_heartbeat`, persistence `backup`, `_ensure_connection`): are their owners guaranteed to be strongly referenced in every supported app topology? If there is any doubt, keep those strong.
3. Agree with keeping `weak=False` as the default for the reasons in the description — explicitly-held repeaters shouldn't stop just because the object happens to be collectable.
4. Small: `test_repeater.py` pokes `_state`/`startup_handlers` internals — acceptable, but a `reset_before_test`-style helper would keep the tests honest against refactors.

evnchn:
**#427 or #429?** Can't have 2 approvals...

falkoschindler:
I think we should aim at #429. Though running multiple Fable 5 sessions last night, arguing against each other, finding more and more issues - I'm rather confused right now. Hard to prove correctness with this codebase...

ftilde:
I haven't looked into the details of Refcounting not always working, but: What is even the point the pattern of "callable modules"? Why not make them (documented side-effect-full) functions?

Specifically:
> A fire-and-forget BackupSchedule() statement — exactly the usage our docs suggest ("you can use the BackupSchedule module")

I don't buy this. Why would you expect that the schedule persists when you don't store the object? The sentence in the readme gives no indication that you can just drop the resulting object on the floor.

Same for `NetworkMonitor` imho. Not storing the object is (or should be) a bug.

ftilde:
I've pushed fixes for the `__qualname__` crashes (apparently also an issue on `_run_handler`, i.e. for example `on_startup`, as well as for the non-weakref-able items. Those cases should be rare, and callers should become aware of the problem via the warning.

falkoschindler:
Thanks for the `__qualname__` fixes, @ftilde — those, the task pruning and above all your leak diagnoses (`UsbDevice`/`SimulatedDevice` pinned by their own repeater so `__del__` never fires; providers leaking on replacement) are valuable regardless of the API question. Those leaks are bugs on `main` today, so I'd suggest fixing them in their own PR, decoupled from the mechanism debate — but as explicit teardown instead of `weak=True`: as committed here they rely on `__del__` firing, which the device ↔ camera cycle (`_handle_new_image_data`/`_on_new_image` back-references) prevents on every path that doesn't already call `disconnect()` — `remove_camera` being the obvious one. Would you author that PR? The audit is yours and you know the sites best. For the `weak=` mechanism itself I've laid out the assumptions and reasoning in one place over in #429 — including why the provider-swap scenario isn't fixed by weak references either (providers are cyclic via their cameras' `NEW_IMAGE` subscriptions). Let's continue the mechanism discussion there.

_Disclaimer: written by Claude (Fable 5)._

ftilde:
I wonder if we should also add a `weak` parameter to `on_shutdown` since it has the same problem of (accidentally?) leaking objects via a bound method and a global registry. I think we should (if we decide it's the right call for `on_repeat`).

`on_startup` doesn't really have this problem in practice, but we could think about doing it for consistency.

ftilde:
@falkoschindler i've implemented the plan we have devised. One more case was surfaced and handled: decorated functions (used by a docs test).

jsb-zz:
I wanted to test this feature with our Feldfreund and had to change code to got it working. This is a breaking change for downstream `on_repeat` callers. Here is a summary on that: 

## Problem Summary
_Disclaimer: written by Claude (Opus)._

The new `_prepare_handler` makes `on_repeat`/`Repeater` **raise `TypeError` at registration time** for two previously-accepted handler kinds:

- **lambdas** (`rosys.py:180`)
- **any plain function with a `__closure__`** — i.e. a nested function that captures variables, including `nonlocal` state or an enclosing `self` (`rosys.py:183-188`)

The design intent reads well and the error messages are clear and actionable. But a few things are worth calling out for downstream projects:

1. **No deprecation window.** Passing a local closure or lambda to `on_repeat` was idiomatic before this PR. Those call sites now fail hard rather than warning. Because the failure happens when the handler is *registered* (often during system/component construction), it can take down startup entirely rather than degrading one feature.

2. **Fails late, at runtime only.** Type checkers and tests won't flag it unless the specific registration path executes, so downstream breakage tends to surface first on real deployments.

3. **Migration is non-trivial.** The prescribed fix — hoist captured locals into instance attributes and convert the handler into a bound method — is more than a rename. A self-contained local closure with `nonlocal` state has to be restructured into scattered instance state.

4. **Inconsistent with `.subscribe`.** Event subscriptions still accept closures and lambdas; only `on_repeat`/`Repeater` reject them. That asymmetry is surprising from a caller's perspective.

### Requests
- Please add a **CHANGELOG / migration-guide entry** documenting the new restriction and the before/after refactor pattern.
- Consider whether capturing closures should be a **warning with strong-handler semantics** (their pre-existing, well-defined behavior: live until shutdown) rather than a hard error — reserving the raise for the genuinely ambiguous cases.
- If the hard error stays, an explicit opt-in (e.g. a `weak=False` / "run until shutdown" flag) would let callers keep concise closures without restructuring into instance attributes.

falkoschindler:
I should have posted this earlier, but this is the plan Dominik and I came up with after multiple meetings and discussions with Claude Code. This is what motivated yesterday's code changes. For future (human or AI) readers, the context and explanation might help to get the full picture:

---

# Repeater lifetime without a `weak` parameter

This picks up where #427 left off. The `WeakMethod` mechanic from that PR is the right one — the open question was under which circumstances to apply it, i.e. whether we need a user-facing `weak` parameter. We think we don't, and the reasoning fits in three steps.

**Why the decision can't be automated for arbitrary callables:** whether a repeater should hold its callback's object strongly (keep it alive) or weakly (die with it) is a question of intent that the call site doesn't reveal. Worse, for callables that _capture_ their object (lambdas, closures, partials), "dies with the object" is provably impossible under reference counting: holding the callback strongly leaks the object or forms an uncollectable cycle, holding it weakly kills it instantly. So instead of asking the user to pick via a parameter, we restrict what the API accepts — then the behavior follows from the callback type alone.

**What real code actually passes:** we checked all `on_repeat`/`Repeater` call sites in rosys, feldfreund, and three private customer projects (~70 sites). Every single one passes a bound method or a module-level function, with one exception (a capture-free nested function that wants to run forever). Restricting the API therefore costs one trivial refactoring across the whole ecosystem.

**The mindset:** a `Repeater` object is a _remote control_, not the anchor of the behavior. Holding it lets you `start()` and `stop()`; throwing it away changes nothing. The repetition's lifetime is attached to the object behind the bound method: behavior belongs to the object it operates on and ends with it — or, for plain functions, to the app itself and ends at shutdown. This mental model replaces both "the repeater keeps things alive" and "dropping the repeater stops it".

**The resulting behavior:**

| You pass                         | User wants                                         | Behavior                                      |
| -------------------------------- | -------------------------------------------------- | --------------------------------------------- |
| `self.step`                      | dies with object                                   | 🔗 weak — stops when the object dies          |
| `MyThing().step`, stored nowhere | ⚠️ forgot the assignment                           | 🪦 never ticks, logs a warning (see below)    |
| module-level `def`               | forever                                            | 💪 strong — runs until shutdown               |
| static-/classmethod              | forever                                            | 💪 effectively forever — no instance involved |
| `lambda: self.step(5)`           | bind an argument + "dies with `self`" — impossible | ❌ `TypeError`                                |
| capturing closure                | forever + local state                              | ❌ `TypeError`                                |
| `partial(self.step, 5)`          | bind arguments, probably method-like               | ❌ `TypeError`                                |
| callable object                  | own it vs. die with it — undecidable               | ❌ `TypeError`                                |

**How the distinction is made:** `inspect.ismethod` identifies bound methods (including `obj.__call__` and classmethods) → `weakref.WeakMethod`; `inspect.isfunction` plus `__name__ == '<lambda>'` and `__closure__` separate lambdas and capturing closures from plain functions; everything else raises a `TypeError` whose message names the supported spelling. Receivers that don't support weak references (`__slots__` without `__weakref__`) fail loudly at construction, too.

**Telling "forgot the assignment" apart from a legitimately short-lived object:** same trick as NiceGUI's `AwaitableResponse` — schedule a check task at construction. If the receiver is already gone once control returns to the event loop, no reference can ever have been stored, so we warn; an object that dies later was stored somewhere, so its death is the expected semantics and only logs at DEBUG.

**Rewriting the rejected cases:** every rejected spelling is an attempt to express per-object behavior without a method, so the fix is always the same — write a small bound method (captured state and bound arguments become attributes or default parameters) or, for run-forever behavior, a named module-level function. The `TypeError` messages say exactly that.

Net effect on #427: the `weak` parameter disappears, the `WeakMethod` mechanic stays, and every call site gets exactly one behavior that is either obvious from the callback type or explained by an immediate error at first startup.

falkoschindler:
Thanks for putting this through its paces on Feldfreund, @jsb-zz! Point by point:

1. **Hard failure at startup:** intentional — it fires deterministically on the first start after a deliberate upgrade, in development, with the fix in the error message. That's the mildest form a breaking change can take; accepting with a warning would keep exactly the silent-leak semantics from #436 alive.
2. **Type checkers / tests:** type checkers can't help even in principle — lambda, closure and plain function are the same type. But since the error fires at *registration*, any test or smoke run that merely constructs the component catches it; that's as early as a runtime check can fire.
3. **Migration:** only 1 of the ~70 call sites we surveyed needed the closure-to-class rewrite (see plan above) — how many did you hit in Feldfreund? A CHANGELOG/migration entry with a before/after example is committed for the release; with that, AI-assisted migration should make quick work of the rest.
4. **`.subscribe`:** agreed — but it's sequencing, not design: events have the same leak mechanism (one of the drivers behind #436) and are their own work item. `on_repeat` goes first because it had the concrete leak reports.

On the two alternatives: a warning with strong-handler semantics would preserve exactly the default we're removing, and a `weak=False` opt-in would reintroduce the per-call-site parameter whose removal is the point of the plan above. The "run forever with state" spelling already exists and costs one line: a named function with module-level state, or a small stored object with a method.

_Disclaimer: drafted by Claude (Fable 5), reviewed by Falko._

ftilde:
Feldfreund problems should be fixed in https://github.com/zauberzeug/feldfreund/pull/555.

falkoschindler (approved):
It's been quite a ride. In the end this PR (hopefully) adds more guardrails than confusion, to avoid leaks and silently non-looping repeaters. If you agree with my recent changes, feel free to merge, @ftilde! 🏁

PR #428, opened 2026-07-01
Make mjpeg transfer faster

ftilde:
### Motivation

Mjpeg cameras handling is currently still bottlenecked by reading the image from the pipe.

### Implementation

Introduce MemfdImage which stores pixels in a memory-backed anonymous file. This handle can cheaply be sent to another process and be unpacked there without copies.
Disadvantage: This is only supported on Linux and only on python versions with a recent libc.

Depends on #425.

### Progress

### DIFF
--- a/rosys/vision/__init__.py
+++ b/rosys/vision/__init__.py
@@ -10,7 +10,7 @@ from .detections import BoxDetection, Detection, Detections, PointDetection
 from .detector import Autoupload, Detector
 from .detector_hardware import DetectorHardware
 from .detector_simulation import DetectorSimulation, SimulatedObject
-from .image import BytesImage, Image, ImageArray, ImageSize, MemfdImage
+from .image import Image, ImageArray, ImageSize, MemfdImage
 from .image_rotation import ImageRotation
 from .mjpeg_camera import MjpegCamera, MjpegCameraProvider
 from .multi_camera_provider import MultiCameraProvider
@@ -25,7 +25,6 @@ __all__ = [
     'Autoupload',
     'BoxAnnotation',
     'BoxDetection',
-    'BytesImage',
     'CalibratableCamera',
     'CalibratableCameraProvider',
     'Calibration',
--- a/rosys/vision/__init__.py
+++ b/rosys/vision/__init__.py
@@ -10,7 +10,7 @@ from .detections import BoxDetection, Detection, Detections, PointDetection
 from .detector import Autoupload, Detector
 from .detector_hardware import DetectorHardware
 from .detector_simulation import DetectorSimulation, SimulatedObject
-from .image import Image, ImageArray, ImageSize
+from .image import BytesImage, Image, ImageArray, ImageSize
 from .image_rotation import ImageRotation
 from .mjpeg_camera import MjpegCamera, MjpegCameraProvider
 from .multi_camera_provider import MultiCameraProvider
@@ -25,6 +25,7 @@ __all__ = [
     'Autoupload',
     'BoxAnnotation',
     'BoxDetection',
+    'BytesImage',
     'CalibratableCamera',
     'CalibratableCameraProvider',
     'Calibration',
--- a/rosys/vision/__init__.py
+++ b/rosys/vision/__init__.py
@@ -10,7 +10,7 @@ from .detections import BoxDetection, Detection, Detections, PointDetection
 from .detector import Autoupload, Detector
 from .detector_hardware import DetectorHardware
 from .detector_simulation import DetectorSimulation, SimulatedObject
-from .image import BytesImage, Image, ImageArray, ImageSize
+from .image import BytesImage, Image, ImageArray, ImageSize, MemfdImage
 from .image_rotation import ImageRotation
 from .mjpeg_camera import MjpegCamera, MjpegCameraProvider
 from .multi_camera_provider import MultiCameraProvider
@@ -46,6 +46,7 @@ __all__ = [
     'ImageRotation',
     'ImageSize',
     'Intrinsics',
+    'MemfdImage',
     'MjpegCamera',
     'MjpegCameraProvider',
     'MultiCameraProvider',
--- a/rosys/vision/image.py
+++ b/rosys/vision/image.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import mmap
+import os
 from dataclasses import dataclass, field
 from typing import Any, ClassVar, Self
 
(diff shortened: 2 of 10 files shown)


NiklasNeugebauer:
sounds good.

This is just my recent obsession leaking, but it reminds me of the process-shared vision buffer openpilot uses -> https://github.com/commaai/msgq/blob/9beb84af67527f7b6bfee349dcbe3dbb8d4f4789/msgq/visionipc/visionbuf.h

evnchn:
*Posted by Claude Code on @evnchn's behalf — I ran the numbers below; happy to share the scripts or re-run anywhere useful.*

**TL;DR — the perf claim holds, with one honest asterisk.**

`MemfdImage` beats pickling the decoded frame through the `Pipe` at **every** resolution, and the lead grows with frame size. But "zero-copy" is mostly a *deferral*, not an elimination:

- `to_image()` hands back an **mmap-backed** array, so the pixel copy is paid on **first access** (page-in).
- Measuring only `recv()` shows up to **1.7×**; once the consumer actually reads every pixel (JPEG re-encode for the stream, the detector) the real win is **1.16–1.30×**.
- At the default 640×480/6 fps the saving is ~0.4 ms/frame — negligible. The payoff lands exactly where the PR motivates it: **high res + high fps** (1080p: ~63 → ~81 fps of headroom).
- **No fd leak** and round-trip is **byte-for-byte** correct — the two things most likely to bite a `memfd` path.

Round-trip latency, median ms (lower = better), 300 timed frames after warmup:

| Resolution | consumer | pickle | memfd | speedup |
|---|---|--:|--:|--:|
| 640×480 (0.92 MB) | recv only | 1.79 | 1.38 | 1.30× |
| 640×480 | **+read pixels** | 2.72 | 2.33 | **1.16×** |
| 1280×720 (2.76 MB) | recv only | 4.71 | 3.07 | 1.53× |
| 1280×720 | **+read pixels** | 7.04 | 5.77 | **1.22×** |
| 1920×1080 (6.22 MB) | recv only | 10.69 | 6.25 | 1.71× |
| 1920×1080 | **+read pixels** | 15.97 | 12.32 | **1.30×** |

<details>
<summary><b>Method &amp; environment (what this proves — and what it doesn't)</b></summary>

- **Env:** Linux x86, kernel 6.17, glibc 2.39, Python 3.12.3, numpy 2.4.3, 4 cores.
- `MemfdImage.from_array` / `to_image` copied **verbatim** from this PR's diff. Compared against the PR's own fallback path (`Image.from_array` + `Connection.send` pickle) over a **real `spawn` `Pipe`** between two processes.
- Ping-pong latency: parent requests a frame, child packages + sends, parent reconstructs (+ optionally consumes), timed per round trip. Fixed source array reused each frame so the only cost measured is package → IPC → reconstruct → consume.
- **Fairness fix I had to make to my own harness first:** my initial pickle path did an extra `src.copy()` before `Image.from_array`, a memcpy the real code never pays (`from_array` just wraps the fresh decode output; `send()` pickles it once). Removing it sped pickle up and shrank memfd's lead to the numbers above — so these are the *conservative* figures.
- **Scope — this isolates only the transfer mechanism.** It does **not** capture the PR's other (likely bigger) benefit: moving decode off the main event loop into the subprocess, which unblocks a busy event loop regardless of transfer cost.
- **Hardware caveat:** x86 with fast RAM. RoSys targets are often ARM/Jetson with lower memory bandwidth, where pickle's copy cost is relatively higher — so memfd's advantage there is plausibly **larger**, not smaller. I couldn't test ARM; glad to if a box is handy.
</details>

<details>
<summary><b>Throughput view (implied max FPS, +read-pixels case)</b></summary>

| Resolution | pickle FPS | memfd FPS |
|---|--:|--:|
| 640×480 | 368 | 429 |
| 1280×720 | 142 | 173 |
| 1920×1080 | 63 | 81 |

pickle is bandwidth-bound (~580 MB/s serialize, roughly flat MB/s across sizes); memfd amortizes fixed fd-passing overhead, reaching ~1000 MB/s recv-only at 1080p. When the consumer reads pixels, both converge toward memory-read bandwidth (~400–500 MB/s), which is why the realistic-case lead is smaller than the recv-only lead.
</details>

<details>
<summary><b>FD-leak probe (the classic <code>memfd</code> footgun) — clean</b></summary>

Minted + consumed **3,000** frames at 1280×720, sampling `/proc/<pid>/fd` in both processes:

```
 frame  parent_fds  child_fds
     0           9          9
   500           9          9
  1000           9          9
  1500           9          9
  2000           9          9
  2500           9          9
  2999           9          9
```

Stable — the mmap-backed array is GC'd and its fd closed in `to_image()`; the sender's original fd is closed in the `finally`. No accumulation in the happy path. (Not tested: the drop-without-consume path, i.e. a `MemfdImage` pickled/sent but never `to_image()`'d — worth a thought if frames can be dropped under backpressure, since the `DupFd` resource-sharer holds the dup until the receiver detaches.)
</details>

<details>
<summary><b>Correctness</b></summary>

Round-trip `from_array` → pickle across process boundary → `to_image()`: `np.array_equal` **True**, and `metadata` / `camera_id` / `time` preserved. ✅
</details>

<details>
<summary><b>Self-contained repro</b></summary>

```python
import mmap, os, statistics, time
from dataclasses import dataclass, field
from typing import Any
import numpy as np

@dataclass(slots=True)
class ImageSize:
    width: int; height: int

@dataclass(slots=True, kw_only=True)
class Image:
    camera_id: str; size: ImageSize; time: float; array: np.ndarray
    metadata: dict = field(default_factory=dict)
    @classmethod
    def from_array(cls, a, *, camera_id, time, metadata=None):
        h, w = a.shape[:2]
        return cls(camera_id=camera_id, size=ImageSize(w, h), time=time, array=a, metadata=metadata or {})

# --- MemfdImage: verbatim from the PR ---
@dataclass(slots=True, kw_only=True)
class MemfdImage:
    fd: Any; nbytes: int; size: ImageSize; camera_id: str; time: float
    metadata: dict = field(default_factory=dict)
    @classmethod
    def from_array(cls, array, *, camera_id, time, metadata=None):
        from multiprocessing.reduction import DupFd
        height, width = array.shape[:2]; nbytes = int(array.nbytes)
        fd = os.memfd_create(f'rosys-image-{camera_id}')
        try:
            os.ftruncate(fd, nbytes)
            with mmap.mmap(fd, nbytes) as buf:
                np.frombuffer(buf, dtype=np.uint8).reshape(height, width, 3)[:] = array
            dup_fd = DupFd(fd)
        finally:
            os.close(fd)
        return cls(fd=dup_fd, nbytes=nbytes, size=ImageSize(width, height),
                   camera_id=camera_id, time=time, metadata=metadata or {})
    def to_image(self):
        fd = self.fd.detach()
        try:
            buf = mmap.mmap(fd, self.nbytes)
        finally:
            os.close(fd)
        arr = np.frombuffer(buf, dtype=np.uint8).reshape(self.size.height, self.size.width, 3)
        return Image.from_array(arr, camera_id=self.camera_id, time=self.time, metadata=self.metadata)

def producer(conn, method, h, w):
    src = np.random.default_rng(0).integers(0, 256, (h, w, 3), dtype=np.uint8)
    while conn.recv() is not None:
        conn.send(MemfdImage.from_array(src, camera_id='c', time=time.time()) if method == 'memfd'
                  else Image.from_array(src, camera_id='c', time=time.time()))
    conn.close()

def run(ctx, method, h, w, n=300, consume=True, warm=30):
    parent, child = ctx.Pipe()
    p = ctx.Process(target=producer, args=(child, method, h, w), daemon=True)
    p.start(); child.close(); lat = []
    for i in range(n + warm):
        t0 = time.perf_counter(); parent.send('go'); pl = parent.recv()
        img = pl.to_image() if isinstance(pl, MemfdImage) else pl
        if consume: int(img.array.sum())
        if i >= warm: lat.append((time.perf_counter() - t0) * 1e3)
    parent.send(None); p.join(3)
    return statistics.median(lat)

if __name__ == '__main__':
    import multiprocessing; ctx = multiprocessing.get_context('spawn')
    for h, w in [(480, 640), (720, 1280), (1080, 1920)]:
        for m in ('pickle', 'memfd'):
            print(f'{m:7} {w}x{h} consume med_ms={run(ctx, m, h, w):.3f}')
```
</details>

**Ask:** nothing blocking — sharing this as empirical validation for review. Net read: the mechanism is correct and the direction is right; the biggest real-world lever is probably the event-loop offload rather than the raw transfer speedup, and the transfer win is most meaningful at high res/fps. If an ARM/Jetson target is handy I'll re-run there, since that's where I'd expect the largest gain.

NiklasNeugebauer:

**Verdict: approach validated, one real bug, needs tests.** I benchmarked the claim with real spawn processes shipping 1080p RGB frames across a pipe: **pickle-over-pipe 64.2 fps (≈400 MB/s) vs memfd 162.7 fps (≈1 GB/s)** — a 2.5× throughput win, and more importantly the parent-side CPU spent deserializing drops to ~zero. Recommend putting numbers like these in the PR description.

1. **Bug (reproduced): `DupFd.detach()` requires the *sending* process to still be alive.** `DupFd` serves the fd lazily through the sender's `resource_sharer` Unix socket. If the child exits while frames are still queued in the pipe (normal `shutdown()` ordering, or a crash), the parent's `_read_images` thread `recv()`s the `MemfdImage` and `to_image()` raises `FileNotFoundError`/`ConnectionRefusedError`. And since `payload.to_image()` sits *outside* the `try/except (EOFError, OSError)`, the reader thread dies with an unhandled traceback on what is a completely ordinary shutdown race. My first benchmark run hit exactly this before I added a keep-alive barrier. Fix: wrap the conversion and drop undeliverable frames silently during shutdown.
2. **One-shot semantics:** `detach()` can only be called once; a second `to_image()` on the same `MemfdImage` raises. Worth a docstring note since `MemfdImage` is exported in `rosys.vision.__all__`.
3. **`to_image()` hardcodes 3 channels** (`reshape(h, w, 3)`), while `from_array` silently accepts any array — a non-RGB array fails at the *receiver* with a confusing reshape error. Either assert `array.ndim == 3 and array.shape[2] == 3` in `from_array` or carry shape/dtype explicitly.
4. **fd budget:** every in-flight frame holds a dup'd fd in the child's resource sharer until the parent detaches. Bounded in steady state (the parent detaches immediately), but a stalled parent at 30 fps grows fds in the child. A latest-only handoff (see the #425 review) would cap this at 1.
5. **Tests:** none. A spawn-Pipe roundtrip test (send `MemfdImage`, receive, compare arrays; plus the sender-already-exited case from (1)) is small and CI-friendly on Linux; skip where `os.memfd_create` is absent.
6. The non-Linux fallback (ship the `Image` itself) keeps correctness — good.
7. Inherits everything from the #425 review, in particular the zombie-connected regression.

ftilde:
Closing since changes (hopefully) land as part of #425 directly.

PR #429, opened 2026-07-01
Add `scope` parameter to `on_repeat` and `Repeater`

falkoschindler:
### Motivation

Alternative to #427, solving the same problem with deterministic lifetimes instead of weak references.

`rosys.on_repeat` keeps its handler's object alive forever via the reference stored in the task — for page-owned objects this leaks one repeater per page load. `ui.timer` is not a substitute because repeaters run on RoSys time (simulated/accelerated), not wall clock. With client scoping, page-owned objects die with their client:

```python
class PageWidget:
    def __init__(self) -> None:
        rosys.on_repeat(self.update, 1.0, client_scoped=True)  # stops when the page's client is deleted

    def update(self) -> None:
        pass
```

Compared to the weak-reference approach in #427, scoping is deterministic and robust in ways weak references cannot be in this codebase:

- RoSys runs with `gc.disable()`, so weak-ref cleanup only happens when pure refcounting reaches zero — any accidental reference cycle (bound-method callbacks are the classic case, see `SimulatedCamera` ↔ `SimulatedDevice`) silently defeats it forever. Client scoping stops the repeater at a well-defined moment regardless of reference graphs. (To be precise: what is deterministic is tearing down the *behavior*; a cyclic object graph may still wait for a pressure-triggered collection. To keep deleted clients from pinning their owners, the repeater releases its handler on client deletion — thanks @evnchn for the analysis.)
- Weak references silently kill fire-and-forget modules whose only anchor was the repeater (`BackupSchedule()`, `NetworkMonitor()`). With client scoping being opt-in, every existing call site keeps today's behavior; nothing needs converting.
- No special handling for non-weakenable handlers (plain functions, partials, slots classes) — any callable works.

### Implementation

- `rosys.on_repeat(handler, interval, client_scoped=False)`:
  - default: process lifetime, exactly as before.
  - `client_scoped=True`: must be created within a UI context; the repeater stops when that client is deleted (raises `RuntimeError` outside a UI context). In NiceGUI script mode this includes the pre-flight client: the pre-flight repeater dies with it, while each per-connection re-execution binds to its own client.
- Context detection mirrors NiceGUI's own `Event.subscribe` idiom (`Slot.get_stack()` check before `context.client`, so no accidental script-mode entry). Teardown uses `client.on_delete(...)` without storing the client on the repeater (no reference cycle) and releases the handler on deletion.
- `CameraProjector` exposes the flag (`client_scoped=False` by default) instead of hardcoding a lifetime.
- Includes the uncontroversial lifecycle fixes from #427: finished tasks are pruned from `Repeater.tasks` via a done-callback and `stop()` uses `discard`. Additionally, `stop()` before or during startup removes the deferred start so a stopped repeater is not revived when startup replays its handlers.

### Design history: why there is no `'auto'`

Earlier revisions had a three-value `scope` parameter including `'auto'` (bind to the client if constructed in a UI context, app lifetime otherwise), and `CameraProjector` hardcoded it. @ftilde's demo in the discussion showed why that is a trap: a shared projector constructed lazily inside a page handler silently became client-scoped and froze app-wide after a page refresh. The general lesson: only the caller knows an object's intended lifetime, so a class must not infer it from where it happens to be constructed — and `'auto'` was exactly that inference, offered as a feature. It is gone; the lifetime is always an explicit choice, which also deleted the script-mode special case that existed only to keep `'auto'` from binding to the doomed pre-flight client. With only two lifetimes left, the parameter then shrank to a single `client_scoped` flag. Thanks @ftilde for the pushback.

### Progress

### DIFF
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -13,7 +13,6 @@ from typing import Any, ClassVar, Literal
 import numpy as np
 import psutil
 from nicegui import Client, Event, app, background_tasks, context, ui
-from nicegui import core as nicegui_core
 from nicegui.slot import Slot
 
 from . import core, helpers, run
@@ -145,28 +144,23 @@ def _run_handler(handler: Callable) -> None:
         log.exception('error while starting handler "%s"', handler.__qualname__)
 
 
-def _resolve_scope(scope: Literal['app', 'client', 'auto'], handler: Callable) -> Client | None:
+def _resolve_scope(scope: Literal['app', 'client'], handler: Callable) -> Client | None:
     """Return the client whose deletion should end the handler's lifetime (None for app lifetime)."""
     if scope == 'app':
         return None
     # NOTE: check the stack first, because accessing `context.client` outside a UI context would enter script mode
     if Slot.get_stack():
-        client = context.client
-        # NOTE: with 'auto' we must not bind to the script-mode client, which is deleted before startup;
-        # with 'client' binding to it is correct: the pre-flight repeater dies with it,
+        # NOTE: this may be the script-mode client, which is deleted before startup;
+        # binding to it is correct: the pre-flight repeater dies with it,
         # while each per-connection re-execution binds to its own client
-        if scope == 'client' or client is not nicegui_core.script_client:
-            return client
-        return None
-    if scope == 'client':
-        raise RuntimeError(f'cannot bind repeater for {handler} to a client outside of a UI context')
-    return None
+        return context.client
+    raise RuntimeError(f'cannot bind repeater for {handler} to a client outside of a UI context')
 
 
 class Repeater:
     tasks: ClassVar[set[asyncio.Task]] = set()
 
-    def __init__(self, handler: Callable, interval: float, *, scope: Literal['app', 'client', 'auto'] = 'app') -> None:
+    def __init__(self, handler: Callable, interval: float, *, scope: Literal['app', 'client'] = 'app') -> None:
         self.handler: Callable | None = handler
         self.interval = interval
         self._task: asyncio.Task | None = None
@@ -244,12 +238,11 @@ class Repeater:
             repeater.cancel()
 
 
-def on_repeat(handler: Callable, interval: float, *, scope: Literal['app', 'client', 'auto'] = 'app') -> Repeater:
+def on_repeat(handler: Callable, interval: float, *, scope: Literal['app', 'client'] = 'app') -> Repeater:
     """Repeatedly call a handler with a given interval.
 
     With ``scope='app'`` the repeater runs for the lifetime of the app.
     With ``scope='client'`` it must be created within a UI context and stops when that client is deleted.
-    With ``scope='auto'`` it binds to the client if created within a UI context and to the app otherwise.
     """
     repeater = Repeater(handler, interval, scope=scope)
     repeater.start()
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -144,31 +144,19 @@ def _run_handler(handler: Callable) -> None:
         log.exception('error while starting handler "%s"', handler.__qualname__)
 
 
-def _resolve_scope(scope: Literal['app', 'client'], handler: Callable) -> Client | None:
-    """Return the client whose deletion should end the handler's lifetime (None for app lifetime)."""
-    if scope == 'app':
-        return None
-    # NOTE: check the stack first, because accessing `context.client` outside a UI context would enter script mode
-    if Slot.get_stack():
-        # NOTE: this may be the script-mode client, which is deleted before startup;
-        # binding to it is correct: the pre-flight repeater dies with it,
-        # while each per-connection re-execution binds to its own client
-        return context.client
-    raise RuntimeError(f'cannot bind repeater for {handler} to a client outside of a UI context')
-
-
 class Repeater:
     tasks: ClassVar[set[asyncio.Task]] = set()
 
-    def __init__(self, handler: Callable, interval: float, *, scope: Literal['app', 'client'] = 'app') -> None:
+    def __init__(self, handler: Callable, interval: float, *, client_scoped: bool = False) -> None:
         self.handler: Callable | None = handler
         self.interval = interval
         self._task: asyncio.Task | None = None
         self._client_deleted = False
-        client = _resolve_scope(scope, handler)
-        if client is not None:
+        if client_scoped:
+            if not Slot.get_stack():  # NOTE: check the stack; accessing `context.client` would enter script mode
+                raise RuntimeError(f'cannot bind repeater for {handler} to a client outside of a UI context')
             # NOTE: not storing the client avoids a reference cycle that would outlive its deletion
-            client.on_delete(self._handle_client_delete)
+            context.client.on_delete(self._handle_client_delete)
 
     def _handle_client_delete(self) -> None:
         self._client_deleted = True
@@ -238,13 +226,12 @@ class Repeater:
             repeater.cancel()
 
 
-def on_repeat(handler: Callable, interval: float, *, scope: Literal['app', 'client'] = 'app') -> Repeater:
+def on_repeat(handler: Callable, interval: float, *, client_scoped: bool = False) -> Repeater:
     """Repeatedly call a handler with a given interval.
 
-    With ``scope='app'`` the repeater runs for the lifetime of the app.
-    With ``scope='client'`` it must be created within a UI context and stops when that client is deleted.
+    A client-scoped repeater must be created within a UI context and stops when that client is deleted.
     """
-    repeater = Repeater(handler, interval, scope=scope)
+    repeater = Repeater(handler, interval, client_scoped=client_scoped)
     repeater.start()
     return repeater
 
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -13,6 +13,7 @@ from typing import Any, ClassVar, Literal
 import numpy as np
 import psutil
 from nicegui import Client, Event, app, background_tasks, context, ui
+from nicegui import core as nicegui_core
 from nicegui.slot import Slot
 
 from . import core, helpers, run
@@ -148,8 +149,11 @@ def _resolve_scope(scope: Literal['app', 'client', 'auto'], handler: Callable) -
     """Return the client whose deletion should end the handler's lifetime (None for app lifetime)."""
     if scope == 'app':
         return None
-    if Slot.get_stack():  # NOTE: check the stack first; accessing `context.client` outside a UI context would enter script mode
-        return context.client
+    # NOTE: check the stack first, because accessing `context.client` outside a UI context would enter script mode
+    if Slot.get_stack():
+        client = context.client
+        if client is not nicegui_core.script_client:  # NOTE: the script-mode client is deleted before startup
+            return client
     if scope == 'client':
         raise RuntimeError(f'cannot bind repeater for {handler} to a client outside of a UI context')
     return None
@@ -162,24 +166,25 @@ class Repeater:
         self.handler = handler
         self.interval = interval
         self._task: asyncio.Task | None = None
-        self._client = _resolve_scope(scope, handler)
-        if self._client is not None:
-            self._client.on_delete(self.stop)
+        self._client_deleted = False
+        client = _resolve_scope(scope, handler)
+        if client is not None:
+            # NOTE: not storing the client avoids a reference cycle that would outlive its deletion
+            client.on_delete(self._handle_client_delete)
+
+    def _handle_client_delete(self) -> None:
+        self._client_deleted = True
+        self.stop()
 
     def start(self) -> None:
-        if self.running:
+        if self.running or self._client_deleted:  # a client-scoped repeater must not outlive its client
             return
         if _state.startup_finished:
             self._task = background_tasks.create(self._repeat())
             self.tasks.add(self._task)
-            self._task.add_done_callback(self._handle_task_done)
+            self._task.add_done_callback(self.tasks.discard)
         elif self.start not in startup_handlers:
             startup_handlers.append(self.start)
-
-    def _handle_task_done(self, task: asyncio.Task) -> None:
-        self.tasks.discard(task)
-        if self._task is task:  # not a stale callback from a task replaced by a restart
-            self._task = None
 
     async def _repeat(self) -> None:
         await sleep(self.interval)  # NOTE delaying first execution so not all actors rush in at the same time
@@ -259,7 +264,7 @@ async def startup() -> None:
 
     _state.startup_finished = True
 
-    for handler in startup_handlers:
+    for handler in list(startup_handlers):  # NOTE: snapshot, because a handler can mutate the list via Repeater.stop()
         _run_handler(handler)
 
 
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -265,7 +265,8 @@ async def startup() -> None:
     _state.startup_finished = True
 
     for handler in list(startup_handlers):  # NOTE: snapshot, because a handler can mutate the list via Repeater.stop()
-        _run_handler(handler)
+        if handler in startup_handlers:  # NOTE: a previous handler may have removed this one via Repeater.stop()
+            _run_handler(handler)
 
 
 async def _garbage_collection() -> None:
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -152,8 +152,12 @@ def _resolve_scope(scope: Literal['app', 'client', 'auto'], handler: Callable) -
     # NOTE: check the stack first, because accessing `context.client` outside a UI context would enter script mode
     if Slot.get_stack():
         client = context.client
-        if client is not nicegui_core.script_client:  # NOTE: the script-mode client is deleted before startup
+        # NOTE: with 'auto' we must not bind to the script-mode client, which is deleted before startup;
+        # with 'client' binding to it is correct: the pre-flight repeater dies with it,
+        # while each per-connection re-execution binds to its own client
+        if scope == 'client' or client is not nicegui_core.script_client:
             return client
+        return None
     if scope == 'client':
         raise RuntimeError(f'cannot bind repeater for {handler} to a client outside of a UI context')
     return None
@@ -163,7 +167,7 @@ class Repeater:
     tasks: ClassVar[set[asyncio.Task]] = set()
 
     def __init__(self, handler: Callable, interval: float, *, scope: Literal['app', 'client', 'auto'] = 'app') -> None:
-        self.handler = handler
+        self.handler: Callable | None = handler
         self.interval = interval
         self._task: asyncio.Task | None = None
         self._client_deleted = False
@@ -175,6 +179,9 @@ class Repeater:
     def _handle_client_delete(self) -> None:
         self._client_deleted = True
         self.stop()
+        # NOTE: the client never clears its delete handlers, so release the handler to not pin its owner;
+        # this is safe because stop() has cancelled the task and _client_deleted blocks any restart
+        self.handler = None
 
     def start(self) -> None:
         if self.running or self._client_deleted:  # a client-scoped repeater must not outlive its client
@@ -187,24 +194,26 @@ class Repeater:
             startup_handlers.append(self.start)
 
     async def _repeat(self) -> None:
+        handler = self.handler
+        assert handler is not None  # NOTE: the handler is only released after the task has been cancelled
         await sleep(self.interval)  # NOTE delaying first execution so not all actors rush in at the same time
         while True:
             start = time()
             try:
                 if is_stopping():
-                    log.info('%s must be stopped', self.handler)
+                    log.info('%s must be stopped', handler)
                     break
-                await invoke(self.handler)
+                await invoke(handler)
                 dt = time() - start
             except (asyncio.CancelledError, GeneratorExit):
                 return
             except Exception:
                 dt = time() - start
-                log.exception('error in "%s"', self.handler.__qualname__)
+                log.exception('error in "%s"', handler.__qualname__)
                 if self.interval == 0 and dt < 0.1:
                     delay = 0.1 - dt
                     log.warning(
-                        f'"{self.handler.__qualname__}" would be called to frequently ' +
+                        f'"{handler.__qualname__}" would be called to frequently ' +
                         f'because it only took {dt*1000:.0f} ms; ' +
                         f'delaying this step for {delay*1000:.0f} ms')
                     await sleep(delay)
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -12,7 +12,8 @@ from typing import Any, ClassVar, Literal
 
 import numpy as np
 import psutil
-from nicegui import Client, Event, app, background_tasks, ui
+from nicegui import Client, Event, app, background_tasks, context, ui
+from nicegui.slot import Slot
 
 from . import core, helpers, run
 from .config import Config
@@ -143,13 +144,27 @@ def _run_handler(handler: Callable) -> None:
         log.exception('error while starting handler "%s"', handler.__qualname__)
 
 
+def _resolve_scope(scope: Literal['app', 'client', 'auto'], handler: Callable) -> Client | None:
+    """Return the client whose deletion should end the handler's lifetime (None for app lifetime)."""
+    if scope == 'app':
+        return None
+    if Slot.get_stack():  # NOTE: check the stack first; accessing `context.client` outside a UI context would enter script mode
+        return context.client
+    if scope == 'client':
+        raise RuntimeError(f'cannot bind repeater for {handler} to a client outside of a UI context')
+    return None
+
+
 class Repeater:
     tasks: ClassVar[set[asyncio.Task]] = set()
 
-    def __init__(self, handler: Callable, interval: float) -> None:
+    def __init__(self, handler: Callable, interval: float, *, scope: Literal['app', 'client', 'auto'] = 'app') -> None:
         self.handler = handler
         self.interval = interval
         self._task: asyncio.Task | None = None
+        self._client = _resolve_scope(scope, handler)
+        if self._client is not None:
+            self._client.on_delete(self.stop)
 
     def start(self) -> None:
         if self.running:
@@ -157,8 +172,14 @@ class Repeater:
         if _state.startup_finished:
             self._task = background_tasks.create(self._repeat())
             self.tasks.add(self._task)
+            self._task.add_done_callback(self._handle_task_done)
         elif self.start not in startup_handlers:
             startup_handlers.append(self.start)
+
+    def _handle_task_done(self, task: asyncio.Task) -> None:
+        self.tasks.discard(task)
+        if self._task is task:  # not a stale callback from a task replaced by a restart
+            self._task = None
 
     async def _repeat(self) -> None:
         await sleep(self.interval)  # NOTE delaying first execution so not all actors rush in at the same time
@@ -188,13 +209,15 @@ class Repeater:
                 return
 
     def stop(self) -> None:
+        if self.start in startup_handlers:  # a deferred start must not revive the repeater at startup
+            startup_handlers.remove(self.start)
         if not self._task:
             return
 
         if not self._task.done():
             self._task.cancel()
 
-        self.tasks.remove(self._task)
+        self.tasks.discard(self._task)
         self._task = None
 
     @property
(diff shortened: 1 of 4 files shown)


falkoschindler:
*Disclaimer: this review was written by Claude (Fable 5) via a multi-agent review pass — 8 finder angles, one adversarial verifier per finding.*

## Overview

The PR adds a `scope='app'|'client'|'auto'` parameter to `rosys.on_repeat`/`Repeater` so client-created repeaters stop when their NiceGUI client is deleted (deterministic alternative to the weak-ref approach in #427), makes `CameraProjector` the first `scope='auto'` consumer, and bundles lifecycle fixes (done-callback task pruning, `stop()` via `discard`, removal of a deferred start on pre-startup `stop()`). The mechanism is sound and well-tested, but verification confirmed two significant correctness gaps (both around the new `stop()`/client-binding interactions), a GC-related lifetime issue, and several smaller items. All 8 findings below were CONFIRMED by adversarial verification; the top two were reproduced end-to-end against the pinned NiceGUI 3.13.0.

## Findings (most severe first)

1. **`rosys/rosys.py:212` — `stop()` mutates `startup_handlers` while `startup()` iterates the live list, silently skipping the next startup handler.** `startup()` runs `for handler in startup_handlers: _run_handler(handler)` with no copy; a synchronous startup handler that stops a not-yet-started repeater triggers `startup_handlers.remove(self.start)` mid-iteration, and CPython's position-based iterator then skips the following handler. Reproduced live: with `[Repeater.start ×3, stopper, victim]`, `victim` never ran — a subsystem silently never initializes. Fix: iterate a snapshot (`for handler in list(startup_handlers)`) in `startup()`, or mark-instead-of-remove in `stop()`.

2. **`rosys/rosys.py:151-167` — `_resolve_scope` copies NiceGUI's `Event.subscribe` idiom but omits its `core.is_script_mode_preflight()` guard, so script-mode apps silently kill `scope='auto'` repeaters.** In NiceGUI 3.13.0 script mode (module-level `ui.*` then `ui.run()` — exactly the shape of `docs/examples/areas/main.py` etc.), module-level code runs inside the entered script client, so a `CameraProjector` constructed after the first UI element binds to it; `ui.run()` then unconditionally deletes that client before startup (`ui_run.py:159-160`), `stop()` removes the deferred start, and the app-lifetime projector instance never runs — silently, contradicting the docstring's "binds … to the app otherwise". NiceGUI's own `event.py:94` guards exactly this with `not core.is_script_mode_preflight()`. Fix: return `None` from `_resolve_scope` (for `'auto'`) when `core.is_script_mode_preflight()`, mirroring `event.py`.

3. **`rosys/rosys.py:167` — the `client.on_delete(self.stop)` registration is never deregistered, pinning stopped repeaters and leaking cyclic garbage under `gc.disable()`.** Two facets, both verified against NiceGUI 3.13.0: (a) a repeater stopped manually while its client lives stays strongly referenced by the live client's `delete_handlers` — the handler and everything it captures are pinned for the client's remaining lifetime, and even a `gc.collect()` cannot reclaim reachable objects (long-lived dashboard clients are the typical RoSys deployment); (b) after client deletion, `delete_handlers` is never cleared, so `Repeater._client → Client → delete_handlers → Repeater.stop` is an unreachable cycle that refcounting can't free — with `gc.disable()` it lingers until the <300 MB pressure collection. Fix: in `stop()`, remove `self.stop` from `self._client.delete_handlers` and null `self._client`.

4. **`rosys/rosys.py:169` — `start()` after client deletion revives a client-scoped repeater with unbounded app lifetime.** The binding is one-shot: `Client.delete()` fires `delete_handlers` once and the client is gone, but `start()` checks only `self.running`, so a held reference restarted later (pause/resume patterns — restart-after-stop is an explicitly tested feature) runs forever with nothing left to stop it. Fix: guard `start()` with `self._client is not None and self._client.is_deleted` (skip or raise).

5. **`tests/test_repeater.py:91-111` — `test_stop_before_startup_prevents_deferred_start` has no failure-safe teardown and poisons the rest of the session.** `reset_after_test()` only runs if all asserts pass; `reset_before_test()` does not clear `_state.startup_finished`. Reproduced: injecting a failure makes every subsequent `rosys_integration` test error with `RuntimeError('should be only executed once')`, burying the real failure. The test genuinely can't use the fixture (it needs pre-startup state), so wrap the body in `try/finally` or use a small local yield-fixture.

6. **`rosys/rosys.py:179-182` — `_handle_task_done`'s `_task`-nulling and stale-callback guard are redundant.** `running`, `start()`, and `stop()` are all indifferent to a done-but-non-null `_task`, and nothing else reads it. `self._task.add_done_callback(self.tasks.discard)` is behaviorally equivalent, deletes the method plus the guard, and eliminates the restart race by construction (`test_repeater_can_restart_after_stop` still passes).

7. **`rosys/core.py:14` — `core.on_repeat`'s annotation `Callable[[Callable, float], Repeater]` was not updated for the new keyword-only `scope` parameter.** Callers through the `core` indirection (persistence layer) can't pass `scope` without mypy rejecting it, and the two entry points now state diverging contracts. A `Protocol` with the kw-only param would keep them interchangeable.

8. **`rosys/rosys.py:151` — comment line is 128 characters, exceeding the 120-column house limit** (AGENTS.md "House style": "…120-column lines"). Neither ruff (E501 ignored) nor autopep8 (doesn't rewrap comments) will catch it, so it lands permanently unless fixed in review.

## Notes (not findings)

- The `CameraProjector` `scope='auto'` change also means a _shared_ projector lazily constructed inside a page builder dies with that first client, freezing projections for all other consumers — that is the intended semantics of `'auto'`, but worth a sentence in the docstring or PR description as a usage caveat.
- Checked and clean: `Slot.get_stack()`/`context.client`/`Client.on_delete` exist and behave as assumed in the pinned NiceGUI; the `stop_all()` iteration is safe against done-callbacks (they run via `call_soon`); bound-method equality for `startup_handlers` membership works; no in-repo `on_repeat` call sites pass positional extras; `ui.timer` cannot replace `scope='client'` (wall-clock vs. rosys time).

falkoschindler:
Thanks — two of these were real catches. All findings except 7 are addressed in 14eb9023:

1. **Startup iteration skip:** fixed — `startup()` now iterates a snapshot (`list(startup_handlers)`), with a regression test that reproduces the skipped-handler scenario.
2. **Script-mode client binding:** fixed, but with a sharper guard than suggested. The global `is_script_mode_preflight()` check turned out to be too coarse — under pytest, an earlier test can leave `core.script_mode = True` with `app.is_started` false forever, which made the guard misclassify a *real* client (our own test suite caught this). Instead, `_resolve_scope` now compares the resolved client against `core.script_client` by identity, which targets exactly the one throwaway client that is deleted before startup.
3. **Client pinning/cycle:** addressed structurally — the `Repeater` no longer stores the client at all. It registers `_handle_client_delete`, which sets a flag and stops; with no `Repeater → Client` edge there is no cycle to leak under `gc.disable()`. (The suggested fix — removing `self.stop` from `delete_handlers` inside `stop()` — would mutate the very list `Client.delete()` is iterating, i.e. the same iterator bug as finding 1, so I avoided list surgery entirely.) The residual is the same one NiceGUI's `Event.subscribe` accepts: a manually stopped repeater stays referenced by its *live* client until that client is deleted.
4. **Revive after client deletion:** fixed via the flag from 3 — `start()` returns early once the client is deleted, asserted in the client-scope test.
5. **Test teardown:** both fixture-less tests now wrap their body in `try/finally` around `reset_after_test()`.
6. **`_handle_task_done` simplification:** taken — it's now a plain `add_done_callback(self.tasks.discard)`; `stop()` remains the only place that clears `_task`.
7. **`core.on_repeat` annotation:** intentionally left as is. No caller through the `core` indirection passes `scope` today, and we prefer keeping `core.py` minimal over a speculative `Protocol`; we'll add one when a `core`-level caller actually needs `scope`.
8. **Line length:** the comment was rewrapped; I also added an explicit >120 check to my local routine since none of the configured linters catches long comments.

The `CameraProjector` caveat from the notes is a fair point — happy to add a sentence to its docstring if you think it's worth it.

*Disclaimer: written by Claude (Fable 5).*

evnchn:
*Posted by Claude Code on @evnchn's behalf — a fresh-lineage review pass (a non-Claude model, Codex/GPT-5, did an independent adversarial read; the load-bearing finding below I re-verified by running it on this branch).*

**TL;DR:**

**Verdict — the `scope=` mechanism here is the better of the two approaches, but not merge-ready as-is.** Under RoSys's `gc.disable()`, #427's weak-refs can't collect the cyclic UI objects both PRs target (conceded in that thread: *"at some point we have to collect garbage"*), so its central case never fires. This PR instead **deterministically stops the ticking** on client deletion, which works regardless of reference cycles — that's the real win.

Two things stand between it and merge:

1. **One confirmed correctness bug (blocking):** the new `startup()` snapshot revives a repeater that was *explicitly stopped during startup*. Reproduced end-to-end on `14eb902`; one-line fix below passes the repro **and** all 7 of this PR's own tests.
2. **One claim to soften:** "deterministic … robust in ways weak references cannot be" is true for *stopping the behavior*, but a residual memory pin remains under `gc.disable()` (details below). Still strictly better than #427 — just not leak-proof.

**Ask (for @falkoschindler):** does *"land this PR's `scope=` mechanism (after the two fixes) as the API, and separately preserve @ftilde's leak-site fixes from #427, with credit"* match your intent? Those leak fixes are orthogonal to the API choice and correct regardless of which PR wins.

<details>
<summary><b>Finding 1 (blocking) — <code>startup()</code> snapshot revives an explicitly-stopped repeater · CONFIRMED (ran it on 14eb902)</b></summary>

The review feedback changed `startup()` to iterate a snapshot (`for handler in list(startup_handlers)`) to fix the *"stopping mid-iteration skips the next handler"* bug. That fixes the skip — but introduces the reverse: if an **earlier** startup handler calls `repeater.stop()` on a *pre-startup* repeater whose deferred `repeater.start` sits **later** in the snapshot, `stop()` removes `repeater.start` from the live list, but the loop is still walking the stale snapshot, so it calls `repeater.start` anyway. `startup()` has already set `_state.startup_finished = True`, so `start()` takes the `background_tasks.create(...)` branch and the stopped repeater runs.

This violates the invariant asserted by the PR's own `test_stop_before_startup_prevents_deferred_start` — but that test stops *before* `startup()`, so the removal predates the snapshot and it passes. The failing case is stop **during** startup, in this ordering.

**Repro** (drop in `tests/test_repeater.py`, run under pytest so `is_test` is set):

```python
async def test_finding1_snapshot_revives_stopped_repeater():
    from nicegui import core
    core.loop = asyncio.get_event_loop()
    rosys.reset_before_test()
    try:
        holder: dict = {}
        def stopper() -> None:
            holder['repeater'].stop()
        rosys.on_startup(stopper)                        # lands BEFORE the deferred start
        repeater = rosys.on_repeat(lambda: None, 0.01)   # its .start is appended AFTER stopper
        holder['repeater'] = repeater
        assert startup_handlers.index(stopper) < startup_handlers.index(repeater.start)
        await rosys.startup()
        assert not repeater.running, 'stopped repeater was revived by the startup() snapshot'
    finally:
        await rosys.shutdown()
        rosys.reset_after_test()
```

**Output on `14eb902` (unpatched):**
```
order in startup_handlers: stopper@1  repeater.start@2
repeater.running after startup: True     ← revived
FAILED — AssertionError: stopped repeater was revived by the startup() snapshot
```

**Fix** — keep the snapshot (it fixes the skip), but skip handlers a prior handler removed:
```python
for handler in list(startup_handlers):
    if handler in startup_handlers:  # a prior handler may have removed this one via Repeater.stop()
        _run_handler(handler)
```

**After the fix:** `repeater.running after startup: False` (repro passes) **and** `tests/test_repeater.py` → `7 passed` (no regression on the original skip-bug test). *Scope: this proves the revive and the fix in the isolated harness; it does not exercise a real subsystem's startup ordering — but the ordering is reachable whenever any `on_startup` handler stops a pre-startup repeater.*
</details>

<details>
<summary><b>Finding 2 (soften the claim) — residual memory pin under <code>gc.disable()</code> · verified by code inspection, not a runtime leak demo</b></summary>

`Repeater.__init__` registers `client.on_delete(self._handle_client_delete)`. In NiceGUI 3.13.0, `Client.delete()` (`nicegui/client.py:453`) fires the delete handlers but **never clears `self.delete_handlers`** (it's initialized at `client.py:125`, appended at `340`, iterated at `459`, never reset). So after deletion the chain `client.delete_handlers → _handle_client_delete → repeater → handler → owner` persists, and if the client-scoped owner holds any NiceGUI element (element → `.client` → back to the client — the norm for UI-context objects), that's a `client ↔ owner` cycle that refcounting can't break under `gc.disable()`.

**Why this is still better than #427, not a regression:** the repeater's *ticking is stopped* deterministically by `_handle_client_delete → stop()` — the actual CPU/behavior leak is gone. What remains is a *memory* pin, and only for cyclic owners. #427 stops neither for those owners. So the approach wins; the wording *"robust in ways weak references cannot be"* just wants a caveat that it's the **behavior** that's deterministically torn down, not necessarily the object graph.

**Clean fix:** break the chain on teardown — e.g. null `self.handler` in `_handle_client_delete` after `stop()` — so the never-cleared `delete_handlers` entry no longer transitively pins the owner.
</details>

<details>
<summary><b>Finding 3 (nit) — <code>on_delete</code> registration never deregistered on manual <code>stop()</code> · code inspection</b></summary>

A client-scoped repeater `stop()`ed while its client is still alive stays referenced by `client.delete_handlers` until the client dies, so it (and its handler's owner) can't be reclaimed earlier. Bounded — it dies with the client — so low severity; worth a mention only alongside finding 2, since the same "break the chain / allow deregistration" fix covers both.
</details>

<details>
<summary><b>What #427 gets right — preserve it regardless of which PR lands</b></summary>

@ftilde's PR did a real leak audit that is **orthogonal to the `weak=` vs `scope=` question**: the `UsbDevice.__del__`-never-fires observation, the `SimulatedDevice`-on-disconnect leak, and the provider teardown fixes are correct bugs independent of the `on_repeat` API. Whichever mechanism lands, those explicit-`.stop()` fixes should come with it, credited to @ftilde. The mechanism recommendation here is *not* "discard #427" — it's "adopt this PR's lifetime API, keep #427's leak-site fixes."
</details>

<details>
<summary><b>Provenance — what was independently checked vs. traced</b></summary>

- **Fresh-lineage pass:** the diff was handed to a non-Claude model (Codex / GPT-5) for an adversarial correctness read, *because* this PR and its inline review were both Claude-authored (same-lineage self-review is weak). Finding 1 originated there.
- **Independently re-verified by me (evnchn's session):** finding 1 **ran on `14eb902`** (repro + fix + `7 passed` regression check, above); findings 2–3 traced against NiceGUI 3.13.0 source (`nicegui/client.py` line refs cited).
- **Not done:** no runtime demo of the finding-2 cyclic pin (asserted from source, labeled as such); #427 not independently re-run (its blocker is conceded in-thread).
- `_resolve_scope` checks out on the stated semantics — `Slot.get_stack()` before `context.client`, correct `script_client` fallback to app-lifetime, `scope='client'` raises rather than binding to a doomed client. No wrong-client case found.
</details>

falkoschindler:
All open points are addressed on the branch (8bf324d7, 91cd9a02); thanks @evnchn for the fresh-lineage pass.

- **Finding 1 (snapshot revival):** fixed with the proposed membership check, plus the flipped-order regression test (`test_stopping_repeater_during_startup_does_not_revive_it`).
- **Finding 2 (memory pin):** the repeater now releases its handler on client deletion, so the never-cleared `delete_handlers` entry no longer pins the handler's owner; the PR description is softened accordingly (deterministic teardown of the behavior, not the object graph).
- **Preflight `scope='client'`:** instead of raising during the pre-flight run, `'client'` now binds to the script-mode client — the pre-flight instance dies with it (correct client semantics), and per-connection re-executions bind to their own clients. `'auto'` still falls back to app lifetime.
- **On the #427 question:** yes — the intent is to land the `scope=` mechanism as the API and preserve @ftilde's orthogonal leak-site fixes (`UsbDevice`, `SimulatedDevice`, provider teardown) separately, with credit.

*Disclaimer: written by Claude (Fable 5).*

ftilde:
Hmm, didn't we discuss this approach in person, @falkoschindler ? I'm not sure if we want it, because it's hugely surprising that a ~~CameraProvider~~ CameraProjector (as an example) may stop working (ticking) depending on where it was constructed. Demo: Open the following app, create a ~~CameraProvider~~ CameraProjector and then refresh the page. The step counter doesn't increase anymore.

```python
#!/usr/bin/env python3
from nicegui import ui

from rosys.vision import CameraProjector, SimulatedCalibratableCamera, SimulatedCameraProvider

provider = SimulatedCameraProvider()
provider.add_camera(SimulatedCalibratableCamera.create_calibrated(id='cam'))


class CountingProjector(CameraProjector):
    step_count = 0

    async def step(self) -> None:
        await super().step()
        CountingProjector.step_count += 1


# global, app-wide -- but instantiated on demand inside a page handler
projector: CameraProjector | None = None


def create_projector() -> None:
    global projector
    if projector is None:
        projector = CountingProjector(provider, interval=0.5)


@ui.page('/')
def index() -> None:
    ui.button('Create global projector', on_click=create_projector)
    ui.label().bind_text_from(CountingProjector, 'step_count', lambda n: f'projector steps: {n}')
    ui.label().bind_text_from(globals(), 'projector',
                              lambda p: 'projector: none' if p is None else 'projector: alive')


ui.run(title='repeater scope bug', reload=False)

```

falkoschindler:
You're right, @ftilde — and your pushback ended up simplifying the design beyond the immediate fix (fad4425d, b66feb8b). Your demo shows a class guessing its lifetime from where it happens to be constructed, and `'auto'` was exactly that guess offered as a feature — so it's gone entirely, not just from `CameraProjector` (which now defaults to `'app'`, so your demo keeps ticking across refreshes; regression test included). Dropping `'auto'` also deleted the script-mode special case in `_resolve_scope`, which only existed to protect the guess from binding to the doomed pre-flight client.

What remains is minimal: `scope='app'` (default, process lifetime — every existing call site unchanged) or `scope='client'` (must be created in a UI context, dies with that client). The mechanism still addresses the original leak — one repeater per page load for page-owned objects — which `'app'` can't fix and which weak references under `gc.disable()` can't reliably fix either. The PR description now documents this design history, with credit.

*Disclaimer: written by Claude (Fable 5).*

falkoschindler:
Small follow-up in 36296d37: with only two lifetimes left, the `scope` parameter shrank to a boolean — `rosys.on_repeat(handler, interval, client_scoped=True)`. That in turn made `_resolve_scope` pointless (there is no scope left to resolve; the binding is four lines in `Repeater.__init__`) and let the tests consolidate, taking ~50 lines out of the PR. The description is updated accordingly — earlier comments referring to `scope='client'` read as `client_scoped=True` now.

*Disclaimer: written by Claude (Fable 5).*

ftilde:
I'm still leaning towards #427: This PR does nothing against the leaks of various system-side classes, e.g. CameraProviders. Example where this PR fails:

Imagine a page where the user can select between a USB-CameraProvider and a RTSP-CameraProvider. Each time we change and construct a new Provider we leak one and it's repeater. While one could argue that that is a constructed scenario, the pattern of repeaters started with classes in fairly common in rosys and shows up in various places.

falkoschindler:
@ftilde Before answering the provider-swap example: I think we have been arguing from unstated assumptions, so let me make them explicit — push back on any of these, because agreeing on them decides the PR question almost mechanically.

**1. When does GC actually run?** Refcounting always; reference _cycles_ only when `_garbage_collection` sees free memory below `config.garbage_collection_mbyte_limit` (default 300 MB, checked every 60 s). On a machine with a few GB of free RAM that threshold may never trip within a process lifetime — so for cyclic objects, "eventually" ranges from minutes to never. Unconventional, but deliberate status quo.

**2. Are we worried about leaks _until_ or _beyond_ GC?** Worth noting: neither PR leaks anything _beyond_ a collection — #427's `WeakMethod` clears at collect time and the loop exits; #429 releases its handler on client deletion and residual cycles die at the pressure collection. The entire difference lives in the _until_ window, which per 1 is unbounded. So here is the principle I propose we adopt: **occupying memory until the pressure collection is the accepted price of `gc.disable()`; running _behavior_ (a ticking repeater, a capturing device, image emission) until the pressure collection is not acceptable** — it burns the single event loop and may effectively never stop. A lifetime mechanism should therefore be judged by whether it tears down _behavior_ deterministically. Weak references tie the behavior clock to the memory clock — the very clock `gc.disable()` made unbounded. That, not taste, is the core argument for scoping and explicit teardown over `weak=`.

**3. Is not storing an object a user error?** Today it isn't a communicated convention: the docs model fire-and-forget (`BackupSchedule()`), and nothing warns. The convention you're implying — "if you drop a RoSys object, expect it to stop working" — may well be reasonable, but adopting it is a breaking, to-be-documented decision, and its enforcement should be loud (warning or error), not "your backups silently stopped". Until we make that decision explicitly, the framework has to keep unreferenced modules working, which rules out the blanket `weak=True` conversion.

**4. Who decides an object's lifetime?** Its creator, explicitly. Not the construction context (that was `'auto'`, removed after your demo), and not the reference graph (that's `weak=`, unreliable per 1). Having deliberately disabled timely GC, RoSys cannot use the reference graph as a lifetime oracle — lifetimes must be stated, not inferred.

Measured against these, your provider-swap example actually argues _against_ weak references:

- The swapped-out provider is pinned first of all by its own repeater, not by a mysterious cycle: `Repeater.tasks → task → _repeat coroutine → handler → provider` is a strong, _acyclic_ chain, so no amount of weak back-edges inside the provider graph frees anything while the repeater runs.
- With `weak=True` (as on your branch) the repeater no longer pins it — but the provider _is_ cyclic: `camera.NEW_IMAGE.subscribe(self.NEW_IMAGE.emit)` gives every camera a strong edge back to its provider, on top of the camera ↔ device cycle we already verified. Per 1, the weak handler never clears, and the swapped-out provider keeps scanning and ticking until a pressure collection that may never come. Your own scenario is not fixed by `weak=` for exactly these classes.
- The same limitation runs through the leak fixes on your branch: they are all `weak=True` conversions that resurrect `__del__`-based cleanup (your diagnosis that the repeater pinned the device and `__del__` never fired is spot on). But `__del__` only fires once refcounting frees the device, and the device ↔ camera cycle (`UsbDevice` holds the camera's bound `_handle_new_image_data`, `SimulatedDevice` its `_on_new_image`) means that only happens on the paths where `disconnect()` already breaks the cycle by dropping `self.device`. On the drop-without-disconnect paths — `remove_camera` is a bare `del self._cameras[id]`, and your provider swap drops everything at once — the cycle survives, the weak handler stays alive, and the 100 Hz capture loop keeps ticking. So even within your PR, the ingredient that actually works is the explicit disconnect step; `weak=` only piggybacks on it. Swapping a provider is a lifetime decision made by the code doing the swap (principle 4), so that code needs an explicit teardown to call.

Which leads to the concrete proposal:

1. **Fix the leaks #427 uncovered in their own PR, and merge that first.** To be transparent about what this means: content-wise, little of #427's implementation would survive — the fixes on that branch are `weak=True` conversions, so redoing them as explicit teardown (stop the capture task in `disconnect()`/`release_capture()` instead of relying on `__del__`, let `remove_camera` disconnect so the drop path is covered too) is new code; only the `__qualname__` hardening carries over as is, and the task pruning is already part of this PR. What survives — and it's the durable contribution of #427 — is the audit: the leak sites and the repeater-pins-owner/`__del__`-never-fires diagnosis are your findings, and they are bugs on `main` today, independent of which API wins. That's also why this deserves its own PR rather than being folded in here: these fixes shouldn't be held hostage by the mechanism debate. Ideally you author that PR yourself, @ftilde — you know the sites best.
2. **Land this PR** (`client_scoped=`) for the page-owned case — the one lifetime where "explicit teardown" would mean every page hand-wiring `client.on_delete`, which is precisely what the flag packages.
3. **Open an issue for dynamically replaced app-level modules** (your swap scenario): a teardown API on `CameraProvider` & friends that stops repeaters, disconnects devices and unsubscribes — deterministic, covers the cyclic parts, testable. Weak child→parent back-edges can come on top as hygiene, so refcounting reclaims most of the graph immediately after teardown — but as an optimization over teardown, not a substitute for it (per 2, they only accelerate the memory side, which was never the problem).
4. **`weak=` stays out**: per 2 it doesn't stop behavior deterministically, per 3 its blanket application silently breaks legitimate fire-and-forget usage, and per 1 its central use case — cyclic UI owners — can't fire at all.

If you disagree with one of the four assumptions above, let's argue about _that_ — it's the actual fork in the road.

_Disclaimer: researched and written by Claude (Fable 5). The weak-handler-under-`gc.disable()` behavior (acyclic collected, cyclic not) was verified by executing #427's `_weaken` on the branch; the cycle edges (`NEW_IMAGE` subscription, `_handle_new_image_data`/`_on_new_image` back-references, bare `del` in `remove_camera`) were verified by reading the code, not by a runtime demo._

ftilde:
> Is not storing an object a user error? Today it isn't a communicated convention: the docs model fire-and-forget (BackupSchedule()), and nothing warns

I *still* want a source for that. The only thing I found is the following

> If you want to automatically keep daily backups, you can use the `BackupSchedule` module.

... and I don't see how that suggests it's fine not to keep a handle to it. In fact, the GC vs. non-GC. lifetime discussion shows the insanity of this: With the current rosys state and not keeping a handle to `BackupSchedule` it appears to work until eventually GC runs. Good luck finding that issue.

> camera.NEW_IMAGE.subscribe(self.NEW_IMAGE.emit) also "leaks" without GC

That's a fair point, but doesn't mean that we cannot improve the current state of affairs from the repeater side imho.
My point is not (mainly) that we need to fix issues in camera cleanup immediately (though there may be value to that), but that the pattern of repeaters in `__init__` is common in rosys and we should probably replace it (at least) with `on_repeat(... weak=True)`.

ftilde:
I think the general problem/conflict is

1) We are drawn to (imho clean, but in python flawed/difficult) [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) patterns. Or we don't think enough about cleanup (which would be not as necessary when applying RAII). The pain is that it kind of works until we (accidentally) create cycles or use similar mechanisms like the classvar registration of repeater.
2) Due to our "realtime" requirements we disable (or postpone) GC which makes it really obvious (hundreds of "leaked" objects) that RAII doesn't really work in python.

I'm a little bit lost on how we go from here. The immediate problem of `CameraProjector` is solved by both PRs. #427 solves this a _bit_ more generally, since the underlying problem is not only client vs. server context (which is hard to make out sometimes), but can appear solely on the server-side as well.

ftilde:
Thinking about it some more, #427 is the better fix, this one is not:

```python
from nicegui import ui

import rosys
from rosys.rosys import Repeater
from rosys.vision import CameraProjector, SimulatedCameraProvider

camera_provider = SimulatedCameraProvider(auto_scan=False)
projector = CameraProjector(camera_provider)


def set_interval(interval: float) -> None:
    global projector
    # NOTE: the only way to change the interval is a new projector; the old repeater keeps running forever
    projector = CameraProjector(camera_provider, interval=interval)


ui.label('CameraProjector interval')
ui.slider(min=0.1, max=2.0, step=0.1, value=1.0, on_change=lambda e: set_interval(e.value)) \
    .props('label-always') \
    .style('width: 300px')
count_label = ui.label()
ui.timer(0.5, lambda: count_label.set_text(f'{len(Repeater.tasks)} repeaters running'))
ui.label('baseline is 3 (2 rosys-internal + 1 initial projector); every slider change adds one forever') \
    .classes('text-caption text-grey')

ui.run(title='CameraProjector repeater leak')
```

We could fix this immediate problem by allowing the repeater's interval to be changed, but that's not the point. It's not at all obvious that this code is problematic. Why would I not be allowed to create a new CameraProjector?

ftilde:
>... and I don't see how that suggests it's fine not to keep a handle to it. In fact, the GC vs. non-GC. lifetime discussion shows the insanity of this: With the current rosys state and not keeping a handle to BackupSchedule it appears to work until eventually GC runs. Good luck finding that issue.

Btw. I was wrong here. BackupSchedule *is* leaked and thus keeps running without `weak=True`. But it shows the same problem. It should be fine to replace a BackupSchedule and then don't have multiple ones running.

(Note that the difference/my confusion is/was that Repeaters don't leak by reference cycle, but by global registration. So not even GC cleans them up.)

> With weak=True (as on your branch) the repeater no longer pins it — but the provider is cyclic: camera.NEW_IMAGE.subscribe(self.NEW_IMAGE.emit) gives every camera a strong edge back to its provider, on top of the camera ↔ device cycle we already verified.

This isn't a cycle, by the way, since `emit` is a method of the event, not of `CameraProvider`.

falkoschindler:
Let's close this PR in favor of the updated PR #427.

PR #431, opened 2026-07-04
Exclude GPIO0 from the strapping-pin guard when flashing Core or P0

### LINKED ISSUE #430: flash_p0 is blocked by the strapping-pin guard whenever GPIO0 idles high — which is its normal run state
## What happened

On our RB60 (Jetson Orin, NAND/v0.5 piggyboard, Lizard v0.12.0), clicking **Flash P0** in the Robot Brain developer UI always fails with

> Flashing P0 failed. Check strapping pins.

## Why

`LizardFirmware.flash_p0` (and `flash_core`, line 130) guards with

```python
if any(await self.robot_brain.esp_pins_p0.get_strapping_pins()):
```

and `EspPins.get_strapping_pins()` returns the levels of GPIO **[0, 2, 12]** (esp_pins.py line 87). But GPIO0 idles **high** in normal operation (internal pull-up; low would mean bootloader mode), so the guard can never pass on our brain:

```
p0: GPIO_Status[0]|  Level: 1| Pullup: 1| Pulldown: 0   ← blocks the guard, but is the normal run state
p0: GPIO_Status[2]|  Level: 0| Pullup: 0| Pulldown: 1   ← fine
p0: GPIO_Status[12]| Level: 0| Pullup: 0| Pulldown: 1   ← fine
```

According to zauberzeug/lizard#75 and the fix in zauberzeug/lizard#95, `p0.flash()` **sets GPIO0 to the correct state itself**; only **GPIO2** (bootloader entry) and **GPIO12** (flash voltage: high = 1.8 V) are relevant preconditions.

## Evidence

Bypassing the rosys guard and sending `p0.flash()` directly over the serial line worked flawlessly with the exact pin states above: `replicator: Replica complete.`, P0 went from v0.8.0 to v0.12.0.

## Suggested fix

Restrict the guard to GPIO2 and GPIO12 (e.g. give `get_strapping_pins()` a `numbers=[2, 12]` parameter or drop GPIO0 from the check in the two flash guards).

Environment: rosys main (077008a9), Lizard v0.12.0, Robot Brain RB60 (Jetson Orin, NAND/v0.5 piggyboard).

falkoschindler:
### Motivation

Clicking **Flash P0** (and likewise flashing the Core) always failed with "Flashing P0 failed. Check strapping pins." because the guard in `LizardFirmware.flash_p0`/`flash_core` checks GPIO **0**, 2 and 12 — but GPIO0 idles **high** in normal operation (internal pull-up), so the `any(...)` guard could never pass. Per zauberzeug/lizard#75 and zauberzeug/lizard#95, Lizard pulls the boot pin (GPIO0) low itself during `flash()`; only GPIO2 (bootloader entry) and GPIO12 (flash voltage) are genuine external preconditions that RoSys can pre-check.

Fixes #430.

### Implementation

`EspPins.get_strapping_pins()` gets a `numbers: Sequence[int] = (0, 2, 12)` parameter, keeping its current behavior for existing callers (e.g. the "Check Strapping Pins" developer menu). The two flash guards in `LizardFirmware` now call it with `numbers=(2, 12)`, with a `NOTE` comment explaining why GPIO0 cannot be pre-checked from the outside. This is option 1 from the issue discussion — the minimal change that keeps the early UI feedback instead of surfacing Lizard's own error after the fact.

### Progress

- [x] Pytests have been added (or are not necessary). *(hardware-only module without simulation twin; no existing tests for `EspPins`/`LizardFirmware`)*

### DIFF
--- a/rosys/hardware/esp_pins.py
+++ b/rosys/hardware/esp_pins.py
@@ -81,12 +81,12 @@ class EspPins:
         await self.robot_brain.send(f'{self.name}.set_pin_level({pin.gpio}, {1 if level else 0})')
 
     async def get_strapping_pins(self, numbers: Sequence[int] = (0, 2, 12)) -> list[bool]:
-"""Get the state of the strapping pins.
+        """Get the state of the strapping pins.
 
-       See https://github.com/zauberzeug/lizard/issues/75 and https://github.com/zauberzeug/lizard/pull/95.
-       
-       :param numbers: GPIO numbers to check (default: strapping pins 0, 2 and 12)
-       """
+        See https://github.com/zauberzeug/lizard/issues/75 and https://github.com/zauberzeug/lizard/pull/95.
+
+        :param numbers: GPIO numbers to check (default: strapping pins 0, 2 and 12)
+        """
         return [await self.get_pin_level(number) for number in numbers]
 
     def developer_ui(self) -> None:
--- a/rosys/hardware/esp_pins.py
+++ b/rosys/hardware/esp_pins.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


rosys/hardware/esp_pins.py
```diff
@@ -79,12 +80,12 @@ async def get_pin_level(self, pin_id: int) -> bool:
     async def set_pin_level(self, pin: GpioPin, level: bool) -> None:
         await self.robot_brain.send(f'{self.name}.set_pin_level({pin.gpio}, {1 if level else 0})')
 
-    async def get_strapping_pins(self) -> list[bool]:
+    async def get_strapping_pins(self, numbers: Sequence[int] = (0, 2, 12)) -> list[bool]:
         """Get the state of the strapping pins.
 
         See https://github.com/zauberzeug/lizard/issues/75 and https://github.com/zauberzeug/lizard/pull/95.
         """
```
pascalzauberzeug:
```suggestion
    async def get_strapping_pins(self, numbers: Sequence[int] = (0, 2, 12)) -> list[bool]:
"""Get the state of the strapping pins.

       See https://github.com/zauberzeug/lizard/issues/75 and https://github.com/zauberzeug/lizard/pull/95.
       
       :param numbers: GPIO numbers to check (default: strapping pins 0, 2 and 12)
       """
```

pascalzauberzeug (dismissed):
I suggest to add `:param` to the doc string, looks good beyond that!

pascalzauberzeug (dismissed):

pascalzauberzeug (approved):

PR #432, opened 2026-07-08
Verify Lizard startup checksum when the ESP connects

jsb-zz:
### Motivation

When deploying a robot — during setup, testing or in the field — the only automatic indication that the Lizard startup code on the microcontroller is outdated is the core message field-count check in `RobotHardware.update`. That check misses any change that keeps the number of output fields the same (e.g. changed pins or module parameters). The user should be made aware on startup and after configuring, without interrupting operation.

### Implementation

- Subscribe new `RobotBrain._check_lizard_code` to `ESP_CONNECTED`, which fires both on startup and after `configure()`; on mismatch, show a short negative `rosys.notify` with warning log level — no exception, the robot keeps running
- Reuse `LizardFirmware.read_local_checksum()` and `read_core_checksum()` (`core.startup_checksum()`) for the comparison and expose the result as new `LizardFirmware.checksums_match` property (`True`/`False`/`None` if unknown), so apps can bind UI elements (e.g. a status icon) to it
- Fix premature `ESP_CONNECTED` after `restart()`: waiting for a single line after `core.restart()` was not enough — buffered core messages from the ESP that is about to reboot re-established the hardware time, so `ESP_CONNECTED` fired while the ESP was going down and the checksum check ran into the void; now `restart()` waits until core messages cease before clearing the hardware time
- Reset `core_checksum` to `None` at the beginning of `read_core_checksum()` so a timeout can no longer leave a stale value from a previous read
- Add `tests/test_robot_brain.py` with a `CommunicationSimulation` that establishes an ESP connection and answers `core.startup_checksum()`, covering the matching and mismatching case as well as the configure-restart-reconnect flow with a buffered core message (regression test for the premature `ESP_CONNECTED`)

Tested on real hardware (Field Friend f21): outdated startup code is reported on startup, configuring resolves it and the check runs again after the ESP reconnects.

### Progress

---
⬛ claude-fable-5

### DIFF
--- a/rosys/hardware/lizard_firmware.py
+++ b/rosys/hardware/lizard_firmware.py
@@ -108,7 +108,7 @@ class LizardFirmware:
             self.log.error('Could not read startup checksum from Core. Robot Brain is not ready.')
             return
         deadline = rosys.time() + 5.0
-        while rosys.time() < deadline:
+        while rosys.time() < deadline and self.robot_brain.is_ready:
             if response := await self.robot_brain.send_and_await('core.startup_checksum()', 'checksum:', timeout=1):
                 self.core_checksum = response.split()[-1]
                 self.log.info('core checksum: %s', self.core_checksum)
--- a/rosys/hardware/lizard_firmware.py
+++ b/rosys/hardware/lizard_firmware.py
@@ -96,6 +96,7 @@ class LizardFirmware:
         self.log.info('local checksum: %s', self.local_checksum)
 
     async def read_core_checksum(self) -> None:
+        self.core_checksum = None
         if not self.robot_brain.is_ready:
             self.log.error('Could not read startup checksum from Core. Robot Brain is not ready.')
             return
--- a/rosys/hardware/lizard_firmware.py
+++ b/rosys/hardware/lizard_firmware.py
@@ -38,8 +38,8 @@ class LizardFirmware:
 
     @property
     def checksums_match(self) -> bool | None:
-        """Whether the local and core startup checksums match, or ``None`` if the core checksum is unknown."""
-        if self.core_checksum is None:
+        """Whether the local and core startup checksums match, or ``None`` if either checksum is unknown."""
+        if self.local_checksum is None or self.core_checksum is None:
             return None
         return self.local_checksum == self.core_checksum
 
--- a/rosys/hardware/lizard_firmware.py
+++ b/rosys/hardware/lizard_firmware.py
@@ -36,6 +36,13 @@ class LizardFirmware:
         self._p0_flash_last_complete: float = 0.0
         self.robot_brain.FLASH_P0_COMPLETE.subscribe(lambda: setattr(self, '_p0_flash_last_complete', rosys.time()))
 
+    @property
+    def checksums_match(self) -> bool | None:
+        """Whether the local and core startup checksums match, or ``None`` if the core checksum is unknown."""
+        if self.core_checksum is None:
+            return None
+        return self.local_checksum == self.core_checksum
+
     async def read_all(self) -> None:
         await self.read_online_version()
         self.read_local_version()
--- a/rosys/hardware/robot_brain.py
+++ b/rosys/hardware/robot_brain.py
@@ -46,6 +46,7 @@ class RobotBrain:
         self.communication = communication
         self.lizard_code = ''
         self.lizard_firmware = LizardFirmware(self)
+        self.ESP_CONNECTED.subscribe(self._check_lizard_code)
 
         self.waiting_list: dict[str, str | None] = {}
         self._clock_offset: float | None = None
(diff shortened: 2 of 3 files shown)


evnchn:
Some shower thoughts:

Huh, will it cause the boot to be slow? Will it cause the flash wear? 

Assuming the flash bit flip is not a likely scenario (flashing already verify the hash, I think? If not we can verify ourselves) then actually we don't need a hash but just a nonce.

Imagine, flash-time, we assign a UUID to the build. Then host can bail on wrong UUID reported by Lizard. 

However, where do we store the expected nonce? But I imagine where do we store the expected hash is also a problem this has to solve either way. 

---

I'm going to dispatch agent to answer this question. If you want to know it earlier, consider do it yourself 😉

evnchn:
Dispatched-agent report 🤖 — answering the shower thoughts against the actual RoSys + Lizard source.

**TL;DR:** No boot slowdown, no flash wear (this PR writes nothing to flash), and — nice instinct — but the nonce is actually the *wrong* tool here, and the "where do we store the expected value" problem is precisely the one the current checksum **avoids**. Details below.

---

### 1. Boot slow? — No, negligible.

`core.startup_checksum()` sums an `std::string` that's **already resident in RAM** (`Storage::startup`, loaded from NVS once at `Storage::init()`), so the ESP-side compute is microseconds. Host side likewise sums an in-memory string. It's subscribed to `ESP_CONNECTED` and `await`ed off the event loop, so it doesn't block the robot coming online.

One honest caveat: `read_core_checksum()` has a **5 s bounded retry window** (1 s `send_and_await` in a loop) — so in the happy path it's a single round-trip, but if the ESP *doesn't* answer it'll occupy that handler for up to 5 s before logging "could not verify" and moving on. Non-blocking for the robot, but not literally "one command" in the sad path.

### 2. Flash wear? — No, this PR adds **zero** flash writes.

Reading the checksum never touches flash. The only NVS write is `save_startup()` (`nvs_set_str` + `nvs_commit`), which fires on `configure()` — **pre-existing behaviour, unchanged by this PR.** So the feature is read-only w.r.t. flash.

### 3. Hash vs nonce — first, it isn't even a hash.

Both sides compute a **16-bit additive checksum** of the startup string — not a crypto hash, not a CRC:

| side | code |
|---|---|
| host | `sum(ord(c) for c in lizard_code) % 0x10000` |
| ESP | `uint16_t checksum = 0; for (char c : Storage::startup) checksum += c;` |

So it's already a cheap change-detector, not an integrity guarantee (it's order-insensitive and collision-prone — see the folded note). Your bit-flip worry mostly evaporates because we were never buying integrity here anyway.

Now the **nonce idea specifically**: a UUID assigned *at flash time* fingerprints the **firmware build**, not the startup script. But the startup script is set/changed at **`configure()` time, independently of flashing** (flash the interpreter once → RoSys streams the script → Lizard persists it to NVS). A flash-time nonce wouldn't change when someone reconfigures pins/params — which is *exactly* the drift this PR is trying to catch. So it can't do the job. To make a nonce work you'd have to regenerate + re-store it every time the script changes… at which point it's no longer a flash-time nonce, it's just a worse content hash.

### 4. Where do we store the expected value? — the checksum is *why we don't have to.*

This is the crux, and I think it flips your conclusion. Because the checksum is **content-derived**, the host stores **nothing**: RoSys recomputes the expected value on the fly from `lizard_code`, which it already holds in memory (it's the code RoSys itself generated and sent). ESP side reuses `Storage::startup`, already in NVS to *run* the code — no new storage either.

A **nonce is content-independent**, so it would *force* the host to persist "the nonce I expect" and keep it in sync with the code. That's the storage/sync burden you were worried about — and it's introduced by the nonce, not solved by it. The current design sidesteps it entirely. 👍

---

<details>
<summary>Two latent notes for the PR (not blockers)</summary>

**(a) Encoding / `char` signedness mismatch.** Python `ord(c)` yields Unicode code points; the ESP sums `char` over a `std::string` (byte-wise, and `char` signedness is platform-defined). For pure-ASCII startup scripts (which Lizard configs are, in practice) both agree. But any byte ≥ 128 — a stray non-ASCII char in a config — could make host and ESP disagree and fire a **false** "differs from expected" warning. Worth a one-line assumption/guard if non-ASCII is ever possible.

**(b) Collision strength.** A 16-bit additive sum is order-insensitive (`"ab"` == `"ba"`) and a `+1/−1` char swap collides. Fine for a best-effort, non-blocking operator heads-up. If you ever want it to be a real "is the firmware what I expect" guard, **CRC32 is nearly free and order-sensitive** — strictly better as an *accidental*-change detector. (It's still not tamper-resistant; if adversarial integrity were ever the goal, that's a different tool.)

**(c) Parsing.** `response.split()[-1]` is permissive — a malformed line ending in some other token could be accepted as the checksum. Very low stakes given the `checksum:` prefix match, just noting.

</details>

<details>
<summary>Source evidence</summary>

- Host: `rosys/hardware/lizard_firmware.py` → `read_local_checksum` (16-bit sum), `read_core_checksum` (5 s retry loop, `core_checksum = None` reset added by this PR).
- Firmware: `lizard/main/modules/core.cpp` → `startup_checksum` method (`uint16_t += c`).
- Firmware: `lizard/main/storage.cpp` → `Storage::startup` is a RAM `std::string` loaded once via `Storage::get()` at `init()`; writes only via `save_startup()` (`nvs_set_str`/`nvs_commit`).

Cross-checked with a fresh non-Claude model lineage (Codex): confirmed the read-only-flash, nonce-can't-detect-script-drift, and no-host-storage claims; it flagged the encoding/signedness point in (a).

</details>

pascalzauberzeug (commented):
Do we want to keep [the old check](https://github.com/zauberzeug/rosys/blob/e91f628bf901455ae775eb9b7b668d140db5d6f5/rosys/hardware/robot.py#L85-L93) or remove it?

jsb-zz (commented):
@pascalzauberzeug, i would keep it. The old check is a runtime parse-alignment guard that happens to warn; the checksum is a connect-time drift detector. Those are different scopes.  The old check is checking during run-time, so it provides a addtional value to the checksum check, which works at every connect.

falkoschindler (approved):
Thanks for the PR, @jsb-zz! I pushed a small follow-up commit:

- `read_core_checksum()` now also re-checks `is_ready` in its retry loop, so a disconnect during the ≤5 s retry window (e.g. triggering `configure()` right after connect, which is what the new notification asks for) ends with the normal error log instead of an unhandled `EspNotReadyException` in the event-handler task.
- The `connect()` helper comment in the test now says "on a _first_ connect" — `clock_offset` survives a restart, so on a reconnect `ESP_CONNECTED` already fires on the first core message.

Apart from that, the PR looks good to me. 🚀

PR #433, opened 2026-07-15
Speed up Pose3d matrix conversions

ftilde:
### Motivation

Profiling a robot application (py-spy flamegraphs) showed pose resolution as a major CPU cost — up to ~45% of process CPU with a 3D scene open, but also relevant without (pose attachment to images, odometry updates). Frame-chain resolution (`resolve`/`relative_to`/`__matmul__`) constantly builds 4×4 matrices, and `np.block` spends most of its time in pure-Python bookkeeping (`_block_setup`, `_block_check_depths_match`) rather than arithmetic. `np.block` probably only really shines for larger matrices.

### Implementation

- Update `Pose3d.matrix` and `Pose3d.inverse_matrix` to fill a preallocated `np.eye(4)` via slice assignment instead of `np.block` (`matrix`: 9.7 µs → 1.7 µs, ~5.7×)
- Update `inverse_matrix` to use `rotation.matrix.T` directly instead of constructing an intermediate `Rotation`
- Add `test_matrix`, `test_inverse_matrix` and `test_matmul_matches_matrix_product` covering the previously untested properties; results verified element-wise against the previous implementation

### Progress

---
⬛ claude-fable-5

### DIFF
--- a/rosys/geometry/pose3d.py
+++ b/rosys/geometry/pose3d.py
@@ -58,11 +58,18 @@ class Pose3d(Object3d):
 
     @property
     def matrix(self) -> np.ndarray:
-        return np.block([[self.rotation.matrix, self.translation_vector], [0, 0, 0, 1]])
+        M = np.eye(4)
+        M[:3, :3] = self.rotation.matrix
+        M[:3, 3] = self.translation
+        return M
 
     @property
     def inverse_matrix(self) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


NiklasNeugebauer (approved):
Stumbled upon this and was interested so I essentially reviewed it fully.

ftilde:
@NiklasNeugebauer Excellent, thanks!

PR #434, opened 2026-07-15
Don't run CameraProjector and CameraObjects when invisible

ftilde:
### Motivation

Profiling a robot application showed that the 3D scene's camera visualization dominates CPU usage even when it is not shown: `CameraObjects` is hidden by default in our app (`visible(False)`), but its update timer and the `CameraProjector` keep running for every client — resolving extrinsics, re-projecting ground grids, and updating textures nobody sees.

### Implementation

- Extend `CameraProjector` with `acquire()`/`release()`: a consumer count that starts the repeater on the first acquire and stops it when the last consumer releases, so a projector shared between clients only runs while at least one of them shows it
- Extend `CameraObjects` to hold the projector while visible: acquired on construction and `visible(True)`, released on `visible(False)` and on garbage collection (`__del__`), with an idempotent guard so repeated calls cannot unbalance the count
- Deactivate the `CameraObjects` update timer while the element is hidden
- Add a repeater test using `rosys_integration` and simulated time covering the full acquire/release/re-acquire lifecycle

### Progress

---
⬛ claude-fable-5

### DIFF
--- a/rosys/vision/camera_projector.py
+++ b/rosys/vision/camera_projector.py
@@ -26,10 +26,23 @@ class CameraProjector:
 
     def __init__(self, camera_provider: CalibratableCameraProvider, *, interval: float = 1.0) -> None:
         self.camera_provider = camera_provider
+        self._consumer_count = 0
+        self._repeater = rosys.Repeater(self.step, interval)
 
         self.projections: dict[str, Projection] = {}
 
-        rosys.on_repeat(self.step, interval)
+    def acquire(self) -> None:
+        """Register interest in up-to-date projections; the projector only runs while at least one consumer holds it."""
+        self._consumer_count += 1
+        if self._consumer_count == 1:
+            self._repeater.start()
+
+    def release(self) -> None:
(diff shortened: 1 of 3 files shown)


rosys/vision/camera_projector.py
pascalzauberzeug, line 1: We could also start and stop the repeater here, like we do with the ui timer.

ftilde: Yes, that works now! Done.

ftilde:
Sorry, claude decided to edit the commit and force push it when asked to implement the changes...

rosys/vision/camera_projector.py
pascalzauberzeug, line 1: Instead of `== 1` and `== 0` I would check for `Repeater.running`. `_consumer_count` is private, yes, but could be altered from the outside

pascalzauberzeug (changes requested):

pascalzauberzeug (approved):

PR #435, opened 2026-07-16
Add MCAP recording with Foxglove converters and recordings page

pascalzauberzeug:
### Motivation

Robots built on RoSys need a way to record sensor and state data for replay and analysis in Foxglove Studio. This recorder was developed and field-proven in an application on top of RoSys and only depends on RoSys types, so it moves upstream where every RoSys-based robot can use it.

### Implementation

- Add `rosys.analysis.recording` with `McapRecorder`: messages are enqueued on the event loop and written by a background consumer — encoding, ZSTD compression, and file I/O never block the loop, including the final drain in the async `stop()` — with size-based file rotation, disk budget enforcement, a backlog queue capped by message count and payload bytes (drop-oldest), per-recording topic selection, and a `developer_ui()`; `RECORDING_STARTED`/`RECORDING_STOPPED` fire per file
- Add converters from RoSys types (`Pose`, `Velocity`, `GnssMeasurement`, `BmsState`, `ImuMeasurement`, `Image`) to Foxglove well-known JSON schemas, with a type registry for auto-dispatch and `add_event_topic` / `add_timer_topic` / `add_pose_topic` helpers; app-specific topics use `custom_message` with a plain JSON schema
- Add `RecordingsPage`: a management UI at `/recordings` (list, filter by date, rename, delete, download via `/api/recordings/{name}`); recordings left unfinished by a crash are detected and rebuilt in place
- Outside the new package: a public `Image.detections_by_detector_id` accessor (so the annotation converter does not touch private state) and a type-only import fix in `hardware/imu.py` to break an import cycle
- Add a runnable documentation example under `docs/examples/mcap_recording/`
- Add the `mcap` and `zstandard` dependencies

### Progress

---
⬛ claude-fable-5


NiklasNeugebauer:
I am wondering if we could integrate this with the existing recorder/ replay (which writes to individual images) and a simple parameter for setting the backend / output?

pascalzauberzeug:
@NiklasNeugebauer I think that should be possible, but I think we should tackle that in another PR

falkoschindler (approved):
Due to the sheer size of this PR (+3888/-2) I focussed on everything outside the new recording module. Looks good to me. 👍🏻

Issue #436, opened 2026-07-17
Simulation actors register process-global repeaters/handlers with no per-instance disposal (memory leak in per-request embeddings)


evnchn:
*Filed by evnchn; drafted with Claude Code, repro run and claims verified against source.*

**TL;DR:**

RoSys actors that are natural to instantiate inside a NiceGUI page create **process-global** repeaters / shutdown handlers / routes with **no per-instance disposal path** — so in per-client or per-request embeddings they survive client cleanup and keep running until the process shuts down.

Verified examples (each discards the `Repeater` that `rosys.on_repeat(...)` returns, so no one can stop it):

- `RobotSimulation` — a 100 Hz repeater (`self.step`), handle discarded
- `Steerer` — a 20 Hz repeater, handle discarded
- `Odometer` — a 1 Hz repeater (`self.prune_history`), handle discarded, plus a retained event subscription
- `Automator` — a global `on_shutdown` handler closing over `self`, one per instance
- `Camera` / `SimulatedCamera` — global `/images/<id>/…` routes; `SimulatedDevice` keeps a repeater that `SimulatedCamera.disconnect()` never stops

For a single-robot app this is a non-issue. But anything that builds actors **per request** — interactive docs, a multi-tenant dashboard, any embedding where visitors come and go — leaks a fresh set of loops per visit, forever.

**Ask:** give these actors a per-instance disposal path (or bind their repeaters to the creating client's lifecycle) so they can be created and released per client. A draft PR with one concrete shape follows — glad to reshape it.

<details>
<summary><b>Runnable reproduction (RoSys 0.34.0, NiceGUI 3.14.0, Python 3.12)</b></summary>

`pip install rosys`, then `python leak.py`. The page builds a `RobotSimulation` + `Odometer` (a 100 Hz and a 1 Hz repeater); each page-view adds both and never releases them:

```python
import threading
import urllib.request

from nicegui import app, ui
import rosys                 # noqa: F401
import rosys.rosys as core

@ui.page('/')
def demo() -> None:          # a normal live "steer a simulated robot" demo
    from rosys.driving import Odometer, robot_object
    from rosys.geometry import Prism
    from rosys.hardware import RobotSimulation, WheelsSimulation
    wheels = WheelsSimulation()
    RobotSimulation([wheels])                                   # -> 100 Hz Repeater(self.step)
    with ui.scene():
[...]

PR #437, opened 2026-07-17
Drop a repeater's pending start when it is stopped before startup

evnchn:
**TL;DR:** Reduced to the two lines that are still needed after #427. A repeater created *before* startup defers its start into `startup_handlers`; `stop()` returned early while `self._task` was still `None`, so `startup()` replayed the deferred start and the stopped repeater ran anyway.

Your read was right: with #427 merged, the per-actor `tear_down()` machinery this PR originally proposed is redundant — a repetition now ends when its object is collected — so I dropped all of it and force-pushed onto current `main`. The diff is `rosys/rosys.py +2` plus one regression test.

```python
     def stop(self) -> None:
+        if self.start in startup_handlers:
+            startup_handlers.remove(self.start)  # queued but not yet started: cancel the pending start
         if not self._task:
             return
```

#427 does not cover this path: its `startup_handlers.clear()` runs *after* the replay, and its `_alive` check catches a *collected* object, not a *stopped* one.

<details>
<summary><b>The regression test, seen to fail without the fix</b></summary>

Added next to its #427 siblings in `tests/test_repeater.py`:

```python
async def test_stop_before_startup_prevents_the_deferred_start():
    # NOTE: a repeater created pre-startup defers its start; stopping it must also drop that pending start.
    core.loop = asyncio.get_event_loop()
    rosys.reset_before_test()
    assert not _state.startup_finished

    calls: list[float] = []
    try:
        ticker = Ticker(calls)  # kept alive, so a dead handler is not what stops it
        repeater = rosys.on_repeat(ticker.step, 0.01)
        assert repeater.start in startup_handlers

        repeater.stop()

        await rosys.startup()  # replays the deferred start handlers

        assert not repeater.running  # the stopped repeater was not started by the replay
        assert not calls
    finally:
        await rosys.shutdown()  # NOTE: don't let a failure above leak startup state into the next test
        rosys.reset_after_test()
```

With `rosys/rosys.py` stashed, i.e. the two lines reverted:

```
FAILED tests/test_repeater.py::test_stop_before_startup_prevents_the_deferred_start - assert not True
1 failed in 0.07s
```

With the fix: `22 passed`.

Gates on this commit (Python 3.13, macOS): `pytest` 338 passed / 7 skipped (skips are the `arp-scan` and Linux-only camera tests, none on this path), `mypy` clean on 217 source files, `pylint` 10.00/10, `pre-commit` clean.
</details>

<details>
<summary><b>Why the rest of this PR is gone — measured on post-#427 <code>main</code></b></summary>

One actor kind per page-view, 30 views, then `Client.prune_instances(client_age_threshold=0)` + `gc.collect()`:

| per-page actor | instances alive after client delete | live repeaters |
|---|---|---|
| `RobotSimulation` | 0 | back to baseline |
| `Steerer` | 0 | back to baseline |
| `Odometer` | 28 / 30 | elevated |
| `Automator` | 30 / 30 | back to baseline |
| `SimulatedCamera` | 0 | back to baseline |

So `RobotSimulation` and `Steerer` need nothing — their repetitions end with the object, exactly as #427 intends. And the surviving `Odometer` is not a repeater problem at all: it runs because the object is pinned (see below), so stopping its repeater by hand would have been treating the symptom.

The `Odometer` unsubscribe this PR also proposed turns out to be unnecessary for page-scoped actors: after client deletion `wheels.VELOCITY_MEASURED.callbacks` was **empty**, because NiceGUI 3.13's `Event.subscribe` auto-unsubscribes callbacks registered inside a UI context. (It does still pin outside a UI context, e.g. from a background task.)
</details>

<details>
<summary><b>What remains of #436</b></summary>

Two global-registry vectors, not five actors:

1. **`rosys.on_shutdown` holds strong references.** #427's lifetime rule reached `on_repeat` only. That list is the whole retention chain — `shutdown_handlers` → `WheelsSimulation` (`rosys.on_shutdown(self.stop)`) → `wheels.VELOCITY_MEASURED` → `Odometer`. Removing *only* the per-instance entries and re-collecting took odometers 3 → 0 and automators 3 → 0. Note `Automator.__init__` registers a **lambda**, which #427's own rule rejects for `on_repeat`, so that site needs rewriting rather than weakening.
2. **`app.routes` accumulates per distinct camera id** (3 entries each; 90 for 30 cameras). The camera *object* is already safe via `weakref.ref` in `create_image_route`, so this is registry growth, independent of vector 1.

Happy to write this up properly on #436 with the full measurements if useful.
</details>

Note on CI: `check-milestone` fails on every external PR until a maintainer assigns a milestone — not a code failure. Still a draft, as suggested.

_Disclaimer: measurements and this description by Claude (Opus 5), reviewed against a second model lineage._

### DIFF
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -275,6 +275,8 @@ class Repeater:
             log.debug('"%s" stops repeating because its object was garbage-collected', _handler_name(self.handler))
 
     def stop(self) -> None:
+        if self.start in startup_handlers:
+            startup_handlers.remove(self.start)  # queued but not yet started: cancel the pending start
         if not self._task:
             return
 
--- a/rosys/rosys.py
+++ b/rosys/rosys.py
@@ -325,7 +325,7 @@ async def startup() -> None:
 
     _state.startup_finished = True
 
-    for handler in startup_handlers:
+    for handler in list(startup_handlers):  # NOTE: a handler may stop a repeater, removing its pending start
         _run_handler(handler)
     startup_handlers.clear()  # NOTE: release the handlers so they don't keep their objects alive forever
 
--- a/tests/test_repeater.py
+++ b/tests/test_repeater.py
@@ -206,11 +206,65 @@ async def test_stop_before_startup_prevents_the_deferred_start():
         assert repeater.start in startup_handlers
 
         repeater.stop()
+        assert repeater.start not in startup_handlers  # the pending start was dropped
 
         await rosys.startup()  # replays the deferred start handlers
 
         assert not repeater.running  # the stopped repeater was not started by the replay
         assert not calls
+    finally:
+        await rosys.shutdown()  # NOTE: don't let a failure above leak startup state into the next test
+        rosys.reset_after_test()
+
+
+async def test_restart_before_startup_keeps_the_deferred_start():
+    # NOTE: stopping and starting again before startup must leave the start queued exactly once.
+    core.loop = asyncio.get_event_loop()
+    rosys.reset_before_test()
+    assert not _state.startup_finished
+
+    calls: list[float] = []
+    try:
+        ticker = Ticker(calls)
+        repeater = rosys.on_repeat(ticker.step, 0.01)
+
+        repeater.stop()
+        repeater.start()
+        assert startup_handlers.count(repeater.start) == 1  # re-queued, but not twice
+
+        await rosys.startup()  # replays the deferred start handlers
+
+        assert repeater.running  # the restarted repeater was started by the replay
+    finally:
+        await rosys.shutdown()  # NOTE: don't let a failure above leak startup state into the next test
+        rosys.reset_after_test()
+
+
+async def test_stopping_a_repeater_during_startup_keeps_the_other_handlers():
+    # NOTE: stop() removes a pending start from the list that startup() is iterating; it must not skip a sibling.
+    core.loop = asyncio.get_event_loop()
+    rosys.reset_before_test()
+    assert not _state.startup_finished
+
+    calls: list[float] = []
+    started: list[str] = []
+    try:
+        ticker = Ticker(calls)
+        repeater = rosys.on_repeat(ticker.step, 0.01)  # queued before the handlers below
+
+        def stopper() -> None:
+            repeater.stop()  # drops the earlier pending start, shortening the list
+
+        def sibling() -> None:
+            started.append('sibling')
+
+        rosys.on_startup(stopper)
+        rosys.on_startup(sibling)
+
+        await rosys.startup()  # replays the deferred start handlers
+
+        assert not repeater.running  # the stopped repeater was not started by the replay
+        assert started == ['sibling']  # ...and the handler behind it was not skipped
     finally:
         await rosys.shutdown()  # NOTE: don't let a failure above leak startup state into the next test
         rosys.reset_after_test()
--- a/tests/test_repeater.py
+++ b/tests/test_repeater.py
@@ -193,6 +193,29 @@ async def test_collected_object_does_not_start_orphan_task_after_startup():
         rosys.reset_after_test()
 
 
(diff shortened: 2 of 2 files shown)


evnchn:
Note on CI: the only failing check is `check-milestone`, which fails on every external PR until a maintainer assigns a milestone (I don't have push access to set one) — not a code failure. The test suite is push-triggered and doesn't auto-run on a fork PR, so for confidence I ran the full suite locally against this branch:

<details>
<summary><b>Full test suite against this branch — 176 passed, 7 skipped</b></summary>

```
$ pytest tests/ --ignore=tests/test_examples.py   # test_examples needs the shell/example toolchain
176 passed, 7 skipped in 10.37s
```

Python 3.12, RoSys `main` + this branch; includes the new `tests/test_actor_teardown.py`.
</details>

falkoschindler:
Thanks for the draft (and for catching the pre-startup `stop()` gap — that one is confirmed and also surfaced in the #429 review; we'll make sure it lands once, not twice). As outlined in #436, the repeater part of this PR will likely be superseded by the upcoming `Repeater` redesign, while the unsubscribe/teardown part stays relevant as its own topic. Suggest keeping this as a draft until the design issues are up — we'll ping you here.

_Disclaimer: written by Claude (Fable 5)._

falkoschindler (approved):
Thanks for cutting this down to the actual defect — the two-line version is exactly right, and the diagnosis matches: #427's `startup_handlers.clear()` runs after the replay, so a stopped-but-still-queued repeater slips through. Putting the removal in `stop()` is the right level too, since the `Repeater` is what queued the entry in the first place.

Reviewing it turned up one consequence worth pinning down, so I pushed a follow-up commit onto your branch rather than sending you round again.

`stop()` is the first code path in RoSys that _removes_ from `startup_handlers`, and `startup()` iterates that list directly:

```python
for handler in startup_handlers:
    _run_handler(handler)
```

So a **synchronous** startup handler that stops a repeater queued before it shifts the list under the iterator, and the following handler is silently skipped. Async handlers are safe — they are deferred to background tasks that only run after the loop finishes — so the window is narrow, but it fails quietly and the symptom points nowhere near the cause. Call sites that can reach `stop()` synchronously during the replay: `CameraProjector.release()`, `_TimerSource.stop()`, and `UsbDevice.__del__` (we run with `gc.disable()`, so `__del__` fires on refcount drop, inside whatever frame drops the last reference).

The fix is to replay a copy, plus a regression test that fails without it:

```python
for handler in list(startup_handlers):  # NOTE: a handler may stop a repeater, removing its pending start
    _run_handler(handler)
```

I also added `test_restart_before_startup_keeps_the_deferred_start` — stop-then-start before startup must re-queue exactly once, the other half of the behavior your change defines — and one assertion to your test (`repeater.start not in startup_handlers`) so it pins the mechanism rather than only the symptom.

Gates on the merged result: `pytest` 340 passed / 7 skipped, `mypy` clean on 217 files, `pylint` 10.00/10, `pre-commit` clean.

One thing I am knowingly leaving as is: a repeater stopped before startup keeps its `_warn_if_object_was_never_stored` check queued, so if its handler's object is then collected you get a _"will never be called … store the object"_ warning for a repeater that was deliberately stopped. Misleading, but harmless and rare enough that a flag on `Repeater` costs more than it buys.

And yes — please do write up the #436 findings on that issue. The `shutdown_handlers` retention chain and the `app.routes` growth are worth tracking separately, and the `Automator` lambda registration is a real wrinkle given #427's rule. The measurements alone are useful.

evnchn:
*Posted by Claude Code on evnchn's behalf.*

**Written up** — the #436 findings are on that issue now, with the two work items they decomposed into open as #443 (the `shutdown_handlers` retention chain, including the `Automator` lambda) and #444 (the camera image routes).

Also noted the queued `_warn_if_object_was_never_stored` check you're knowingly leaving as is — I won't file that separately.

PR #438, opened 2026-07-20
Restrict Lizard downloads and flashing to supported versions

pascalzauberzeug:
### Motivation

Some robots are wired in ways that only certain Lizard versions support — since Lizard 0.14.0, end stops of MotorAxes on p0 are no longer allowed, so existing robots with this wiring must not be flashed with 0.14.0 or newer. Flashing an incompatible version must be impossible, not just discouraged.

### Implementation

- Extend `RobotBrain` with `supported_lizard_versions`, a PEP 440 version specifier (e.g. `'<0.14.0'`) passed on to `LizardFirmware`; invalid specifiers fail fast at construction
- Add `LizardFirmware.is_version_supported()` and skip unsupported releases in `read_online_version()`, so they never appear in the version selection
- Refuse unsupported versions in `download()`, `flash_core()` (local version) and `flash_p0()` (core version), independent of the UI; with a restriction configured, unknown versions are refused as well
- Warn when `read_core_version()` or `read_p0_version()` report an already flashed unsupported version
- Add `packaging` dependency for specifier evaluation
- Add tests covering comparison bounds, combined specifiers, the unrestricted default, refusal to download/flash and the warning for already flashed versions

### Progress

⬛ claude-fable-5


pascalzauberzeug:
@falkoschindler I've done a double review with Claude, but I think you should have a look because it's a Lizard related decision.
I'm not really happy with handing `supported_lizard_versions` through `RobotBrain`, but using `**kwargs` is also not really helpful.

falkoschindler (approved):
During review we pushed a few simplifications directly to this branch:

- **Simplified the tests**: the specifier semantics are now a single parametrized test in which each case states its specifier explicitly (instead of relying on the fixture's bound), a `notifications` fixture replaces the repeated subscribe boilerplate, the six flash refusal/pass tests are merged into one parametrized `test_flash_version_gate`, and test names are shortened since the docstrings carry the explanation — 224 → 140 lines with more covered cases (18 → 23).
- **Polished docstrings and messages** in `LizardFirmware`.
- **Merged the version checks into a single `_refuse_version` gate** that also handles unknown versions, removing the duplicated `if self.supported_versions is not None:` blocks from `flash_core()`/`flash_p0()`; `flash_core()` now always reads the local version before flashing, so the "Flashing Lizard firmware None to Core..." notification can no longer occur.

With that, the implementation does exactly what the PR promises — the gate lives in `LizardFirmware` itself rather than the UI, unrestricted configurations keep the old behavior, and the tests cover bounds, refusal and pass-through paths. Approving.

PR #440, opened 2026-07-22
Compute the serial checksum over UTF-8 bytes

evnchn:
### Motivation

`augment()`/`check()` XOR Unicode **code points**, while Lizard XORs the raw **bytes** it receives (`checksum ^= buffer[i]` over a `char` buffer). The transport is explicitly UTF-8 — `serial_communication.py` does `.decode(errors='replace')` on read and `.encode()` on write — so the two sides disagree for any non-ASCII line.

Above U+00FF it's worse than a disagreement: `'{:02x}'` is a *minimum* width, not a truncation, so the checksum overflows to three hex digits and `check()` stops finding the `@`. rosys silently drops its own message — `augment`/`check` don't even round-trip with themselves:

```python
augment('grün')  ->  'grün@87'    # Lizard computes 04 over the UTF-8 bytes
augment('日本')  ->  '日本@2c9'    # three digits -> check() returns '' -> message dropped
```

Riding along, a one-line guard for a second defect in the same function: `check()` called `int(line[-2:], 16)` unguarded, so a corrupted suffix raised `ValueError` out of the serial drain loop (`read_lines()`). `Repeater` catches it and the robot recovers, so this is log-spam plus a dropped batch rather than a crash — but it's free to fix while here, and it's the same bug I just reported in the firmware (zauberzeug/lizard#242).

### Implementation

Both functions now XOR `line.encode()`. **For ASCII this is byte-for-byte identical**, so all existing traffic is unaffected; it only changes the cases that are already broken.

In `check()` the encode goes *inside* the `try`. `.encode()` raises `UnicodeEncodeError` (a `ValueError` subclass) on a lone surrogate, where the old `ord()` loop never raised — the serial path can't produce one (`errors='replace'` yields U+FFFD) but the socketio/JSON path can, and a PR whose point is "stop `check()` raising" shouldn't introduce a new raise. `check()` stays total.

`augment()` deliberately still raises: outbound data is rosys' own, and manufacturing a checksum for unencodable text would hide an application bug. Both behaviours are pinned by tests.

<details>
<summary><b>Verification</b></summary>

All six new test functions were confirmed to **fail on the unpatched code** (stashing the fix takes the file from 17 passed to 5 failed / 11 passed) — not green-only tests.

Gates from `AGENTS.md`, all green:

| gate | result |
|---|---|
| `uv run pre-commit run --all-files` | passed (ruff, autopep8, codespell, mdformat) |
| `uv run mypy ./rosys` | `Success: no issues found in 217 source files` |
| `uv run pylint ./rosys` | `10.00/10` |
| `uv run pytest` | `318 passed, 7 skipped` |
| `cd tests && uv run ./test_examples.py` | all `Ok` |

The 7 skips were read individually — all camera / `arp-scan` / Linux-USB, none related to this change.

</details>

<details>
<summary><b>Second-lineage review (Codex, verbatim)</b></summary>

> **Verdict: Most of this is sound.** I do not see a blocker in the diff as shown. **BLOCKER: None found.**
>
> **Wire compatibility:** "This is a breaking checksum change for non-ASCII messages, but I would classify it as a bugfix, not a protocol migration. The protocol peer is the ESP32 firmware, and the firmware computes over raw bytes. Python was wrong for UTF-8 payloads. For U+0100 and above, Python was not merely incompatible with firmware; it emitted variable-width checksum text and broke its own `check(augment(x))` contract."
>
> **Surrogate decision:** "Putting `.encode()` inside the `try` in `check()` is right. `check()` is an inbound validator and already returns `''` for malformed/corrupt input… The asymmetry with `augment()` is defensible: outbound messages are generated by rosys, and silently dropping or manufacturing a checksum for unencodable text would hide an application bug."
>
> **Fix Python or firmware?** "Fix Python. The firmware is doing the natural thing for a byte-oriented serial protocol… Changing firmware to compute Unicode code points would be unnatural in C++, ambiguous after decoding errors, and incompatible with the actual bytes on the wire."
>
> **Ready to ship upstream: 8/10** — with one extra test documenting `augment('\ud800')` raising, and a direct assertion that non-ASCII checksums are exactly two hex digits.

Both of its suggested tests were added (`test_augment_rejects_undecodable_line`, and the two-hex-digit assertion).

</details>

### Progress

### DIFF
--- a/rosys/hardware/lizard_firmware.py
+++ b/rosys/hardware/lizard_firmware.py
@@ -121,7 +121,7 @@ class LizardFirmware:
         self.log.error('Could not read Lizard version from P0')
 
     def read_local_checksum(self) -> None:
-        checksum = sum(ord(c) for c in self.robot_brain.lizard_code) % 0x10000
+        checksum = sum(self.robot_brain.lizard_code.encode()) % 0x10000
         self.local_checksum = f'{checksum:04x}'
         self.log.info('local checksum: %s', self.local_checksum)
 
--- a/rosys/hardware/robot_brain.py
+++ b/rosys/hardware/robot_brain.py
@@ -375,8 +375,8 @@ class EspNotReadyException(Exception):
 
 def augment(line: str) -> str:
     checksum = 0
-    for c in line:
-        checksum ^= ord(c)
+    for byte in line.encode():
+        checksum ^= byte
     return f'{line}@{checksum:02x}'
 
 
@@ -385,11 +385,14 @@ def check(line: str | None) -> str:
         return ''
     if line[-3:-2] != '@':
         return ''
-    check_ = int(line[-2:], 16)
-    line = line[:-3]
     checksum = 0
-    for c in line:
-        checksum ^= ord(c)
+    try:
+        check_ = int(line[-2:], 16)
+        line = line[:-3]
+        for byte in line.encode():
+            checksum ^= byte
+    except ValueError:
+        return ''
     if checksum != check_:
         return ''
     return line
--- a/tests/test_robot_brain.py
+++ b/tests/test_robot_brain.py
@@ -106,27 +106,22 @@ async def test_local_checksum_is_computed_over_utf8_bytes(robot_brain: RobotBrai
     assert robot_brain.lizard_firmware.local_checksum == '02c6'
 
 
-@pytest.mark.parametrize('line', ['hello', 'wheels.speed(1, 2)', 'grün', 'café', 'ß', '日本'])
-def test_checksum_round_trip(line: str) -> None:
+@pytest.mark.parametrize('line, checksum', [
+    ('hello', '62'),
+    ('wheels.speed(1, 2)', '47'),
+    ('grün', '04'),  # NOTE: XOR over the UTF-8 bytes, like Lizard, not over code points (which would give '87')
+    ('café', '0e'),
+    ('ß', '5c'),
+    ('日本', '02'),  # NOTE: code points would XOR to 0x2c9 and overflow the two-digit '{:02x}' format
+])
+def test_augment_and_check(line: str, checksum: str) -> None:
+    assert augment(line) == f'{line}@{checksum}'
     assert check(augment(line)) == line
 
 
-@pytest.mark.parametrize('line', ['hello', 'grün', '日本'])
-def test_checksum_is_computed_over_utf8_bytes(line: str) -> None:
-    checksum = 0
-    for byte in line.encode():  # NOTE: Lizard XORs the raw bytes it receives, not Unicode code points
-        checksum ^= byte
-    assert augment(line) == f'{line}@{checksum:02x}'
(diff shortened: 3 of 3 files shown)


evnchn:
**The one red check is the milestone policy gate, not the code** — `Milestone Assigned` fails with "No milestone assigned to the pull request." I don't have push access here, so that one needs a maintainer (recent merges are all on `0.35.0`).

`pytest.yml` is `workflow_call`-only, invoked from `on_push.yml`, so it doesn't run for a PR from a fork. I ran the equivalent locally on Linux across all three CI versions instead — **318 passed / 7 skipped on each**.

<details>
<summary><b>Linux runs — 3.11 / 3.12 / 3.13</b></summary>

`docker run python:3.X-slim` → `uv sync --all-extras` → `uv run pytest -q -rs`:

| Python | start method | result |
|---|---|---|
| 3.11.15 | `linux fork` | `318 passed, 7 skipped` |
| 3.12.13 | `linux fork` | `318 passed, 7 skipped` |
| 3.13 | `linux fork` | `318 passed, 7 skipped` |

Worth noting explicitly because my development machine is macOS (`spawn`); these runs use `fork`, matching CI. The 7 skips are all physical-hardware gates (`No USB camera detected`, `No MJPEG camera detected`, etc.), none related to this change.

Also green locally: `pre-commit run --all-files`, `mypy ./rosys` (217 files, no issues), `pylint ./rosys` (10.00/10), and `tests/test_examples.py`.

</details>

falkoschindler (approved):
Great find, @evnchn, and a very well-argued PR — the byte-vs-code-point analysis, the surrogate asymmetry and the red-first tests all check out, and I could reproduce all gates locally.

While reviewing I noticed the same code-point assumption in one more place: `LizardFirmware.read_local_checksum()` computed the _startup_ checksum as `sum(ord(c) for c in lizard_code) % 0x10000`, while the firmware's `core.startup_checksum()` sums the raw bytes of the stored script into a `uint16` (`char` is unsigned on Xtensa — `DEFAULT_SIGNED_CHAR 0` in gcc's `xtensa.h`). Your fix actually makes this reachable for the first time: previously non-ASCII Lizard code never survived `configure()`, now it does — and would then permanently trigger the "Lizard startup code is outdated" notification.

Since this PR effectively introduces non-ASCII support end to end, I pushed two commits onto your branch:

1. **Align the startup checksum as well** — same `encode()` approach in `read_local_checksum()`, mirrored test helper, plus a literal-value test pinning `'grün'` → `'02c6'`. For ASCII it is byte-for-byte identical, so the compatibility argument of your PR is unchanged.
2. **Consolidate your checksum tests around literal expected values** — the round-trip, byte-formula and two-digit tests collapsed into one parametrized table whose literal checksums are the firmware truth, and the undecodable line joined the corrupted-lines table. Same coverage, three tests instead of five, and nothing tautological left.

The remaining `ord()` twins live in the Lizard repo (`monitor.py`, `configure.py`, `examples/ros/scripts/lizard.py` and `examples/trajectory_planner/serial_connection.py`) — independent of RoSys, we'll tackle them over there.

PR #441, opened 2026-07-23
Add MAC prefix 40:7a:a4 for Dahua cameras

denniswittich:
Adds the MAC prefix `40:7a:a4` to the vendor mapping so Dahua cameras with this prefix are detected during RTSP camera discovery.


pascalzauberzeug (approved):
Because I place great trust in you, I am approving this PR without checking the MAC address.

PR #443, opened 2026-07-27
Derive the shutdown handler lifetime from the handler type

### LINKED ISSUE #436: Simulation actors register process-global repeaters/handlers with no per-instance disposal (memory leak in per-request embeddings)

evnchn:
**TL;DR:** Applies #427's lifetime rule to the sibling registry. `rosys.on_shutdown` held its handlers **strongly** in a process-global list, so every per-request module that registers one was pinned for the life of the process. Measured: 60 page-views → **60/60 `Automator` instances still alive 10 s after their clients were deleted**, permanently. With this change they are collected in ~1 s — and #436's headline scenario drops from **1088 MB to 220 MB** at 160 page-views (table below).

This is the part of #436 that #427 did not reach. Deliberately kept separate from #437 so that one stays reviewable.

**Two commits, and the second is separable** — drop it and the first still fixes the leak:

1. *Derive the shutdown handler lifetime from the handler type* — weakens bound methods, so the module is no longer pinned.
2. *Let a dead shutdown handler remove itself from the registry* — the entry itself no longer accumulates, so `shutdown_handlers` is bounded by the number of **live** modules instead of by total registrations. Measured over 500 page-views: the registry stays at **1 entry** (was 501) and the object graph grows by **1 object** (was 3,208), with RSS flat.

```python
 def on_shutdown(handler: Callable) -> None:
+    """Call the handler on shutdown; a bound method is held weakly, so a discarded object is not kept alive."""
+    if inspect.ismethod(handler):
+        try:
+            handler = _WeakHandler(handler)
+        except TypeError:
+            log.warning('"%s" stays alive until shutdown because its object does not support weak references',
+                        handler.__qualname__)
     shutdown_handlers.append(handler)
```

That is the whole change to the lifecycle: **+11/-1 lines of production code**, reusing #427's existing `_WeakHandler`. Plain functions and other callables stay strong. `Automator.__init__` registered `lambda: self.stop(...)`, a closure over `self` that cannot be weakened, so it now registers a `_stop_on_shutdown` bound method instead.

<details>
<summary><b>Measured before/after (post-#427 <code>main</code>, NiceGUI 3.13, Python 3.13)</b></summary>

One actor per page-view, then `Client.prune_instances(client_age_threshold=0)`; the table is a time series *after* the clients are gone, which separates a leak from a collection lag:

`Automator`, 60 views — **before**:

| t after prune (s) | instances alive | shutdown_handlers |
|---|---|---|
| 0 | 60 | 121 |
| 2 | 60 | 121 |
| 10 | **60** | 121 |

**after**:

| t after prune (s) | instances alive | shutdown_handlers |
|---|---|---|
| 0 | 60 | 121 |
| 1 | **0** | 121 |
| 10 | **0** | 121 |

`WheelsSimulation` (registers `rosys.on_shutdown(self.stop)` in `__init__`) went from 5/5 pinned to 0.

**This is what fixes #436's headline 1 GB scenario.** The issue reported that a realistically-sized robot — each carrying a ~5.5 MB local map — crosses 1 GB by ~160 page-views. That is still true on current `main`, because the map hangs off the `WheelsSimulation` that `shutdown_handlers` pins. Same scenario (map attached to the per-page wheels), measured:

| page-views | post-#427 `main` | with this PR |
|---|---|---|
| 40 | 423 MB | 213 MB |
| 80 | 646 MB | 215 MB |
| 120 | 869 MB | 218 MB |
| 160 | **1088 MB** | **220 MB** |
| after prune + 3 s + gc | 1090 MB (never recovers) | 220 MB |

Baseline was 196 MB, so +892 MB → +24 MB. The issue's own figure for this scenario was 1142 MB, so `main` still reproduces it closely.

In the *bare* repro (no local map) RSS is ~215 MB either way — the pinned `WheelsSimulation` objects are individually small, so the object leak only becomes visible once they own real state. That is why this looked minor until the map was attached.
</details>

<details>
<summary><b>Tests — seen to fail without the fix</b></summary>

Six tests added to `tests/test_repeater.py` (the file #427 established for handler-lifetime semantics). Each was checked against the specific line it covers, not just against reverting the whole PR:

| test | proof it bites |
|---|---|
| `..._does_not_keep_a_discarded_module_alive` | `assert <WheelsSimulation object ...>` — still alive without the fix |
| `..._does_not_keep_a_discarded_automator_alive` | `assert <Automator object ...>` |
| `..._is_still_awaited_while_its_object_lives` | an **async** bound method, so it also covers awaiting through the weak wrapper |
| `..._do_not_accumulate_for_discarded_modules` | revert only the pruning → `assert 5 == 0` |
| `..._registered_from_a_temporary_leaves_no_dead_entry` | revert only the append-before-finalize ordering → `assert 1 == 0` |
| `..._dying_during_shutdown_does_not_skip_another_handler` | revert only the `list()` snapshot → `assert ['victim', 'dropper'] == ['victim', 'dropper', 'last']` |

Gates: `pytest` 343 passed / 7 skipped (skips are the `arp-scan` and Linux-only camera tests), `mypy` clean on 217 files, `pylint` 10.00/10, `pre-commit` clean, and all 28 example smoke-runs Ok.
</details>

<details>
<summary><b>Two design questions for you — I picked the conservative option in both</b></summary>

1. **Non-weak-referenceable objects.** `on_repeat` raises `TypeError` for these. Raising in `on_shutdown` would be a breaking change for any slotted module, so this PR falls back to holding the handler strongly and logs a warning naming the object. Say the word if you'd rather it mirror `on_repeat` and raise.
2. **Lambdas and closures.** `on_repeat` rejects them outright. `on_shutdown` accepting them is common in user code, so this PR leaves them strong (and therefore still pinning) rather than rejecting them. Only RoSys' own `Automator` needed rewriting.

Likewise I left the README alone — #427 documented the `on_repeat` lifetime rule there, so say the word and I'll add the matching sentence for `on_shutdown`.

The sharpest behavior change is `PathPlanner`: it registers `on_startup(self.startup)` (which spawns a subprocess) plus a now-weak `on_shutdown(self.shutdown)`. An app that never stores its planner would previously have been kept alive by the registry and would now be collected, skipping that subprocess cleanup. Every example in the repo stores it, so this is not a regression here — but it is the one case where the weak rule changes a real outcome, and it applies equally to `on_repeat` after #427.
</details>

<details>
<summary><b>Second-lineage review</b></summary>

Reviewed adversarially by Codex (different model lineage) across both commits. It confirmed two things I was unsure about: `invoke()` calls the handler and *then* awaits the returned `Awaitable`, so it does not depend on `inspect.iscoroutinefunction` and `_WeakHandler` needs no coroutine metadata forwarding; and `_WeakHandler.__qualname__` keeps the shutdown logging intact. It also confirmed the finalizer does not pin the module, and that `startup()` has no equivalent skip bug (no `await` in its loop).

It found one real defect in the pruning commit, which is fixed above: registering the finalizer *before* appending meant `handler = weak_handler` could drop the last reference to a handler passed straight from a temporary — the object then died mid-registration, the finalizer ran before the entry existed, and the dead entry was appended anyway. Its earlier criticisms of the first commit (a silent `TypeError` fallback that contradicted the docstring; no test proving a *live* weak handler still runs) are addressed too. One open NIT it raised, not acted on: `_forget_shutdown_handler` uses `in`/`remove`, so a pathological callable with a custom `__eq__` could in principle match a weak handler and remove the wrong entry — I kept the existing `_forget_task` idiom for consistency, but an identity scan is a two-line change if you prefer it. It also flagged the `PathPlanner` case noted above, and judged `Persistable` safe (a dropped persistable can only skip its final backup if it is genuinely unreachable).
</details>

_Disclaimer: measurements, implementation and this description by Claude (Opus 5)._

> [!NOTE]
> **Maintainer update (2026-08-04):** Follow-up commits on this branch reorganized the tests into `tests/test_shutdown.py` and `tests/test_path_planning.py`, so the test names in the table above are outdated (coverage is unchanged or stronger).
> Both design questions were settled in favor of the conservative options, now documented in the `on_shutdown` docstring and the README.
> The `PathPlanner` caveat turned out to be benign: the planner process ends on EOF once a collected planner's pipe closes — covered by `test_a_discarded_path_planner_stops_its_process`.


falkoschindler (dismissed):
Thanks a lot for this contribution, @evnchn, — a minimal, well-measured fix with tests that were each seen to fail without it. Great work!

We pushed a few follow-up commits onto the branch:

- Reworded the registration NOTE, documented the strong-fallback for non-weak-referenceable objects in the `on_shutdown` docstring, and added the `on_shutdown` lifetime rule to the README — settling both design questions in favor of your conservative choices.
- Moved the new tests into their own `tests/test_shutdown.py` (leaving `test_repeater.py` untouched relative to `main`) and simplified them: parametrized the module-collection test with the registry-pruning assertion folded in, reduced the mid-shutdown-removal test to plain functions, rewrote the weak-await test against a real `WheelsSimulation`, and extracted a `rosys_lifecycle` fixture. Coverage is unchanged or stronger; the test names in the description's table are therefore stale.
- Added `test_a_discarded_path_planner_stops_its_process`: the `PathPlanner` concern turned out to be benign, because the planner process ends on EOF once the collected planner's pipe closes.

falkoschindler (dismissed):

falkoschindler (approved):

PR #445, opened 2026-07-30
Sample a topic's value on the event loop, and only encode it on the writer

felixwrz:
### Motivation

A converter's `extract` is part of `encode`, and `McapRecorder.log_message` runs `encode` off the event loop on the background writer. Anything that reads live state through it — a hardware module's attribute, a camera's calibration — is therefore read at *flush* time, not at the instant the topic was written. Every message of one flush batch gets the same value.

A live signal then comes out of the file as a staircase whose step width is the writer's flush period, while the timestamps stay evenly spaced. Nothing about such a recording looks wrong, which is what makes it expensive.

On a robot, the per-track speeds and driver tick counters changed at about **1.4 Hz** in the file, where they should have changed at the **core message rate**, around 22 Hz — one sample per core message is all such a topic can ever be. Roughly fifteen out of sixteen samples therefore carried a value that did not belong to their timestamp — and since the timestamps themselves are evenly spaced, the file gives no hint of it. That cost a day of looking for the fault everywhere but here.

### Implementation

`Converter` carries a `sample` next to `encode`, and the two deliberately run in different places:

- `sample` reduces the payload to the value to be recorded, **on the event loop**, when the topic is written. Anything reading live state belongs here.
- `encode` turns that value into bytes, **on the background writer** — which is where the expensive JSON and JPEG work should stay, and the reason the split is worth having at all rather than simply encoding early.

Every factory that took an `extract` now passes it as `sample`, so **no call site changes**: `scalar`, `pose_in_frame`, `frame_transform`, `transform_3d`, and by extension `add_pose_topic`. `camera_calibration` gains one too, since it read `camera.calibration` inside its build. `custom_message` takes `sample=` so app-specific converters can do the same.

The sharp edge is what `sample` must return, and it is documented on the class: **an immutable snapshot**. Handing back a live mutable object merely moves the reference onto the queue, the writer still encodes whatever that object holds later, and the staircase survives the fix. The same caveat applies to a payload with no `sample` at all — `Odometer.FRAME_UPDATED` re-emits one `Frame3d` that is mutated in place.

Three ordering details in `_TopicWriter.write`, each of which is silent when wrong:

1. the `None` guard and auto-dispatch come **first**, so a custom `timestamp` that reads a field off the payload never sees a `None` that would make it raise;
2. the log time is then taken **from the payload**, because the sampled value no longer carries `time` and would silently fall back to the current time, shifting the whole trace;
3. auto-dispatch stays on the **payload** type, as before, so a topic's schema does not depend on whether it samples.

### Progress

- [x] Pytests have been added (or are not necessary). `test_extract_reads_live_state_at_write_time` pins the guarantee end to end: it emits with the live value at 1, moves it to 2 before the writer runs, and expects 1 in the file. Verified to fail without the change. `test_none_payload_is_skipped_before_the_timestamp_is_read` pins ordering detail 1. Two existing `camera_calibration` tests were updated, since that converter's `encode` now receives the sampled calibration rather than reading the camera itself. 91 pass in `test_mcap_converters.py`, `test_mcap_recorder.py` and `test_recorder_replay.py`.
- [x] Documentation has been added (or is not necessary). The `Converter` class documents where each callable runs, what happens when they are swapped, and the immutability requirement; the module docstring points at it.

### DIFF
--- a/rosys/analysis/recording/converters.py
+++ b/rosys/analysis/recording/converters.py
@@ -1,15 +1,18 @@
 """Registry, topic sources and JSON machinery that feed the MCAP recorder.
 
-A :class:`Converter` bundles a :class:`TopicSchema` with an ``encode`` returning
-JSON bytes (or ``None`` to skip). The registry maps RoSys types to converter
-factories so :func:`add_event_topic` / :func:`add_timer_topic` pick the right
-converter from the value's runtime type, with explicit overrides. App-specific
-topics without a foxglove schema use :func:`custom_message` with a JSON schema.
+A :class:`Converter` bundles a :class:`TopicSchema` with an optional ``sample``
+reducing the payload to the value to record, and an ``encode`` returning JSON
+bytes (or ``None`` to skip). The registry maps RoSys types to converter factories
+so :func:`add_event_topic` / :func:`add_timer_topic` pick the right converter from
+the value's runtime type, with explicit overrides. App-specific topics without a
+foxglove schema use :func:`custom_message` with a JSON schema.
 
 The concrete converters for RoSys/Foxglove data types live in :mod:`.foxglove`;
 this module only holds the machinery that wires values into the recorder. Values
-are enqueued unencoded and :func:`Converter.encode` runs on the recorder's
-background writer, so no JSON/JPEG encoding happens on the event loop.
+are sampled and timestamped on the event loop but enqueued unencoded:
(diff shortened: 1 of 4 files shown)


pascalzauberzeug (approved):
Thanks for that fix @felixwrz !
I did three small cleanups that Claude found. Feel free to merge

PR #446, opened 2026-07-31
Make the JPEG quality configurable for image encoding and recording

pascalzauberzeug:
### Motivation

Camera topics dominate an MCAP recording's size, but the JPEG quality was hardcoded at 90 with no way to lower it.

### Implementation

- Add `quality` to `Image.to_jpeg_bytes()` and to the `compressed_image()` converter, reachable as `add_event_topic(recorder, '/camera/front/image', event=camera.NEW_IMAGE, quality=75)`
- Rename `encode_image_as_jpeg`'s `compression_level` to `quality` (as in TurboJPEG and PIL) and extract the shared default into `DEFAULT_JPEG_QUALITY`
- Reject qualities outside 1 to 100 with `validate_jpeg_quality`
- Fix the image route's fast path, which ignored the requested `compression`, and default both image routes to `DEFAULT_JPEG_QUALITY`
- Let `CameraObjects` request its projected images at quality 60 via the new `image_quality` argument (`image_shrink_factor` became an argument too)

⚠️ Fixing the fast path changes what the image routes serve by default:

| Request | Before | Now |
| --- | --- | --- |
| no parameters | 90 | 90 |
| `?compression=60`, unshrunk | 90, the request was ignored | 60, as requested |
| `?shrink=…`, `/undistorted`, `?crop_*`, `?rotation=…` without `compression` | 60 | 90, about 2.2x the bytes |
| any of those *with* `compression` | as requested | as requested |

Views that shrink, undistort, crop or rotate therefore get sharper and heavier unless they pass `?compression=60`, which is what `CameraObjects` now does. Recordings, placeholders and every other `to_jpeg_bytes()` caller stay at 90.

<details>
<summary>Details</summary>

**Why validate.** TurboJPEG answers an out-of-range quality with an unspecific `tjCompress2(): Invalid argument`, and the failure surfaces wherever the encode happens: a topic configured with a bad quality raised once per frame, which the recorder logs once and then drops silently for the rest of the recording. The converter now fails while the topic is set up, and the route answers 400 instead of 500.

**Why 90 stays the route default.** The fast path skipped `_process` only for `compression == 60` and then encoded at 90, so the documented default was the one value that was silently ignored; any other value took the processing path and was honored. No caller passed `compression`, so unshrunk views - calibration dialogs, circle sights, field creator - were served at 90 all along, and honoring the documented 60 would have made exactly those more lossy. The default follows the encoders instead, and callers that want fewer bytes ask for them: quality 60 is about 2.2 times smaller than 90 on a real 1920x1080 camera frame (108 kB vs 246 kB), which is why the 3D scene's textures - one per calibrated camera, up to ten times per second - request 60. The query parameter keeps its name, so URLs are unchanged.

**Tests.** The quality parameter, its validation, and a `quality` override travelling through `add_event_topic` into the encoded JPEG. The knob is documented in the MCAP example README.

</details>

### Progress

---
⬛ claude-opus-5[1m]


NiklasNeugebauer:
i'd like to keep 90 as the default

NiklasNeugebauer:
If you want to keep quality a little higher, another way would be to remember the jpeg bytes handed to `Image.from_jpeg_bytes` and lower the camera's quality directly. But might be more code complexity to do that. Right now you are doing two jpeg encodings.

pascalzauberzeug:
@NiklasNeugebauer I changed the default to 90 and adjusted it for `CameraObjects`.
This means that the default for shrink/undistorted/crop/rotation calls is now also 90 and not 60, that's why it's still marked as a breaking change.

Have a look at the table above.

rosys/vision/image_processing.py
```diff
@@ -19,20 +19,40 @@
...
+    """Ensure a JPEG quality is within the range both encoders accept.
+
+    TurboJPEG only reports an unspecific ``Invalid argument`` for out-of-range values,
+    and the failure would surface deep inside an encode - possibly per frame.
+
+    :param quality: the JPEG quality to check.
+    :raises ValueError: if the quality is not between 1 and 100.
+    """
+    if not 1 <= quality <= 100:
+        raise ValueError(f'JPEG quality must be between 1 and 100, not {quality}')
```
NiklasNeugebauer:
```suggestion
        raise ValueError(f'JPEG quality must be between 1 and 100. Got "{quality}"')
```

NiklasNeugebauer (approved):

PR #447, opened 2026-08-06
Verify the startup checksum when configuring Lizard

evnchn:
**TL;DR** — `configure()` verifies the deployed startup script and retries on mismatch, instead of assuming the write worked. **Reworked to need no firmware change at all**: it queries the `core.startup_checksum()` command that Lizard has always had. Not verified on hardware — see Progress.

### Motivation

Reported in zauberzeug/lizard#256 by Jens Ogorek with measurements: deploying the startup script is fire-and-forget. `configure()` sends `!-` to clear, one `!+<line>` per line, then `!.` to persist — and **none of these produce a response**.

At 115,200 under 74.6 % bus load he measured 0.035 % line loss (full campaign: zauberzeug/lizard#260). When the `!-` was the line that got dropped, the following `!+` lines appended to the **old** script. Invisible until the next restart, at which point:

- the script exists twice in NVS (a readback showed 93 lines instead of 47),
- the second `rdyp = Output(15)` throws at boot, so the rest of the script — including `core.output(...)` — never runs,
- the core comes up **silent**, which looks exactly like dead hardware. It still answers direct commands; that was the only clue.

This happened twice in one day on a real robot. Nothing in RoSys noticed, because nothing was looking.

### Implementation

After `!.`, ask the core what it actually stored and compare against the locally computed checksum. On mismatch, redo the whole rewrite, up to `MAX_CONFIGURE_ATTEMPTS` (3). If it never matches, notify and **skip the restart** — booting a script known to be corrupt is the failure this is trying to prevent.

**No firmware change is required.** `core.startup_checksum()` and its `checksum: <4 hex>` reply already exist and are already what `LizardFirmware.read_core_checksum()` uses to detect an outdated script — the command was simply never consulted at write time. So this works against **every Lizard already deployed in the field**, with no coordinated release.

<details>
<summary><b>Why not <code>read_core_checksum()</code>, and why not a firmware acknowledgement</b></summary>

**Not `read_core_checksum()`** — it early-returns when `robot_brain.is_ready` is `False`:

```python
async def read_core_checksum(self) -> None:
    self.core_checksum = None
    if not self.robot_brain.is_ready:
        self.log.error('Could not read startup checksum from Core. Robot Brain is not ready.')
        return
```

`configure()` deliberately runs every send with `force=True` precisely because the brain may not be ready, so reusing that method would bail on every attempt and fail configuration outright. The inline `send_and_await(..., force=True)` is the same one-line idiom the method itself uses.

**Not a firmware acknowledgement.** The first version of this PR added a new reply to `!.` in Lizard and depended on it. That is strictly more expensive — a wire-protocol change, a coordinated release, and no benefit for any controller already in the field — for the sole gain of one round trip instead of two. The issue itself points at the cheaper route, calling a client-side readback-and-retry loop "a workable interim client-side fix (used successfully throughout the measurement campaign)". The firmware side has been dropped.
</details>

<details>
<summary><b>Test evidence — 3 of 3 new tests fail with the fix reverted</b></summary>

Reverting **only** the `configure()` body to its pre-PR form (keeping `MAX_CONFIGURE_ATTEMPTS`, since the tests import it):

```
FAILED tests/test_robot_brain.py::test_configure_retries_on_checksum_mismatch
FAILED tests/test_robot_brain.py::test_configure_fails_after_repeated_mismatches
FAILED tests/test_robot_brain.py::test_configure_fails_when_checksum_is_unavailable
3 failed, 15 passed
```

With the fix in place: `18 passed`. A green-only test proves nothing, so each new test was watched failing first.

The three cover: mismatch → retry then succeed; permanent mismatch → give up, no `core.restart()`, failure notification; core never answers the query → same give-up path rather than a hang.

CI-equivalent gates, all green:

```
pre-commit  ruff / autopep8 / trailing-whitespace / end-of-file / double-quote-fixer / codespell   Passed
mypy ./rosys      Success: no issues found in 217 source files
pylint ./rosys    10.00/10
pytest            18 passed
```
</details>

<details>
<summary><b>Behaviour change worth a maintainer's attention</b></summary>

`configure()` could previously only succeed; it now has a failure path that **skips `await self.restart()`**. A robot whose script genuinely cannot be written will stay on the old script and say so, rather than restarting into a corrupt one. I believe that is the right call — the whole bug is a silent boot into a broken script — but it is a real change in observable behaviour, not just added checking.

Two smaller notes:

- **Old firmware is unaffected.** Every Lizard version RoSys supports answers `core.startup_checksum()`; `LizardFirmware` already depends on it. An unanswered query therefore means a genuinely broken link, and failing loudly is correct. (The previous ack-based version had to treat "no response" as *success* for backwards compatibility — a compromise that is simply gone now.)
- **Pre-existing, not introduced here:** `read_local_checksum()` hashes `robot_brain.lizard_code`, while `configure()` transmits `splitlines()`-normalised content. These agree only because `RobotHardware.generate_lizard_code()` always newline-terminates. Anyone assigning `lizard_code` directly without a trailing newline is already broken today via `checksums_match`; this PR promotes their symptom from a boot-time warning to a hard configure failure. Computing the local checksum over what is actually sent would close it in one line — happy to add if wanted.
</details>

<details>
<summary><b>Adversarial review (Codex, independent lineage) — one blocker, documented rather than fixed</b></summary>

Verbatim:

```
**BLOCKER** rosys/hardware/robot_brain.py:193: `send_and_await` can still be satisfied by a
stale `checksum:` line. It registers `waiting_list['checksum:']` before sending, and matching
is only by first word. A delayed response from an earlier `_check_lizard_code()` /
`read_core_checksum()` can arrive after the register and before or during this request, match
the local checksum, and let `configure()` restart after a corrupt upload. The new tests only
model ordered fresh responses produced by the just-sent request.

**SUGGESTION** robot_brain.py:194: Parse the checksum response more strictly.
`response.split()[-1] == local_checksum` accepts malformed lines like `checksum: bad 1234`.

**SUGGESTION** tests: add the adversarial stale-line case. [...] This will likely fail today,
which is the point.

The `for`/`else` itself is correct: `break` only on verified match, otherwise the loud failure
path runs after 3 attempts. The behavior change of refusing to restart on checksum failure
matches the stated fix, though it is operationally significant [...]
```

**The mechanism is real and I am not fixing it here.** `send_and_await` sets `waiting_list[ack] = None` *before* sending (robot_brain.py:287) and `read_lines()` fills it on a first-word match (`:225`), so a late reply can satisfy a later await. But:

- **It is pre-existing, not introduced by this PR.** `LizardFirmware.read_core_checksum()` (`lizard_firmware.py:135`) uses the identical pattern with `timeout=1` inside a 5 s retry loop — that is the code most likely to *leave* late `checksum:` replies in flight, and `checksums_match` is exposed to the same race on `main` today.
- **A false pass additionally requires the stale value to equal the intended checksum.** `configure()` is normally reached precisely because `checksums_match is False` (the developer-UI button is only visible then), so the in-flight stale value is the *mismatching* one and the comparison still fails.
- **Closing it properly requires request IDs in the wire protocol** — a change to the messaging layer affecting every `send_and_await` caller, not something to smuggle into a bugfix.

Net effect: this PR strictly reduces silent corruption; it does not claim to make verification airtight. Happy to follow up with a correlated-request change if that is wanted.

The strict-parse suggestion I also left alone: `response.split()[-1]` is the existing house idiom at `lizard_firmware.py:136`, the firmware emits a fixed `checksum: %04x`, and the stricter form would not address the blocker above. Codex's third suggestion is to add a test that fails today — noted, but a knowingly-red test is not shippable.
</details>

### Progress

- [x] Pytests have been added (or are not necessary). Three, all verified to fail with the fix reverted.
- [x] Documentation has been added (or is not necessary). Internal behaviour of `configure()`; the user-visible change is the failure notification.

**Not tested on a real Robot Brain.** The lossy-link scenario that motivates this has only been exercised in the simulation harness — the rig that produced the 0.035 % measurement is @JensOgorek's. If a dropped `!+` line can be reproduced there, the confirmation this needs is watching RoSys retry rather than report success.

### DIFF
--- a/rosys/hardware/lizard_firmware.py
+++ b/rosys/hardware/lizard_firmware.py
@@ -121,7 +121,9 @@ class LizardFirmware:
         self.log.error('Could not read Lizard version from P0')
 
     def read_local_checksum(self) -> None:
-        checksum = sum(self.robot_brain.lizard_code.encode()) % 0x10000
+        # NOTE: the Core stores one newline-terminated line per '!+' command, which is what configure() sends
+        startup = ''.join(f'{line}\n' for line in self.robot_brain.lizard_code.splitlines())
+        checksum = sum(startup.encode()) % 0x10000
         self.local_checksum = f'{checksum:04x}'
         self.log.info('local checksum: %s', self.local_checksum)
 
@@ -132,8 +134,8 @@ class LizardFirmware:
             return
         deadline = rosys.time() + 5.0
         while rosys.time() < deadline and self.robot_brain.is_ready:
-            if response := await self.robot_brain.send_and_await('core.startup_checksum()', 'checksum:', timeout=1):
-                self.core_checksum = response.split()[-1]
+            if checksum := await self.robot_brain.read_startup_checksum(timeout=1):
+                self.core_checksum = checksum
                 self.log.info('core checksum: %s', self.core_checksum)
                 return
         self.log.error('Could not read startup checksum from Core')
--- a/rosys/hardware/robot_brain.py
+++ b/rosys/hardware/robot_brain.py
@@ -216,18 +216,21 @@ class RobotBrain:
         rosys.notify('Lizard configured successfully.', 'positive')
         return True
 
-    async def read_startup_checksum(self, *, timeout: float = float('inf'), force: bool = False) -> str | None:
+    async def read_startup_checksum(self, *, timeout: float = 3.0, force: bool = False) -> str | None:
         """Read the checksum of the startup script that the microcontroller currently holds.
 
         Requests are serialized, because responses are matched by their ``checksum:`` prefix and would
         otherwise be delivered to whichever concurrent request happens to wake up first.
+        The timeout must be finite, because an unanswered request would hold the lock
+        and stall all future requests, including the ones sent by ``configure()``.
 
-        :param timeout: response timeout
+        :param timeout: Response timeout
         :param force: Whether to send the message even if the ESP is not ready
-        :raises EspNotReadyException: When the ESP is not ready and force is ``False``
-        :return: the checksum or ``None`` if the timeout is reached
+        :return: The checksum, or ``None`` if the timeout is reached or the ESP is not ready
         """
         async with self._checksum_lock:
+            if not self.is_ready and not force:
+                return None  # NOTE: the ESP may have gone away while we were waiting for the lock
             response = await self.send_and_await('core.startup_checksum()', 'checksum:', timeout=timeout, force=force)
         return response.split()[-1] if response else None
 
--- a/rosys/hardware/robot_brain.py
+++ b/rosys/hardware/robot_brain.py
@@ -207,6 +207,9 @@ class RobotBrain:
                 f'checksum mismatch ({checksum} instead of {self.lizard_firmware.local_checksum})'
         else:
             rosys.notify(f'Configuring Lizard failed: {reason}.', 'negative', log_level=logging.ERROR)
+            # NOTE: restart to reload the persisted script into RAM; otherwise core.startup_checksum() would keep
+            # reporting the unpersisted upload and the startup check could show a false "checksums match"
+            await self.restart()
             return False
         await self.send('!.', force=True)
         await self.restart()
--- a/rosys/hardware/robot_brain.py
+++ b/rosys/hardware/robot_brain.py
@@ -191,6 +191,7 @@ class RobotBrain:
         """
         rosys.notify('Configuring Lizard...')
         self.lizard_firmware.read_local_checksum()
+        await self.read_startup_checksum(timeout=1.0, force=True)  # drain any stale checksum: response
         for _ in range(MAX_CONFIGURE_ATTEMPTS):
             await self.send('!-', force=True)
             for line in self.lizard_code.splitlines():
--- a/rosys/hardware/robot_brain.py
+++ b/rosys/hardware/robot_brain.py
@@ -70,6 +70,7 @@ class RobotBrain:
         self.esp_pins_p0 = EspPins(name='p0', robot_brain=self)
 
         self._esp_lock = asyncio.Lock()
+        self._checksum_lock = asyncio.Lock()
 
     @property
     def clock_offset(self) -> float | None:
@@ -96,13 +97,15 @@ class RobotBrain:
             await self.lizard_firmware.download()
             await self.lizard_firmware.flash_core()
             await self.restart()
-            await self.configure()
+            if not await self.configure():
+                return
             await self.lizard_firmware.flash_p0()
 
         async def local_update() -> None:
             await self.lizard_firmware.flash_core()
             await self.restart()
-            await self.configure()
+            if not await self.configure():
+                return
             await self.lizard_firmware.flash_p0()
 
         with ui.row().classes('items-center'):
@@ -182,22 +185,48 @@ class RobotBrain:
             return
         await self.send('core.keep_alive()')
 
-    async def configure(self) -> None:
+    async def configure(self) -> bool:
+        """Write the Lizard startup script to the microcontroller and restart it.
+
+        The transferred script is verified before it is persisted, so a lossy connection
+        cannot leave a corrupt script in the microcontroller's storage.
+
+        :return: whether the script was transferred, verified and persisted
+        """
         rosys.notify('Configuring Lizard...')
         self.lizard_firmware.read_local_checksum()
+        reason = 'no checksum received'
         for _ in range(MAX_CONFIGURE_ATTEMPTS):
             await self.send('!-', force=True)
             for line in self.lizard_code.splitlines():
                 await self.send(f'!+{line}', force=True)
-            await self.send('!.', force=True)
-            response = await self.send_and_await('core.startup_checksum()', 'checksum:', timeout=3.0, force=True)
-            if response is not None and response.split()[-1] == self.lizard_firmware.local_checksum:
+            checksum = await self.read_startup_checksum(timeout=3.0, force=True)
+            if checksum == self.lizard_firmware.local_checksum:
                 break
+            reason = 'no checksum received' if checksum is None else \
+                f'checksum mismatch ({checksum} instead of {self.lizard_firmware.local_checksum})'
         else:
-            rosys.notify('Configuring Lizard failed: checksum mismatch.', 'negative', log_level=logging.ERROR)
-            return
+            rosys.notify(f'Configuring Lizard failed: {reason}.', 'negative', log_level=logging.ERROR)
+            return False
+        await self.send('!.', force=True)
         await self.restart()
         rosys.notify('Lizard configured successfully.', 'positive')
+        return True
+
+    async def read_startup_checksum(self, *, timeout: float = float('inf'), force: bool = False) -> str | None:
+        """Read the checksum of the startup script that the microcontroller currently holds.
+
+        Requests are serialized, because responses are matched by their ``checksum:`` prefix and would
+        otherwise be delivered to whichever concurrent request happens to wake up first.
+
+        :param timeout: response timeout
+        :param force: Whether to send the message even if the ESP is not ready
+        :raises EspNotReadyException: When the ESP is not ready and force is ``False``
+        :return: the checksum or ``None`` if the timeout is reached
+        """
+        async with self._checksum_lock:
+            response = await self.send_and_await('core.startup_checksum()', 'checksum:', timeout=timeout, force=force)
+        return response.split()[-1] if response else None
 
     async def restart(self) -> None:
         await self.send('core.restart()', force=True)
--- a/rosys/hardware/robot_brain.py
+++ b/rosys/hardware/robot_brain.py
@@ -93,20 +93,16 @@ class RobotBrain:
             if version_select.options:
                 version_select.value = version_select.options[0]
 
-        async def online_update() -> None:
-            await self.lizard_firmware.download()
-            await self.lizard_firmware.flash_core()
-            await self.restart()
-            if not await self.configure():
-                return
-            await self.lizard_firmware.flash_p0()
-
         async def local_update() -> None:
             await self.lizard_firmware.flash_core()
             await self.restart()
             if not await self.configure():
                 return
             await self.lizard_firmware.flash_p0()
+
+        async def online_update() -> None:
+            await self.lizard_firmware.download()
+            await local_update()
 
         with ui.row().classes('items-center'):
             version_select = ui.select([], label='Lizard Version').style('min-width: 140px;') \
@@ -195,7 +191,6 @@ class RobotBrain:
         """
         rosys.notify('Configuring Lizard...')
         self.lizard_firmware.read_local_checksum()
-        reason = 'no checksum received'
         for _ in range(MAX_CONFIGURE_ATTEMPTS):
             await self.send('!-', force=True)
             for line in self.lizard_code.splitlines():
--- a/rosys/hardware/robot_brain.py
+++ b/rosys/hardware/robot_brain.py
@@ -10,6 +10,7 @@ from .esp_pins import EspPins
 from .lizard_firmware import LizardFirmware
 
 CLOCK_OFFSET_HISTORY_LENGTH = 100
+MAX_CONFIGURE_ATTEMPTS = 3
 
 
 class RobotBrain:
@@ -183,10 +184,18 @@ class RobotBrain:
 
     async def configure(self) -> None:
         rosys.notify('Configuring Lizard...')
-        await self.send('!-', force=True)
-        for line in self.lizard_code.splitlines():
-            await self.send(f'!+{line}', force=True)
-        await self.send('!.', force=True)
+        self.lizard_firmware.read_local_checksum()
+        for _ in range(MAX_CONFIGURE_ATTEMPTS):
+            await self.send('!-', force=True)
+            for line in self.lizard_code.splitlines():
+                await self.send(f'!+{line}', force=True)
+            await self.send('!.', force=True)
+            response = await self.send_and_await('core.startup_checksum()', 'checksum:', timeout=3.0, force=True)
+            if response is not None and response.split()[-1] == self.lizard_firmware.local_checksum:
+                break
+        else:
+            rosys.notify('Configuring Lizard failed: checksum mismatch.', 'negative', log_level=logging.ERROR)
+            return
         await self.restart()
         rosys.notify('Lizard configured successfully.', 'positive')
 
--- a/tests/test_robot_brain.py
+++ b/tests/test_robot_brain.py
@@ -41,7 +41,8 @@ class CommunicationSimulation(Communication):
 
 
 def matching_checksum(robot_brain: RobotBrain) -> str:
-    return f'{sum(robot_brain.lizard_code.encode()) % 0x10000:04x}'
+    startup = ''.join(f'{line}\n' for line in robot_brain.lizard_code.splitlines())
+    return f'{sum(startup.encode()) % 0x10000:04x}'
 
 
 async def connect(communication: CommunicationSimulation) -> None:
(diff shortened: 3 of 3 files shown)


falkoschindler (changes requested):
Thanks for tackling this — the failure mode is real and `core.startup_checksum()` is the right primitive to use against it. Reading the change against the Lizard firmware, though, the verification currently happens one step too late and checks a different thing than the description suggests, so I'd like a few changes before we merge.

### 1. `!.` persists to NVS before the checksum is verified

`!.` maps to `Storage::save_startup()` → `Storage::put(Storage::startup)` (`main/main.cpp`, `main/storage.cpp:145-148`), which writes the buffer to NVS unconditionally. The loop sends `!.` first and asks for the checksum afterwards, so after three failed attempts the corrupt script is already permanently stored. Skipping `restart()` then protects nothing: the next power cycle boots exactly the doubled-script / silent-Core failure from zauberzeug/lizard#256 — and RoSys can no longer notice it, because a silent Core means no core messages, hence no `ESP_CONNECTED` and no `_check_lizard_code`.

### 2. The checksum covers Lizard's RAM buffer, not NVS

`core.startup_checksum()` sums over `Storage::startup` (`main/modules/core.cpp:78-83`), which `!+` appends to and which `!.` merely copies into NVS. A dropped `!.` therefore passes verification: `configure()` reports success and restarts into the _old_ script. That is the opposite of "ask the core what it actually stored".

**Suggested fix for both:** move the verification between the `!+` lines and `!.`. The RAM scope then becomes exactly the right scope — it validates the transfer before anything is persisted — and `!.` only ever runs on a buffer that is known to be intact. The NVS side is covered for free right after: `Storage::init()` loads `Storage::startup` back from NVS at boot (`main/storage.cpp:16-18`), so the post-restart `_check_lizard_code` on `ESP_CONNECTED` is already a genuine NVS check.

### 3. The failure is not propagated, and `flash_p0()` restarts anyway

`online_update()` / `local_update()` ignore the outcome and continue to `flash_p0()` (`rosys/hardware/robot_brain.py:95-106`), which ends in `await self.robot_brain.restart()` (`rosys/hardware/lizard_firmware.py:202`). In precisely the flow this is meant to protect, the deliberate "skip the restart" is undone a moment later. `configure()` should signal failure — return a `bool` or raise — and the update flows should stop there.

### 4. The `checksum:` slot is shared, so a good upload can hard-fail

`waiting_list` holds one entry per ack prefix (`rosys/hardware/robot_brain.py:59` and `287-292`). `_check_lizard_code` runs as a background task on `ESP_CONNECTED` and polls `core.startup_checksum()` for up to five seconds (`rosys/hardware/lizard_firmware.py:128-139`) — which is exactly the window opened by the `restart()` immediately before `configure()` in both update flows. Whichever coroutine wakes up first pops the reply; the other one blocks until its timeout and sees `None`. Before this PR that produced a spurious "outdated" warning; now it means three retries and a hard failure on a perfectly good upload, leaving the robot on the pre-flash configuration. I think it should be addressed here, since this PR is what turns it from cosmetic into consequential — either by sequencing the two checks or by keying the wait per request instead of per prefix.

### 5. "checksum mismatch" is reported even when nothing answered

`rosys/hardware/robot_brain.py:197` uses the same message for both branches, so a dead link or a Core that never came up sends the operator hunting for a corrupted script. Distinguishing "no reply" from "mismatch" would make the notification actionable.

### 6. The trailing-newline normalisation is now fatal

The local checksum hashes `lizard_code` verbatim, while the wire sends `splitlines()` and Lizard re-appends `'\n'` per line (`main/storage.cpp:121-123`). `generate_lizard_code()` always ends with a newline, so the framework path is fine, but a hand-assembled `lizard_code` without one now fails `configure()` three times instead of merely warning. The one-line fix you offered in the PR description is worth taking.

evnchn:
**All six addressed**, rebased onto your `a61affe0` so the parametrized test survives. The verification now sits between the `!+` lines and `!.`, so nothing corrupt reaches NVS; `configure()` returns a `bool` that both update flows honour; the `checksum:` requests are serialized behind a lock; the failure reason distinguishes silence from mismatch; and the local checksum is computed over the newline-normalised script that is actually sent.

One thing I want to flag before you merge rather than fix unilaterally: **"the NVS side is covered for free" holds in general but not in the repair case** — details in the last fold. It is the only open question I have.

### What changed

| # | Your point | Change |
|---|---|---|
| 1 | `!.` persists before verification | `!.` moved after the loop; it only ever runs on a buffer whose checksum matched |
| 2 | Checksum covers RAM, not NVS | Same move — RAM is now exactly the right scope, pre-persist |
| 3 | Failure not propagated; `flash_p0()` restarts anyway | `configure() -> bool`; `online_update()` / `local_update()` return early |
| 4 | Shared `checksum:` slot can hard-fail a good upload | New `RobotBrain.read_startup_checksum()` serializes every request behind `_checksum_lock`; `LizardFirmware.read_core_checksum()` now goes through it too |
| 5 | "mismatch" reported when nothing answered | `no checksum received` vs `checksum mismatch (<got> instead of <want>)` |
| 6 | Trailing-newline normalisation now fatal | `read_local_checksum()` hashes `''.join(f'{line}\n' for line in code.splitlines())` |

On 4 I took the sequencing option rather than per-request keys: it keeps the change inside this PR's scope, and every current caller passes a finite timeout, so `asyncio.Lock`'s FIFO fairness bounds the wait at one poll iteration.

<details>
<summary><b>What your parametrization needed, and why</b></summary>

Your `a61affe0` is kept as the shape — the deque still drives everything and `None` still stands for a missing response. Two assertions had to move, both because of point 1:

```python
-    assert communication.sent.count('!.') == len(checksums)
+    assert communication.sent.count('!-') == len(checksums)
+    assert communication.sent.count('!.') == (0 if reason else 1), 'only a verified script may be persisted'
```

`!.` is no longer one-per-attempt — it is sent exactly once, after the loop, and never at all on failure. `!-` is what now counts the attempts. That assertion is the load-bearing one for points 1 and 2: a corrupt script must never reach the storage at all, not merely avoid a restart.

`expect_success` became `reason`, which collapses your two parameters into one and covers point 5 in the same table:

```python
@pytest.mark.parametrize('checksums, reason', [
    (['ffff', MATCHING], None),
    (['ffff'] * MAX_CONFIGURE_ATTEMPTS, 'checksum mismatch (ffff instead of'),
    ([None] * MAX_CONFIGURE_ATTEMPTS, 'no checksum received'),
], ids=['retry_then_match', 'repeated_mismatch', 'no_response'])
```

`test_startup_checksum_requests_are_serialized` also moved onto your `None`-in-deque idiom, so `answers_checksum` is gone entirely.
</details>

<details>
<summary><b>Test evidence — three targeted reverts, each failing exactly its own assertions</b></summary>

21 tests in `tests/test_robot_brain.py`. Each fix was reverted on its own, on the rebased tree, and the suite re-run:

```
### revert the verification back to after '!.'
FAILED test_configure_verifies_startup_checksum[retry_then_match]
FAILED test_configure_verifies_startup_checksum[repeated_mismatch]
FAILED test_configure_verifies_startup_checksum[no_response]
3 failed, 18 passed

### remove _checksum_lock
FAILED test_startup_checksum_requests_are_serialized        # 2 requests on the wire, expected 1
1 failed, 20 passed

### hash lizard_code verbatim again
FAILED test_configure_without_a_trailing_newline
FAILED test_local_checksum_matches_what_lizard_stores       # got 02c6, expected 02d0
2 failed, 19 passed
```

`test_configure_does_not_deadlock_with_the_startup_check` is honestly named: it does **not** fail pre-fix. The simulation answers every query, so both waiters get satisfied and the slot-steal cannot be staged there. It guards the other direction — that waiting on the new lock does not stall a good upload while `_check_lizard_code` polls. `test_startup_checksum_requests_are_serialized` is the one that actually fails without the lock.

Gates:

```
pre-commit   ruff / autopep8 / trailing-whitespace / end-of-file / double-quote-fixer / codespell / mdformat   Passed
mypy ./rosys      Success: no issues found in 217 source files
pylint ./rosys    10.00/10
pytest            361 passed, 7 skipped
```

The 7 skips are `UsbCamera is only supported on Linux` and six `arp-scan is not installed` — nothing near this code. `tests/test_path_planning.py` is intermittently flaky on macOS with a `TimeoutError: process ...`, a different test each run; it reproduces the same way on unmodified `main` (3 runs, 3 different failing tests), so it is pre-existing and environmental, not from this change.
</details>

<details>
<summary><b>Open question: "the NVS side is covered for free" — true in general, false in the repair case</b></summary>

Your suggested fix relies on the post-restart `_check_lizard_code` as the genuine NVS check, and that is correct whenever the persisted script is *valid*: a dropped `!.` leaves the previous, working script in NVS, the Core boots, `ESP_CONNECTED` fires, and the check warns "startup code is outdated". The operator sees "Lizard configured successfully" immediately followed by that warning, which is a little untidy but self-correcting.

It stops being covered in the one case this PR exists for. Suppose NVS already holds the doubled script from lizard#256 and `configure()` is being run to repair it:

1. the transfer succeeds and verifies — the RAM buffer is correct,
2. `!.` is the line that gets dropped, so NVS still holds the doubled script,
3. `restart()` boots the doubled script, the duplicate `rdyp = Output(15)` throws, the Core comes up silent,
4. no core messages → no `ESP_CONNECTED` → `_check_lizard_code` never runs,
5. and `configure()` has already returned `True` with "Lizard configured successfully".

So the exact failure mode being fixed survives in the repair path, and the success notification is actively misleading there. It is pre-existing rather than introduced here — the current `configure()` reports success unconditionally — but this PR is what makes the claim explicit, so it seems like the right time to decide.

The fix that works even against a silent Core is a post-restart readback: the Core still answers direct commands when its startup script died, and `Storage::init()` has by then reloaded `Storage::startup` from NVS, so one `read_startup_checksum()` after the restart is a true NVS check. Cost is waiting for the Core to come back before `configure()` returns, plus a new way for a good upload to be reported as failed if that readback is unreliable — which is the same class of risk as your point 4, so I did not want to add it without your call.

Three options, happy with any:

- **leave it** — accept the narrow hole, and I will note the limitation in `configure()`'s docstring so `True` is not read as "persisted and verified";
- **post-restart readback** — verify NVS after the restart, ~8 lines, and `True` becomes honest;
- **separate issue** — merge this as prescribed and file the repair-path case on its own.
</details>

<details>
<summary><b>Adversarial review (Codex, independent lineage)</b></summary>

Ran against the working tree with the Lizard sources available. Verbatim on the points that matter:

```
**MUST-FIX**
robot_brain.py:211 `await self.send('!.', force=True)` followed by robot_brain.py:214 `return True`
Why it matters: this still reports success if `!.` is dropped; the RAM buffer was verified, but
persisted NVS was not.

**WORTH-IMPROVING**
tests/test_robot_brain.py:183 -- your read is correct: this test does not prove the pre-fix race;
it mostly guards that the lock path does not deadlock. The real regression test is
test_robot_brain.py:166. Concrete fix: rename the test to make the narrower guarantee explicit.

- `_checksum_lock` does fix the concurrent same-prefix waiter race. It does not just narrow that
  specific issue. With internal callers using finite 1s/3s timeouts and `asyncio.Lock` fairness,
  I do not see starvation or deadlock for `configure()`.
- Your late-reply claim is right. [...] real closure needs request correlation or an acknowledged
  command with unique token.
- The newline normalization is equivalent to what this Python sender produces.
- Leaving dirty RAM on failure is defensible. Next `configure()` starts with `!-` [...] Since RoSys
  does not later send `!.`, I would not block on this.
- `configure() -> bool` is a mild API change but runtime-compatible for callers ignoring the return.
  `bool` is the right shape for the UI/update flow; raising would make button/background-task
  behavior harsher.

Score: 8/10. Highest-leverage fix: verify the post-restart checksum before `configure()` returns True.
Verdict: the core RAM-transfer verification claim holds; the persisted-NVS success claim is still
too strong until `!.` is covered.
```

I took the test rename, and its MUST-FIX is the open question in the fold above — its suggested fix goes beyond what you asked for and touches the timing of `configure()`, so it is yours to decide rather than mine to smuggle in.

Still not tested on a real Robot Brain; the lossy-link scenario has only been exercised in the simulation harness.
</details>

falkoschindler (changes requested):
Thanks for the thorough rework — the core of this PR checks out. I verified the approach against the actual Lizard sources (`core.cpp`, `storage.cpp`, `main.cpp`): the checksum formula and the newline-per-line normalization exactly match what `Storage::append_to_startup` builds and what `core.startup_checksum()` sums, and verifying before `!.` is the right order.

I pushed the smaller findings directly to the branch so we don't have to churn through a list of one-liners — reasoning is in the commit messages:

- **Restart the Core after a failed configure** (bb3d87a6): the failure path left the Core's RAM buffer (which `core.startup_checksum()` sums) holding the unpersisted upload while NVS kept the old script. If only the checksum *responses* were lost, the next startup check would read the RAM checksum and report a false "checksums match" — and the robot would silently boot the old script on the next power cycle. Restarting reloads RAM from NVS, and it also restores the pre-#447 contract that `configure()` always ends in a restart, for callers that ignore the new bool.
- **Harden `read_startup_checksum`** (08a119ca): a finite default timeout (an unanswered request with the old `inf` default would have held `_checksum_lock` forever and starved every future `configure()`), plus a readiness re-check after acquiring the lock (awaiting the lock is a new suspension point between `read_core_checksum`'s `is_ready` guard and `send_and_await`'s ready check, so a not-ready transition during a contended wait raised `EspNotReadyException` uncaught into the `ESP_CONNECTED` handler).
- **Cleanups** (52e8a9cc): dead `reason` initialization, `online_update` = `download()` + `local_update()`, the test helper delegates to `read_local_checksum()` instead of duplicating the formula, dead fallback promotion in the test simulation, `is` → `==` on the `MATCHING` sentinel.

One point I'd like addressed in this PR, and two I'd treat as follow-ups:

**1. Stale `checksum:` responses can verify the wrong attempt (please address here).** `send_and_await` matches responses purely by the `checksum:` prefix, with no request identity, and there is no yield between one attempt's timeout-pop and the next attempt's re-registration (`SerialCommunication.send` never awaits internally). A response that misses the 3 s window — realistically because the event loop stalled — is therefore *deterministically* delivered to the next waiter: attempt N+1 can be "verified" by attempt N's checksum, and `!.` then persists an upload that was never checked. The same works across calls: a timed-out `read_core_checksum` right before `configure()` feeds the *old stored config's* checksum into attempt 1, which matches `local_checksum` exactly in the common "re-configure unchanged code" case. The post-restart `_check_lizard_code` flags it after the fact, but this is precisely the failure mode the PR exists to close. Cheapest fix I see: drain the `checksum:` slot before the retry loop with a throwaway read; nicer would be some request identity in the pairing — your call.

**2. `configure()`'s atomicity is accidental (follow-up).** Concurrent `configure()` calls (double-click, configure button during an update flow) cannot corrupt the persisted script today — but only because `SerialCommunication.send` is fully synchronous inside, so each upload hits the wire as one non-yielding stretch and a partial buffer can never sit under a `!.`. Any future transport that genuinely awaits (or a flow-control await added to serial writes) silently reopens the interleaving. An `asyncio.Lock` around the attempt loop would make the guarantee structural instead of accidental; it would also stop a double-click from burning 3×3 s and reporting a spurious failure.

**3. The `waiting_list` race exists today for `'p0:'` (pre-existing, follow-up issue).** `EspPins('p0').update_pin` and `read_p0_version` register the *identical* ack key `'p0:'`, so any overlap — Refresh vs. "Check P0 Strapping Pins", `update_all`'s 34 sequential waits, `flash_p0`'s strapping-pin read — can hand one waiter's response to the other; a GPIO status line parses into `p0_version = ''` silently. That's the same race class `_checksum_lock` now fixes for `checksum:` only. Moving the per-ack serialization into `send_and_await` itself (e.g. a `dict[str, asyncio.Lock]` keyed by ack) would fix all callers with one mechanism and let `_checksum_lock` disappear again. Not this PR's job — just flagging it while the context is fresh.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%20findings%20and%20wording%20from%20the%20model%3B%20the%20author%20decided%20which%20fixes%20to%20apply%20directly%20and%20which%20to%20request%20here.&a=claude-code "claude-fable-5: Review findings and wording from the model; the author decided which fixes to apply directly and which to request here.")

PR #450, opened 2026-08-11
Fix parallelize not honoring uninterruptible with return_when_first_completed

ftilde:
## Motivation

`parallelize(..., return_when_first_completed=True)` cancelled its siblings the moment the first coroutine finished, even when a sibling was inside an `@uninterruptible` section. That defeats the guarantee `@uninterruptible` exists to give: a tool that has started an atomic step gets to finish it.

## Implementation

- Wait for siblings that are mid-`@uninterruptible` before cancelling, instead of cancelling on first completion
- Expose the uninterruptible state on `Automation` so `parallelize` can see it
- Add `test_first_completed_waits_for_an_uninterruptible_sibling`

---
⬛ claude-opus-5

### DIFF
--- a/rosys/automation/automation.py
+++ b/rosys/automation/automation.py
@@ -20,6 +20,12 @@ class AutomationStopped(BaseException):
     """
 
 
+def in_uninterruptible_section() -> bool:
+    """Whether the currently running automation is inside an ``@uninterruptible`` function."""
+    automation = _CURRENT_AUTOMATION.get()
+    return automation is not None and automation._uninterruptible_depth > 0  # pylint: disable=protected-access
+
+
 def uninterruptible(func: Callable):
     """Decorator to make an async function uninterruptible until it exits.
 
--- a/rosys/automation/parallelize.py
+++ b/rosys/automation/parallelize.py
@@ -1,7 +1,7 @@
 import asyncio
 from collections.abc import Coroutine
 
-from .automation import AutomationStopped
+from .automation import AutomationStopped, in_uninterruptible_section
 
 
 class parallelize:
@@ -12,6 +12,8 @@ class parallelize:
     to run them in parallel.
 
     Note that ``parallelize`` will be uninterruptible if one of its coroutines is marked with ``@rosys.automation.uninterruptible``.
+    This also holds for ``return_when_first_completed``: the remaining coroutines are only stopped
+    once no coroutine is inside an uninterruptible section anymore.
     """
 
     def __init__(self, *coros: Coroutine, return_when_first_completed: bool = False) -> None:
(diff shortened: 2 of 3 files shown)


evnchn:
*Posted by Qwen Code on evnchn's behalf.*

**TL;DR:**

**LGTM — empirically verified.**

The test fails without the fix (`slow done` missing — sibling cancelled mid-`@uninterruptible`) and passes with it (18/18 green). The fix is correct, minimal, and backward compatible.

Three nits, none blocking:

- `in_uninterruptible_section()` touches `_uninterruptible_depth` (private) — fine intra-package, but a `@property` would be cleaner if external callers ever need this
- `stop_when_interruptible` naming is slightly misleading (means "defer stop until all interruptible") — inline comment clarifies, rename optional
- Test covers 1 quick + 1 uninterruptible sibling; multiple uninterruptible siblings and nested depth would strengthen coverage (correct by construction via the shared counter, but explicit tests are cheap insurance)

<details>
<summary><b>Empirical verification — before/after</b></summary>

Ran `tests/test_automations.py` (18 tests) against both branches:

| State | `test_first_completed_waits_for_an_uninterruptible_sibling` | Full suite |
|---|---|---|
| **Without fix** (main) | FAIL — `events == ['quick done']` | 17/18 |
| **With fix** (PR) | PASS — `events == ['quick done', 'slow done']` | 18/18 |

Command: `python -m pytest tests/test_automations.py -v --timeout=30`

</details>

<details>
<summary><b>Correctness analysis — why the shared depth counter works</b></summary>

`in_uninterruptible_section()` reads `_CURRENT_AUTOMATION` (a `ContextVar`) shared by all coroutines under the same `Automation`. Since `@uninterruptible` increments `automation._uninterruptible_depth` — a single counter on the shared object — checking `depth > 0` correctly answers "is ANY sibling currently uninterruptible?"

The `stop_when_interruptible` flag defers cancellation. Each loop iteration re-checks `in_uninterruptible_section()`. The `asyncio.sleep(0)` in `@uninterruptible`'s finally block yields control, allowing `parallelize` to observe `depth == 0` before the coroutine fully completes.

Backward compatible: when no coroutine is `@uninterruptible`, `in_uninterruptible_section()` returns False and the old immediate-cancel path runs unchanged.

</details>

---
*Changelog: 2026-08-13 — reshaped to supercomment format*

[※](https://zauberzeug.github.io/colophon/#m=qwen3.7-plus&t=Code%20review%20and%20empirical%20test%20verification%20of%20rosys%23450%20by%20the%20model%3B%20test%20suite%20run%20and%20before%2Fafter%20comparison%20by%20the%20model%3B%20review%20text%20composed%20by%20the%20model%20with%20operator%20approval. "qwen3.7-plus: Code review and empirical test verification of rosys#450 by the model; test suite run and before/after comparison by the model; review text composed by the model with operator approval.")

PR #451, opened 2026-08-11
Allow forward to stop when automator fails

ftilde:
### Motivation

When an automation raises, `forward()` keeps stepping the simulated clock to the end of the span it was given, and whatever the test reports afterwards — a `TimeoutError` on the condition, or a bare assertion — says nothing about the actual cause.

`forward()` used to guard against this: it read `rosys.get_last_exception()`, which `_watch_emitted_events()` filled in from failed event tasks. That assignment was dropped in #345 when NiceGUI took over event dispatch, so the guard has been dead ever since — `_state.exception` was only read and reset, never set.

### Implementation

- Keep the failure on the object that has it: `Automator.last_exception`, set in `_handle_exception` and cleared when the automator starts its next automation. Useful in its own right — "why did the last run fail" was previously only a fleeting `AUTOMATION_FAILED` argument
- Short-circuit `forward()` on it, before waiting rather than after, raising `AssertionError: automation failed: <exception!r>` chained from the original
- Add `forward(..., fail_on_automation_failure=False)` for tests that provoke a failure on purpose and want to keep waiting; the three such waits in `test_run.py` and `test_automations.py` use it, each still asserting the failure afterwards
- Drop the dead `_state.exception` / `get_last_exception()`
- Clear `helpers.automator`/`driver`/`odometer` on teardown of the `rosys_integration` fixture: `forward()` now reads that module global, and tests like `tests/test_multi_queue_lazy_worker.py` request no fixture at all, so the previous test has to clean up after itself
- Add `test_a_raising_automation_stops_forwarding` (stops at once, keeps the original exception as `__cause__`) and `test_a_new_automation_forwards_past_an_earlier_failure` (the clear-on-start that `test_run.py` silently depends on)

Three limitations, taken knowingly:

- **Narrower than the pre-#345 mechanism**, which caught any event-handler task failure. This covers automations only; an exception in an `on_repeat` or `on_startup` handler still lets `forward()` run its full span.
- **The guard is absent, never wrong, when `helpers.automator` is unset** — a suite whose conftest does not assign it simply keeps today's behaviour.
- **A failed automation's traceback stays reachable until the next `start()`**, which is what "last exception" means, but on a robot it pins that automation's frames and buffers.
-  A failure raised after its automation was superseded is attributed to the current one.

**Breaking for downstream test suites:** `rosys.testing.fixtures` is a pytest plugin, so any downstream test that provokes an automation failure and then forwards will now fail until it opts out. The removal of `get_last_exception()` is not the breaking part — it was never exported from `rosys` and has no users in-tree or in the app.

### Progress

- [x] I chose meaningful labels (if GitHub allows me to so). — needs `breaking change`

---
⬛ claude-opus-5

### DIFF
--- a/rosys/testing/helpers.py
+++ b/rosys/testing/helpers.py
@@ -60,10 +60,13 @@ async def forward(seconds: float | None = None,
     else:
         raise ValueError('invalid arguments')
 
-    log.info('\033[94m%s\033[0m', msg)
-    while not condition():
+    def check_automation_failure() -> None:
         if fail_on_automation_failure and automator is not None and automator.last_exception is not None:
             raise AssertionError(f'automation failed: {automator.last_exception!r}') from automator.last_exception
+
+    log.info('\033[94m%s\033[0m', msg)
+    check_automation_failure()
+    while not condition():
         if rosys.time() > start_time + timeout:
             raise TimeoutError(f'condition took more than {timeout} s')
         if not run.running_cpu_bound_processes:
@@ -71,6 +74,8 @@ async def forward(seconds: float | None = None,
             await asyncio.sleep(0)
         else:
             await asyncio.sleep(0.01)
+        # an automation can only fail while we sleep, and its failure may satisfy the condition in the same step
+        check_automation_failure()
 
 
 def assert_pose(x: float, y: float, *, deg: float | None = None, position_tolerance: float = 0.1, deg_tolerance: float = 1.0) -> None:
--- a/rosys/testing/helpers.py
+++ b/rosys/testing/helpers.py
@@ -60,13 +60,12 @@ async def forward(seconds: float | None = None,
     else:
         raise ValueError('invalid arguments')
 
-    def check_automation_failure() -> None:
+    log.info('\033[94m%s\033[0m', msg)
+    while True:
         if fail_on_automation_failure and automator is not None and automator.last_exception is not None:
             raise AssertionError(f'automation failed: {automator.last_exception!r}') from automator.last_exception
-
-    log.info('\033[94m%s\033[0m', msg)
-    check_automation_failure()
-    while not condition():
+        if condition():
+            break
         if rosys.time() > start_time + timeout:
             raise TimeoutError(f'condition took more than {timeout} s')
         if not run.running_cpu_bound_processes:
@@ -74,8 +73,6 @@ async def forward(seconds: float | None = None,
             await asyncio.sleep(0)
         else:
             await asyncio.sleep(0.01)
-        # an automation can only fail while we sleep, and its failure may satisfy the condition in the same step
-        check_automation_failure()
 
 
 def assert_pose(x: float, y: float, *, deg: float | None = None, position_tolerance: float = 0.1, deg_tolerance: float = 1.0) -> None:
--- a/rosys/testing/helpers.py
+++ b/rosys/testing/helpers.py
@@ -62,9 +62,8 @@ async def forward(seconds: float | None = None,
 
     log.info('\033[94m%s\033[0m', msg)
     while not condition():
-        failure = rosys.get_automation_failure()
-        if fail_on_automation_failure and failure is not None:
-            raise AssertionError(f'automation failed: {failure!r}') from failure
+        if fail_on_automation_failure and automator is not None and automator.last_exception is not None:
+            raise AssertionError(f'automation failed: {automator.last_exception!r}') from automator.last_exception
         if rosys.time() > start_time + timeout:
             raise TimeoutError(f'condition took more than {timeout} s')
         if not run.running_cpu_bound_processes:
--- a/rosys/testing/helpers.py
+++ b/rosys/testing/helpers.py
@@ -26,7 +26,8 @@ async def forward(seconds: float | None = None,
                   y: float | None = None,
                   tolerance: float = 0.1,
                   dt: float = 0.01,
-                  timeout: float = 100):
+                  timeout: float = 100,
+                  fail_on_automation_failure: bool = True):
     start_time = rosys.time()
     if seconds is not None:
         def condition():
@@ -61,6 +62,9 @@ async def forward(seconds: float | None = None,
 
     log.info('\033[94m%s\033[0m', msg)
     while not condition():
+        failure = rosys.get_automation_failure()
+        if fail_on_automation_failure and failure is not None:
+            raise AssertionError(f'automation failed: {failure!r}') from failure
         if rosys.time() > start_time + timeout:
             raise TimeoutError(f'condition took more than {timeout} s')
         if not run.running_cpu_bound_processes:
@@ -68,9 +72,6 @@ async def forward(seconds: float | None = None,
             await asyncio.sleep(0)
         else:
             await asyncio.sleep(0.01)
-        exception = rosys.get_last_exception()
-        if exception is not None:
-            raise RuntimeError(f'error while forwarding time {dt} s') from exception
 
 
 def assert_pose(x: float, y: float, *, deg: float | None = None, position_tolerance: float = 0.1, deg_tolerance: float = 1.0) -> None:
--- a/rosys/automation/automator.py
+++ b/rosys/automation/automator.py
@@ -57,6 +57,8 @@ class Automator:
 
         self.enabled: bool = True
         self.automation: Automation | None = None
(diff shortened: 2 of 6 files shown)


ftilde:
@codingpaula @felixwrz maybe one of you can review?

evnchn:
Dropping the `_state.exception` global is a good move. Two findings.

**1. `automator.py` — a stopped automation's *late* failure lands on the new automation and kills a healthy run.**

`start()` clears `last_exception` and installs the new `Automation` synchronously, but the old one is still being driven through its `finally` cleanup (async cleanup on stop, per PR 403 / `test_async_finally_block_on_stop`). It holds the same `self._handle_exception`, so a raise during that cleanup writes `last_exception` *after* the swap.

<details>
<summary>Repro</summary>

```python
async def failing_cleanup():
    try:
        while True:
            await rosys.sleep(1)
    finally:
        await rosys.sleep(1)
        raise RuntimeError('cleanup of the OLD automation broke')

async def healthy():
    await rosys.sleep(3)

automator.start(failing_cleanup())
await forward(seconds=2)
automator.start(healthy())   # cleared; old cleanup still pending
await forward(seconds=5)
# AssertionError: automation failed: RuntimeError('cleanup of the OLD automation broke')
```

</details>

`test_a_new_automation_forwards_past_an_earlier_failure` guards the same case with the old automation failing *before* the handover; here it fails after, so clearing in `start()` is too early. Binding the failure to its `Automation` covers both — either `Automation` owns `last_exception` and the automator reads through, or `_handle_exception` captures its `Automation` and no-ops once that is no longer `self.automation`.

(Related, pre-existing: the retained handler's `Automator.abort()` acts on whatever is in `self.automation`, so it aborts the new one too.)

**2. `helpers.py` — the guard is sticky, which is most of the downstream churn.**

`last_exception` lives until the next `start()` and is re-read on every later `forward()`, so a test that provokes a failure, asserts it, then forwards for an unrelated reason must repeat `fail_on_automation_failure=False` each time. Raising only when it *changes* during the wait drops that:

```python
exception_at_entry = automator.last_exception if automator is not None else None
while not condition():
    if fail_on_automation_failure and automator is not None \
            and automator.last_exception is not exception_at_entry:
        raise AssertionError(...) from automator.last_exception
```

All four opt-out sites in this PR follow a `start()` that just cleared the field and fail mid-loop, so they are unaffected. It is a real semantic change though: a failure already present at entry and never replaced would no longer raise. `is not` keeps a second, distinct failure detected.

Also removes an asymmetry from the check sitting inside `while not condition():` — after a failure, `forward(until=lambda: automator.is_stopped)` returns silently while `forward(seconds=1)` raises.

rosys/testing/helpers.py
```diff
@@ -61,16 +62,15 @@ def condition():
 
     log.info('\033[94m%s\033[0m', msg)
     while not condition():
+        if fail_on_automation_failure and automator is not None and automator.last_exception is not None:
+            raise AssertionError(f'automation failed: {automator.last_exception!r}') from automator.last_exception
```
evnchn: Committable version of finding 2 above — raise only when `last_exception` *changes* during the wait, so an already-handled failure stops leaking into every later `forward()`.

```suggestion
    log.info('\033[94m%s\033[0m', msg)
    exception_at_entry = automator.last_exception if automator is not None else None
    while not condition():
        if (fail_on_automation_failure and automator is not None
                and automator.last_exception is not exception_at_entry):
            raise AssertionError(f'automation failed: {automator.last_exception!r}') from automator.last_exception
```

Wrapped the condition to stay under the 120-char limit. With this, the four `fail_on_automation_failure=False` sites in the PR still behave identically, and only the unrelated follow-up forwards stop raising.

ftilde: This would introduce a race where an automation starts and breaks immediately (throws an exception) before forward. There is (was) a race in this anyway, however. see below.

ftilde:
2: This is actually desired behavior I would argue. Otherwise we have a race when an automation is started and fails immediately before forward. There was another race with condition vs. the exception check, but I have fixed that in a new commit.

1: This is true, but I would argue that the main defect is: The invariant that automator has at most one automation running at the same time is violated by such a case. This is _already_ an issue though and your example results in the next automation (healthy) being aborted, I think. Worth a fix, but orthogonal to this PR I would argue.

evnchn:
**2 is yours, my suggestion is wrong.** Applied on top of afd7e4d4 it fails your new `test_a_failure_stops_forwarding_to_a_condition_that_already_holds` — the failure is already in `last_exception` before `forward()` is called, so a delta check can never see it. The pre-loop check is the right shape. The churn costs a downstream test one line, `automator.last_exception = None` after it has asserted.

**On 1, one correction: the new automation is not aborted.** `abort()` runs when `self.automation` is already `None`, so it returns early. Old cleanup raises at t=4.0, new automation started at t=2.0 and would finish at t=7.0:

| | main 42992147 | afd7e4d4 |
|---|---|---|
| `abort()` acts on | `None` | `None` |
| new automation | completes at t=7.0 | never gets there |
| `forward(seconds=10)` | runs its full span | raises at t=4.0 |

<details>
<summary>Repro</summary>

```python
async def test_late_failure_of_old_automation(automator: Automator):
    async def failing_cleanup() -> None:
        try:
            while True:
                await rosys.sleep(1)
        finally:
            await rosys.sleep(2)
            raise RuntimeError('cleanup of the OLD automation broke')

    async def healthy() -> None:
        await rosys.sleep(5)
        print(f'NEW completes at t={rosys.time():.1f}')

    automator.start(failing_cleanup())
    await forward(seconds=2)
    automator.start(healthy())
    await forward(seconds=10)
```

main 42992147 prints `NEW completes at t=7.0`; afd7e4d4 raises `AssertionError: automation failed: RuntimeError('cleanup of the OLD automation broke')` at t=4.0.

</details>

**Your invariant is real and older than this PR**, and it needs no exception at all: `start()` over a running automation makes the automator lose its handle, so `stop()` stops nothing. Filed with repro, bisect and a candidate fix as https://github.com/zauberzeug/rosys/issues/461. With that patch applied, `abort()` finds the new automation and 1 becomes exactly what you described — so it dissolves rather than needing anything here.

Ship it. Only ask: a fourth limitation bullet, that while the automator can lose track of its automation, a late failure from a previous one is attributed to the current one and `forward()` stops on it.

evnchn:
So basically 451 is a "can merge, 461 pre-existing" kind of situation, which is a good one.

ftilde:
> So basically 451 is a "can merge, 461 pre-existing" kind of situation, which is a good one.

Alright, I've added the bullet point. So: Approve?

evnchn (approved):
Approved. 

Bullet covers it, CI green on afd7e4d4. 

#461 tracked separately.

falkoschindler (approved):
Looks good, thanks — and thanks @evnchn for the thorough first pass; the `last_exception` attribution and the sticky semantics were already settled in that thread and #461, so I left both alone.

I pushed 679c3db5 with a few small things from my own review, so you don't have to go through another round — revert whatever you disagree with:

- `forward()` now does the failure check once at the top of a `while True` loop instead of once before the loop and once after every step; same semantics (check, condition, timeout, step, …), no closure, no comment
- `_forget_helper_globals()` is only called on teardown of `rosys_integration` — the setup call never changed anything, since the previous test already cleaned up and the first test starts from `None`; I adjusted the corresponding bullet in the description
- `test_a_new_automation_forwards_past_an_earlier_failure` asserts that the second automation actually completed; `automator.is_stopped` was true before `working()` even started
- `test_a_failure_stops_forwarding_to_a_condition_that_already_holds` lost its second `pytest.raises` block — `forward(x=0.0)` hit the same pre-loop check as `forward(until=lambda: True)` and was the only reason for the `odometer` fixture
- the `last_exception` docstring matches its neighbours in style and mentions that `start()` resets it even from an `AUTOMATION_FAILED` handler — a subscriber that restarts the automator on failure clears it inside the emit, so `forward()` will not see that failure; not a regression, just worth knowing

Gates are green locally (pre-commit, mypy, pylint, 361 tests). [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%20findings%20and%20wording%20from%20the%20model%2C%20fix%20selection%20and%20sign-off%20from%20the%20reviewer.&a=claude-code "claude-fable-5: Review findings and wording from the model, fix selection and sign-off from the reviewer.")

PR #452, opened 2026-08-12
Parse the Lizard project name and align version normalization

falkoschindler:
### Motivation

zauberzeug/lizard#268 reports the firmware's project name in front of the version (`version: lizard-zz v0.0.2`), so a robot can tell integrator firmware like lizard-zz from public Lizard. The current parsers predate this. They also normalize inconsistently: `read_core_version` cuts the response at the first `-` (a dev build becomes `0.13.0`), while `read_local_version` string-splits the raw bin header and keeps the full `git describe` suffix — so `Local:` and `Core:` disagree on every dev build and the local-flash button never disappears.

### Implementation

- One regex-based parser extracts an optional project name and the release version from Core and P0 responses; firmware that does not report a project name yet counts as `lizard`.
- `read_local_version` reads the `esp_app_desc_t` structure at its fixed image offset (magic-word checked) instead of string splitting, yields the same normalized version plus the project name, and finds the binary by suffix (`build/*.bin`) because integrator binaries are named after their project (e.g. `lizard-zz.bin`).
- New attributes `core_project`, `p0_project`, `local_project` and `core_description`/`p0_description`/`local_description` properties; the developer UI shows e.g. `Core: lizard-zz 0.0.2` and compares project plus version for the local-update button.
- All versions now run through the same normalization, which fixes the permanent Local/Core mismatch on dev builds.

Parse results (firmware strings measured on hardware in zauberzeug/lizard-zz#15):

| Firmware reports | project | version | previous parser |
| --- | --- | --- | --- |
| `version: v0.13.0` | `lizard` | `0.13.0` | `0.13.0` |
| `version: v0.13.0-3-g248f4ed-dirty` | `lizard` | `0.13.0` | `0.13.0` |
| `version: lizard v0.14.0` | `lizard` | `0.14.0` | `0.14.0` |
| `version: lizard-zz v0.0.2` | `lizard-zz` | `0.0.2` | `0.0.2` |
| `version: lizard-zz v0.0.1-6-g69dfddf-dirty` | `lizard-zz` | `0.0.1` | `0.0.1` |
| `version: unknown` | – | – | `nknown` |

Known limitation, unchanged by this PR: the online version list points at `zauberzeug/lizard`, so `selected_online_version` can never match on a lizard-zz robot — worth a separate issue.

### Progress

- [x] Tests adapted and extended (parametrized parsing tests for old/new formats, dev builds and unparsable responses; local header parsing including a renamed binary); `tests/test_lizard_firmware.py` passes.
- [x] Tested on hardware.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Parser%20design%2C%20implementation%20and%20wording%20by%20the%20model%3B%20requirements%20from%20the%20hardware%20measurements%20and%20discussion%20on%20lizard-zz%2315.&a=claude-code "claude-fable-5: Parser design, implementation and wording by the model; requirements from the hardware measurements and discussion on lizard-zz#15.")

### DIFF
--- a/rosys/hardware/lizard_firmware.py
+++ b/rosys/hardware/lizard_firmware.py
@@ -1,4 +1,5 @@
 import logging
+import re
 import subprocess
 from pathlib import Path
 from typing import TYPE_CHECKING
@@ -19,6 +20,9 @@ class LizardFirmware:
     GITHUB_URL = 'https://api.github.com/repos/zauberzeug/lizard/releases'
     PATH = Path('~/.lizard').expanduser()
     PATH.mkdir(exist_ok=True)
+    VERSION_LINE_PATTERN = re.compile(r'version: (?:(?P<project>\S+) )?(?P<version>\S+)')
+    VERSION_PATTERN = re.compile(r'v?(\d+\.\d+\.\d+)')
+    APP_DESC_MAGIC = b'\x32\x54\xcd\xab'  # marks the esp_app_desc_t at offset 0x20 of an ESP32 app image
 
     def __init__(self, robot_brain: 'RobotBrain', *, supported_versions: str | None = None) -> None:
         """
@@ -34,8 +38,11 @@ class LizardFirmware:
         self.flash_params: list[str] = []
 
         self.core_version: str | None = None
+        self.core_project: str | None = None
         self.p0_version: str | None = None
+        self.p0_project: str | None = None
         self.local_version: str | None = None
+        self.local_project: str | None = None
         self.online_versions: dict[str, str] = {}
         self.selected_online_version: str | None = None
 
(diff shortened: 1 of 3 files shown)


evnchn:
`next(glob('*.bin'), fallback)` picks `ota_data_initial.bin` first on every real ESP-IDF build (verified on 4 build dirs) — it's ~8 KB, created before the firmware, and fails the magic check, so `read_local_version` silently returns without setting the version. Suggest iterating candidates and checking the magic before committing to parse. [※](https://zauberzeug.github.io/colophon/#m=qwen3.7-plus&t=Bug%20report%20and%20hardware%20verification%20by%20the%20model%3B%20logic%20confirmed%20by%20the%20operator. "qwen3.7-plus: Bug report and hardware verification by the model; logic confirmed by the operator.")

ftilde (approved):
Seems to work and looks good.

ftilde:
> `next(glob('*.bin'), fallback)` picks `ota_data_initial.bin` first on every real ESP-IDF build (verified on 4 build dirs) — it's ~8 KB, created before the firmware, and fails the magic check, so `read_local_version` silently returns without setting the version. Suggest iterating candidates and checking the magic before committing to parse. [※](https://zauberzeug.github.io/colophon/#m=qwen3.7-plus&t=Bug%20report%20and%20hardware%20verification%20by%20the%20model%3B%20logic%20confirmed%20by%20the%20operator.)

Oops I've missed that and your comment. I guess we can make a follow-up PR for that?

PR #463, opened 2026-08-21
Add curvature feed-forward to the spline driver

felixwrz:
### Motivation

The carrot follower is pure feedback: on a curved spline it only steers once an error has built up, so the command systematically trails the path. On slow vehicles this lag becomes a standing oscillation around the plan — measured on a stall-limited tracked robot (Feldfreund) following crop rows at 1 cm/s, the steering command lagged the path by 12–16 cm and the robot oscillated at the plan-piece wavelength. Raising the feedback gains instead drove the command into the drive's curvature ceiling and toward the loop's natural frequency.

Steering the known path curvature up front removes the lag at its source: in field tests the cross-track RMS on a prescribed spline chain dropped from 10.2 mm to 3.0 mm at 1 cm/s and from 8.9 mm to 3.2 mm at 8 cm/s — with otherwise unchanged default parameters, less steering activity (κ RMS ⅓), and no more contact with the curvature ceiling.

### Implementation

A new `DriveParameters.curvature_feedforward_gain` (default `0.0` — behavior is unchanged unless enabled):

- In the `drive_spline` loop, the spline's curvature at the robot's foot point is added to the feedback curvature: `curvature += gain * spline.curvature(t_foot)`. A gain of 1.0 matches the path geometry; higher values compensate the ground's yaw slip (tracked vehicles on soil typically need ~1.2–1.4).
- When driving backward, the sign flips: the spline is still traversed forward, so the required yaw rate is `curvature * |linear|`.
- The `minimum_turning_radius` clamp now applies to the **sum** of feedback and feed-forward — an excessive gain must not command a radius the drive cannot deliver (on stall-limited drives that would ask the inner wheel for less than it can physically turn). Reordering the clamp is a no-op for the forced-turn fallback, whose radius is never below the minimum, so gain 0.0 stays bit-identical.
- The feed-forward is skipped in the forced-turn fallback (carrot behind the robot while backward driving is disallowed), which deliberately commands a full-lock turn.
- `DriveState.curvature` includes the feed-forward, so recordings and UIs see the true command.

Tests: driving a curved spline forward **and** backward with gain 1.0 (the backward case pins the sign — a wrong sign veers off the path), plus a clamp test asserting an excessive gain never exceeds `1 / minimum_turning_radius`. Documentation: a note in the navigation example alongside the other carrot parameters.

### Progress

### DIFF
--- a/docs/examples/navigation/README.md
+++ b/docs/examples/navigation/README.md
@@ -28,5 +28,11 @@ You can display a wire frame version of the robot by passing `debug=true` to the
 > It turns the hook 180 degrees to the back of the robot, while preserving the distance `hook_offset` to the robot's coordinate center.
 > This allows the robot to drive backwards to a point behind it instead of turning around and approaching it forwards.
 
+> [!TIP]
+> The carrot mechanism is pure feedback: on a curved spline the robot only steers once an error has built up, so the command trails the path.
+> With `curvature_feedforward_gain` the driver additionally steers the spline's own curvature at the robot's current position up front,
+> leaving the feedback only the disturbances (1.0 matches the path geometry; higher values compensate yaw slip on loose ground).
+> The sum of feedback and feed-forward is limited by `minimum_turning_radius`.
+
 A more complex example can be found in the [RoSys GitHub repository](https://github.com/zauberzeug/rosys/tree/main/examples/obstacles).
 There you can create new obstacles and choose between straight driving or navigation.
--- a/docs/examples/navigation/README.md
+++ b/docs/examples/navigation/README.md
@@ -31,7 +31,8 @@ You can display a wire frame version of the robot by passing `debug=true` to the
 > [!TIP]
 > The carrot mechanism is pure feedback: on a curved spline the robot only steers once an error has built up, so the command trails the path.
 > With `curvature_feedforward_gain` the driver additionally steers the spline's own curvature at the robot's current position up front, leaving the feedback only the disturbances (1.0 matches the path geometry; higher values compensate yaw slip on loose ground).
-> The sum of feedback and feed-forward is limited by `minimum_turning_radius` (if set).
+> The feed-forward term itself is bounded by `curvature_feedforward_limit`, because near-degenerate splines (with control points very close to their end points) reach enormous curvature near `t=0` and `t=1`.
+> The sum of feedback and feed-forward is additionally limited by `minimum_turning_radius` (if set).
 
 A more complex example can be found in the [RoSys GitHub repository](https://github.com/zauberzeug/rosys/tree/main/examples/obstacles).
 There you can create new obstacles and choose between straight driving or navigation.
--- a/docs/examples/navigation/README.md
+++ b/docs/examples/navigation/README.md
@@ -30,9 +30,8 @@ You can display a wire frame version of the robot by passing `debug=true` to the
 
 > [!TIP]
 > The carrot mechanism is pure feedback: on a curved spline the robot only steers once an error has built up, so the command trails the path.
-> With `curvature_feedforward_gain` the driver additionally steers the spline's own curvature at the robot's current position up front,
-> leaving the feedback only the disturbances (1.0 matches the path geometry; higher values compensate yaw slip on loose ground).
-> The sum of feedback and feed-forward is limited by `minimum_turning_radius`.
+> With `curvature_feedforward_gain` the driver additionally steers the spline's own curvature at the robot's current position up front, leaving the feedback only the disturbances (1.0 matches the path geometry; higher values compensate yaw slip on loose ground).
+> The sum of feedback and feed-forward is limited by `minimum_turning_radius` (if set).
 
 A more complex example can be found in the [RoSys GitHub repository](https://github.com/zauberzeug/rosys/tree/main/examples/obstacles).
 There you can create new obstacles and choose between straight driving or navigation.
--- a/rosys/driving/driver.py
+++ b/rosys/driving/driver.py
@@ -25,6 +25,7 @@ class DriveParameters(ModificationContext):
     carrot_step_fraction: float = 0.1
     hook_bending_factor: float = 0.0
     curvature_feedforward_gain: float = 0.0
+    curvature_feedforward_limit: float = 5.0
     minimum_drive_distance: float = 0.01
     throttle_at_end_distance: float = 0.5
     throttle_at_end_min_speed: float = 0.01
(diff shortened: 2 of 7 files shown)


falkoschindler (changes requested):
Nice feature, and thanks for the thorough tests — the parametrized baseline and the dedicated tracking/clamp/NaN tests made the review much easier. I also double-checked the `flip_hook`-keyed sign of the feed-forward against the math and in simulation: it is correct (the proper sign depends on heading alignment with the tangent, not on the momentary drive direction), so no concerns there. One issue is blocking though:

**Near-degenerate splines hang the drive when the feed-forward is enabled.** The `np.isfinite` guard only catches the exact-NaN case, but a *near*-degenerate spline has huge finite curvature near its endpoints (`Spline.from_poses(..., control_dist=0.01)` reaches ≈53,000 at t=1). With the default `minimum_turning_radius = 0.0` the clamp below never fires, so `angular = linear * curvature` explodes and `_throttle` collapses the linear speed to ~1e-5 m/s. In simulation with `gain=1.0` the robot gets to ~0.1 mm before the goal, `foot` pins at t=1.0, and `drive_spline` never returns — even for a mild `control_dist=0.1`, which completes in ~26 s with `gain=0.0`. I'd suggest bounding the feed-forward magnitude independently of `minimum_turning_radius` (or skipping it near the parameter boundaries where the derivative vanishes); the existing NaN test can't catch this because an exactly degenerate spline is a straight chord with zero curvature elsewhere.

Two smaller, related points:

- The 0/0 in `Spline.curvature` emits a numpy `RuntimeWarning` before the `isfinite` guard runs — the new degenerate-spline test fails under `-W error::RuntimeWarning`. Consider handling the degenerate case inside `Spline.curvature` itself (return NaN or 0 explicitly for a vanishing denominator). That would also fix a latent hazard: with all-plain-float coefficients (e.g. a zero-length `Spline.from_points(p, p)`) the same expression raises `ZeroDivisionError`, which the guard can't catch — the NaN contract currently only holds because `Point.polar` happens to produce numpy scalars.
- The new `closest_point` call in the `can_drive_backwards` branch adds a second quintic `np.roots` solve per 10 Hz iteration, and the `foot` ratchet re-implements what `Carrot.move_by_foot` already does. Maybe `Carrot` could track the robot's foot in both modes, so `drive_spline` reads one attribute and doesn't depend on `move_by_foot` internals via the `foot = carrot.t` shortcut? At minimum, line 209's hook projection could get `t_min=foot` for free, giving it the same anti-lobe-jump robustness the NOTE argues for.

Nits, take or leave: guarding `spline.curvature(foot)` directly instead of the whole product would name the actual NaN source and drop the backslash continuation; the `(-1 if flip_hook else 1)` factor duplicates the sign already encoded in `hook_offset` on line 169; the two new test signatures at lines 403/420 exceed 120 columns (the sibling test wraps); `driver.pose.distance(spline.pose(foot))` works directly without the `.point` round-trips; and the `(dx=2, gain=1.0)` cell of the widened parametrization duplicates the tracking test's scenario with a weaker assertion. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%20and%20wording%20from%20the%20model%27s%20multi-agent%20review%3B%20the%20confirmed%20issues%20were%20reproduced%20in%20simulation.&a=claude-code "claude-fable-5: Findings and wording from the model's multi-agent review; the confirmed issues were reproduced in simulation.")

felixwrz:
Thanks for the thorough review — the near-degenerate hang was a good catch, and reproducible exactly as described. All points are addressed in 7b550dc:

**Blocker — feed-forward bound.** The feed-forward magnitude is now clamped by a new parameter `curvature_feedforward_limit` (default 5.0 ≈ 0.2 m radius), independent of `minimum_turning_radius` as suggested. That default sits far above any curvature a drivable spline asks for and four orders below the blow-up, so legitimate splines are unaffected. `test_curvature_feedforward_is_bounded_for_near_degenerate_splines` drives your exact repro (`control_dist=0.01`, `gain=1.0`, default parameters) — it hung before, now completes in ~20 s of simulated time. One consequence: the bounded feed-forward pulls the final heading of that near-degenerate drive to ~51° instead of the chord's 45°, so the new test asserts position only.

**Degenerate case in `Spline.curvature`.** The vanishing denominator is now handled at the source: curvature returns 0 explicitly (the correct limit — the degenerate spline is a straight chord), for scalars and arrays alike. That removes the `RuntimeWarning` (the touched tests pass under `-W error::RuntimeWarning`), fixes the latent `ZeroDivisionError` with plain-float coefficients (unit-tested via `Spline.from_points(p, p)`), and makes the `isfinite` guard in the driver obsolete — dropped entirely, which also resolves the guard-the-actual-source nit.

**Carrot / second `closest_point` solve.** I took the "at minimum" option: the hook projection now gets `t_min=foot` (behavior-identical when the gain is 0, since `foot` stays at 0.0). I left the "Carrot tracks the foot in both modes" refactor out deliberately: `move()` doesn't know the robot pose, so it would either grow a pose parameter or compute the quintic solve even when the feature is off — and at 10 Hz the extra `np.roots` is in the microseconds. Happy to do it if you feel the `foot = carrot.t` coupling is worth that trade.

**Nits** — all five taken: backslash continuation gone, the `(-1 if flip_hook else 1)` factor is now `np.sign(hook_offset.x)`, both long test signatures wrapped, `.point` round-trip dropped, and the duplicate `(dx=2, gain=1.0)` cell removed via explicit `(dx, gain)` pairs.

[※](https://zauberzeug.github.io/colophon/#m=claude&t=Analysis%2C%20fixes%20and%20wording%20by%20the%20model%3B%20the%20hang%20repro%20and%20the%20fixed%20behavior%20were%20verified%20in%20simulation.&a=claude-code "claude: Analysis, fixes and wording by the model; the hang repro and the fixed behavior were verified in simulation.")

falkoschindler (approved):
Thanks for the quick turnaround — 7b550dc addresses all points (verified locally, including `-W error::RuntimeWarning` and the plain-float spline), and I agree with leaving the Carrot refactor out.

A second pass caught one side effect, fixed in feaac949: `curvature()` returning 0.0 made the Delaunay planner accept degenerate splines it used to reject via NaN, breaking turn-in-place — `max_curvature()` now returns `inf` there, with a regression test. Same commit also reverts two suggestions from my first review that turned out to be wrong (`t_min=foot` on the hook projection could only cause premature end-throttling, and `np.sign(hook_offset.x)` has a needless zero edge) — sorry for the churn. Details in the commit message; ready to merge from my side. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Analysis%2C%20follow-up%20fixes%20and%20wording%20by%20the%20model%3B%20the%20planner%20regression%20and%20the%20fixed%20behavior%20were%20verified%20with%20tests%20and%20simulation.&a=claude-code "claude-fable-5: Analysis, follow-up fixes and wording by the model; the planner regression and the fixed behavior were verified with tests and simulation.")

PR #464, opened 2026-08-25
Fix Automator.stop() for automations started over a running one

### LINKED ISSUE #461: Automator.stop() does not stop an automation that was started over a running one
Starting an automation while another one is still running leaves the automator without a handle on the new one: `automator.automation` becomes `None`, `is_stopped` reports `True`, and `stop()`, `pause()` and `abort()` are no-ops on an automation that keeps running.

rosys 0.35.0 and main 42992147, Python 3.12, macOS 15 arm64.

## Repro

```bash
pip install rosys==0.35.0 pytest pytest-asyncio
printf '[pytest]\nasyncio_mode = auto\n' > pytest.ini
pytest test_mre_orphan.py
```

```python
import rosys
from rosys.automation import Automator
from rosys.testing import forward


async def counting(ticks: list[int]) -> None:
    while True:
        await rosys.sleep(1)
        ticks.append(1)


async def test_stop_works_if_the_previous_automation_was_stopped_first(automator: Automator):
    ticks: list[int] = []
    automator.start(counting(ticks))
    await forward(seconds=2)
    automator.stop(because='make room')  # the only difference to the test below
    await forward(seconds=1)

    automator.start(counting(ticks))
    await forward(seconds=2)
    automator.stop(because='user pressed stop')
    ticks_at_stop = len(ticks)
    await forward(seconds=5)
    assert len(ticks) == ticks_at_stop


async def test_stop_is_a_noop_after_a_restart(automator: Automator):
    ticks: list[int] = []
    automator.start(counting(ticks))
    await forward(seconds=2)
    # no stop() here: start() over the RUNNING automation is the only difference

    automator.start(counting(ticks))
    await forward(seconds=2)
    automator.stop(because='user pressed stop')
    ticks_at_stop = len(ticks)
    await forward(seconds=5)
    assert len(ticks) == ticks_at_stop, 'the automation kept running after stop()'
```

Expected: both pass. Actual: the control passes, the second fails, 5 runs out of 5.

```
E       AssertionError: the automation kept running after stop()
E       assert 7 == 2
E        +  where 7 = len([1, 1, 1, 1, 1, 1, ...])

FAILED test_mre_orphan.py::test_stop_is_a_noop_after_a_restart
1 failed, 1 passed in 0.10s
```

## Mechanism

`start()` calls `stop()`, whose `AUTOMATION_STOPPED` schedules `_handle_interrupt(stop=True)` as a task. By the time that task runs, `self.automation` is already the new `Automation`, whose `_is_waited` is still `False` because its own task has not had a turn — so `is_running` is `False`, the wait loop is skipped, and `self.automation = None` clears the new automation instead of the old one.

Handle state right after a restart, same setup as the repro:

```
automator.automation is the new one: False
automator.automation is None:        True
automator.is_stopped:                True
but the new automation ticked:       4 times
```

## Bisect

`self.automation = None` was synchronous inside `stop()` in b4545fed, running before `start()` assigned the new automation. 2baa91ca moved it into the async `_handle_interrupt`.

| commit | result |
|---|---|
| 6bf9a9ed, parent of 2baa91ca | 2 passed |
| 2baa91ca | 1 failed, 1 passed |
| 42992147, main | 1 failed, 1 passed |

## Possible fix

Capture the automation at emit time, while it is still the old one, and clear the field only if it is still that same automation:

```diff
-        self.AUTOMATION_PAUSED.subscribe(lambda _: self._handle_interrupt())
-        self.AUTOMATION_STOPPED.subscribe(lambda _: self._handle_interrupt(stop=True))
-        self.AUTOMATION_FAILED.subscribe(lambda _: self._handle_interrupt(stop=True))
+        self.AUTOMATION_PAUSED.subscribe(lambda _: self._handle_interrupt(self.automation))
+        self.AUTOMATION_STOPPED.subscribe(lambda _: self._handle_interrupt(self.automation, stop=True))
+        self.AUTOMATION_FAILED.subscribe(lambda _: self._handle_interrupt(self.automation, stop=True))

-    async def _handle_interrupt(self, *, stop: bool = False) -> None:
-        assert self.automation is not None
-        while self.automation.is_running:
+    async def _handle_interrupt(self, automation: Automation | None, *, stop: bool = False) -> None:
+        assert automation is not None
+        while automation.is_running:
             await rosys.sleep(0.1)
         if self._on_interrupt:
             if asyncio.iscoroutinefunction(self._on_interrupt):
                 await self._on_interrupt()
             else:
                 self._on_interrupt()
-        if stop:
+        if stop and self.automation is automation:
             self.automation = None
```

Repro passes with it on main 42992147, suite green (`372 passed, 7 skipped`). It does change `_on_interrupt` to run against the automation that was interrupted rather than whatever is current. Happy to open a PR.

falkoschindler:
### Motivation

Starting an automation while another one is still running leaves the automator without a handle on the new one: `automator.automation` becomes `None`, `is_stopped` reports `True`, and `stop()`, `pause()` and `abort()` are no-ops on an automation that keeps running.

Fixes #461.

### Implementation

`start()` calls `stop()`, whose `AUTOMATION_STOPPED` event schedules `_handle_interrupt(stop=True)` as a task. By the time that task runs, `self.automation` is already the new automation, which is not running yet, so the wait loop is skipped and `self.automation = None` clears the new automation instead of the old one (a regression from 2baa91ca, which moved the formerly synchronous cleanup into the async handler).

The review turned up more windows of the same family, so the PR follows two principles instead of patching the one case from the issue.

**Every callback that belongs to an automation receives that automation explicitly** instead of reading `self.automation` at call time.

- The event subscriptions pass `self.automation` into `_handle_interrupt`. `Event.emit` runs the lambdas synchronously, so this captures the automation the event refers to before `start()` replaces the field. The handler clears `self.automation` only if it is still that same one.
- `Automation` passes itself to its exception and completion handlers, so `_handle_exception` and `_on_complete` know which automation they belong to without the Automator closing over it (a closure would form a reference cycle, which matters because RoSys runs with `gc.disable()`). An exception in the cleanup of a superseded automation is logged and ignored instead of aborting the new one and being reported as its `last_exception` (found by @evnchn; on `main` this path was masked by the broken handle). A superseded automation that completes normally, which happens when an automation starts the next one itself and returns, no longer clears the handle or emits `AUTOMATION_COMPLETED` (found by @pascalzauberzeug).
- `start(paused=True)` pauses the automation it started, not whatever a re-entrant `AUTOMATION_STARTED` subscriber may have put into the field meanwhile.

**`start()` retires the old automation in every state**, and the new automation owns the robot from its first turn on.

- `stop()` used to decline for an automation that had not had its first turn yet (`start()` twice in the same event-loop turn, or `start()` directly followed by `stop()`) and for one with a pending pause inside an uninterruptible section (found by @pascalzauberzeug). `start()` replaced the field anyway, leaving the old automation running or parked in `_can_run.wait()` forever. `Automation.is_pending` and `Automator.is_pending` tell a scheduled-but-not-started automation apart from a finished one, and `stop()` and `abort()` treat a pending automation like a running one, so every `AUTOMATION_STARTED` gets its `AUTOMATION_STOPPED`. `is_stopped` is unchanged, a pending automation still counts as stopped.
- `_handle_interrupt` stops waiting once a new automation has taken over. In the plain restart case the handler task is scheduled before the new automation's task, so `on_interrupt` (typically `wheels.stop`) runs before the new automation's first turn, as on `main`. If the new automation has already had a turn when the handler gets there, which happens when `start()` is called while the old one is still cleaning up, `on_interrupt` is skipped: the new automation owns the robot and its commands must not be overridden by the old one's `wheels.stop` (found by @evnchn). In that case the wheels keep the old commanded speed until the new automation drives, as they did on `main`.

**Behavior changes to note:**

- A `stop()` during a pending pause is now executed instead of ignored, and the stop button of the automation controls is enabled in that state, matching the app controls.
- A failure in the cleanup of a superseded automation no longer sets `last_exception`, so `rosys.testing.forward()` does not fail on it. This follows `test_a_new_automation_forwards_past_an_earlier_failure`: a failure belongs to the automation that raised it. The exception is still logged with its traceback.
- `on_interrupt` is not called for an automation whose successor is already running. The docstring says so now (and no longer claims a string parameter that never existed).

Verified: each guard is pinned by a test that fails when the guard is removed; pre-commit, mypy, pylint, the full suite (391 passed, 7 skipped) and the examples smoke test are green.

### Progress

- [x] I chose a meaningful title.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Fix%20as%20proposed%20in%20the%20issue%2C%20extended%20after%20review%3B%20verification%2C%20regression%20tests%20and%20wording%20by%20the%20model.&a=claude-code "claude-fable-5-1: Fix as proposed in the issue, extended after review; verification, regression tests and wording by the model.")

### DIFF
--- a/rosys/automation/automator.py
+++ b/rosys/automation/automator.py
@@ -63,9 +63,11 @@ class Automator:
         if steerer:
             steerer.STEERING_STARTED.subscribe(lambda: self.pause(because='steering started'))
 
-        self.AUTOMATION_PAUSED.subscribe(lambda _: self._handle_interrupt())
-        self.AUTOMATION_STOPPED.subscribe(lambda _: self._handle_interrupt(stop=True))
-        self.AUTOMATION_FAILED.subscribe(lambda _: self._handle_interrupt(stop=True))
+        # NOTE: ``Event.emit`` runs these lambdas synchronously, so ``self.automation`` is captured while it is still
+        # the automation the event refers to -- even if ``start()`` replaces it before the coroutine actually runs
+        self.AUTOMATION_PAUSED.subscribe(lambda _: self._handle_interrupt(self.automation))
+        self.AUTOMATION_STOPPED.subscribe(lambda _: self._handle_interrupt(self.automation, stop=True))
+        self.AUTOMATION_FAILED.subscribe(lambda _: self._handle_interrupt(self.automation, stop=True))
 
         rosys.on_shutdown(self._stop_on_shutdown)
 
@@ -92,16 +94,16 @@ class Automator:
     def is_pausing(self) -> bool:
         return self.automation is not None and self.automation.is_pausing
 
-    async def _handle_interrupt(self, *, stop: bool = False) -> None:
-        assert self.automation is not None
-        while self.automation.is_running:
+    async def _handle_interrupt(self, automation: Automation | None, *, stop: bool = False) -> None:
+        assert automation is not None
+        while automation.is_running:
             await rosys.sleep(0.1)
         if self._on_interrupt:
             if asyncio.iscoroutinefunction(self._on_interrupt):
                 await self._on_interrupt()
             else:
                 self._on_interrupt()
-        if stop:
+        if stop and self.automation is automation:
             self.automation = None
 
     def start(self, coro: Coroutine | None = None, *, paused: bool = False) -> None:
--- a/tests/test_automations.py
+++ b/tests/test_automations.py
@@ -116,6 +116,28 @@ async def test_a_failure_stops_forwarding_to_a_condition_that_already_holds(auto
         await forward(until=lambda: True)
 
 
+async def test_stopping_an_automation_that_was_started_over_a_running_one(automator: Automator):
+    """``start()`` over a running automation must not orphan the new one (regression test for #461)."""
+    ticks: list[int] = []
+
+    async def counting() -> None:
+        while True:
+            await rosys.sleep(1)
+            ticks.append(1)
+
+    automator.start(counting())
+    await forward(seconds=2)
+    automator.start(counting())
+    await forward(seconds=2)
+    assert automator.automation is not None
+    assert automator.is_running
+    automator.stop(because='test')
+    ticks_at_stop = len(ticks)
+    await forward(seconds=5)
+    assert len(ticks) == ticks_at_stop, 'the automation kept running after stop()'
+    assert automator.is_stopped
+
+
 async def test_finally_block(automator: Automator):
     events: list[str] = []
 


rosys/automation/automator.py
```diff
@@ -92,16 +94,16 @@ def is_paused(self) -> bool:
...
         if self._on_interrupt:
             if asyncio.iscoroutinefunction(self._on_interrupt):
                 await self._on_interrupt()
             else:
                 self._on_interrupt()
-        if stop:
+        if stop and self.automation is automation:
             self.automation = None
 
     def start(self, coro: Coroutine | None = None, *, paused: bool = False) -> None:
```
pascalzauberzeug, line 109: **Note** — calling `start()` twice within the same event-loop turn still orphans the first automation: its task has not had a turn yet, so `_is_waited` is `False`, `is_stopped` reports `True`, and the `stop()` call below skips it entirely — no `AUTOMATION_STOPPED` event, no capture. Same family as #461, but fixing it needs `Automation` to distinguish "not started yet" from "stopped" — follow-up issue material, not this PR.

<details><summary>Probe test (fails on this branch, 7 ticks instead of 2)</summary>

```python
async def test_double_start_in_same_loop_turn(automator: Automator):
    ticks: list[int] = []

    async def counting() -> None:
        while True:
            await rosys.sleep(1)
            ticks.append(1)

    automator.start(counting())
    automator.start(counting())  # same synchronous block, first task never had a turn
    await forward(seconds=2)
    automator.stop(because='test')
    ticks_at_stop = len(ticks)
    await forward(seconds=5)
    assert len(ticks) == ticks_at_stop, 'an automation kept running after stop()'
```

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%20from%20the%20model%27s%20read%20and%20probe%20tests%3B%20triage%20and%20the%20design%20question%20from%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-fable-5: Findings from the model's read and probe tests; triage and the design question from the reviewer; wording by the model.")

rosys/automation/automator.py
pascalzauberzeug, line 1: **Note** — `start()` during `is_pausing` (pause requested inside an `@uninterruptible` section) hits the early return in `stop()` at line 159: the old automation is never stopped, the field is replaced anyway, and once its uninterruptible section ends the old task parks in `_can_run.wait()` forever — a dormant task leak with no handle left to resume or stop it. Pre-existing and invisible in behavior (the orphan never ticks); noting it for completeness only.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%20from%20the%20model%27s%20read%20and%20probe%20tests%3B%20triage%20and%20the%20design%20question%20from%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-fable-5: Findings from the model's read and probe tests; triage and the design question from the reviewer; wording by the model.")

pascalzauberzeug:
One question to discuss: should `start()` implicitly stop a running automation at all? Refusing to start while one is running (requiring an explicit `stop()` first) would remove this whole class of races instead of patching one window at a time.

evnchn:
The capture makes `_handle_interrupt` source-aware. `_handle_exception` still is not, and fixing the handle makes that reachable.

An exception in the **old** automation's cleanup reaches [`_handle_exception`](https://github.com/zauberzeug/rosys/blob/ed2b56c0aa631512a28d6beae3a7e023331041be/rosys/automation/automator.py#L205-L209), which calls `self.abort()` on whatever `self.automation` is now — the new one. On main that abort was a no-op because the handle was already broken; here it lands and kills the running new automation. `last_exception` reports the old automation's error as the current one's.

Same test both sides, old cleanup raising at t=4 while the new automation ticks every second: main `4fe54a5a` ticks at 4.0, 5.0, 6.0 — this PR, none.

Not in the diff, and not fixable inside it: `_handle_exception` needs the capture the interrupt handlers just got.

<details><summary>Repro (drop into <code>tests/test_automations.py</code>, passes on both, read the printed log)</summary>

```python
async def test_old_cleanup_exception_aborts_the_new_automation(automator: Automator):
    log: list[str] = []

    async def old() -> None:
        try:
            await rosys.sleep(100)
        finally:
            await rosys.sleep(3)
            log.append(f'{rosys.time():.1f} old cleanup raises')
            raise RuntimeError('cleanup of the old automation failed')

    async def new() -> None:
        while True:
            await rosys.sleep(1)
            log.append(f'{rosys.time():.1f} new tick')

    automator.start(old())
    await forward(seconds=1)
    automator.start(new())
    await forward(seconds=6, fail_on_automation_failure=False)
    print('\n' + '\n'.join(log))
    print(f'is_running={automator.is_running} last_exception={automator.last_exception!r}')
```
</details>

rosys/automation/automator.py
```diff
@@ -92,16 +94,16 @@ def is_paused(self) -> bool:
...
+        while automation.is_running:
             await rosys.sleep(0.1)
         if self._on_interrupt:
             if asyncio.iscoroutinefunction(self._on_interrupt):
                 await self._on_interrupt()
             else:
                 self._on_interrupt()
-        if stop:
+        if stop and self.automation is automation:
             self.automation = None
```
evnchn, line 107: Fires `_on_interrupt` only while this automation is still current, and drops the extra condition below:

```suggestion
        if self.automation is not automation:
            return  # a new automation has taken over in the meantime; it owns the robot now
        if self._on_interrupt:
            if asyncio.iscoroutinefunction(self._on_interrupt):
                await self._on_interrupt()
            else:
                self._on_interrupt()
        if stop:
            self.automation = None
```

Without it the old automation's `wheels.stop` lands after the new one drove: `linear_target_speed` 0.5 on main, 0.0 here. `Driver` re-issues every 0.1 s and masks it; a one-shot `wheels.drive()` keeps it. Suite green with it.

Cost: a new automation that does not drive immediately inherits the old commanded speed.

falkoschindler:
Confirmed, and fixed in 49f288b0: `_handle_exception` and `_on_complete` now receive their automation via closures in `start()`, and an exception from a superseded automation is logged and ignored instead of aborting the current one. Your repro is in as `test_an_exception_in_the_cleanup_of_a_superseded_automation_is_ignored`, with the tick count asserted to keep growing after the old cleanup raised. Thanks for the precise trace.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: You are right that the old `wheels.stop` must not land after the new automation drove. I took a slightly different route than skipping `on_interrupt` unconditionally: the wait loop now ends as soon as a new automation has taken over, and `on_interrupt` is skipped only if the new automation is already running at that point. In the plain restart case the handler task is scheduled in `stop()` before `start()` creates the new automation's task, so it runs first and `wheels.stop` lands before the first drive command, which is what `main` did and avoids the cost you named: a new automation that does not drive immediately does not inherit the old commanded speed. If the new automation is already running, because `start()` was called while the old one was still cleaning up, it owns the robot and the handler returns without touching the wheels, which is your suggestion for exactly the case where it is needed.

Your `linear_target_speed` check is in as `test_a_superseded_automation_does_not_override_the_new_drive_command`, parametrized over the state of the old automation: running, or still cleaning up after `stop()`.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

falkoschindler:
I would keep the implicit stop. Three reasons:

- It is the documented interaction pattern. `click-and-drive`, `obstacles` and `navigation` all call `automator.start(...)` on a click, typically while the robot is already driving. "Click a new target" is exactly the restart case.
- Refusing cannot work cleanly with async cleanup. `stop()` is synchronous and not awaitable, while the old automation's cleanup may await (that is what `_handle_interrupt` exists for). A refusing `start()` would either fail sporadically during `is_stopping`, so every caller needs retry logic, or would have to wait, which a synchronous method cannot. We would trade one race class for a worse API.
- The race class goes away by two principles rather than by patching windows: every callback that belongs to an automation gets its automation passed in instead of reading `self.automation` at call time, and `start()` retires the old automation in every state. The PR now does both, see the updated description.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: I pulled this into the PR after all, together with the `is_pausing` case from your other note, because both are the same family as #461 and the fix turned out to be small once the capture was in place: `Automation.is_pending` distinguishes "not started yet" from "finished", and `stop()` and `abort()` treat a pending automation like a running one, so it gets a regular `AUTOMATION_STOPPED` and its task never runs the body. `is_stopped` is untouched, so `forward()` keeps its "counts as stopped until its task has started" semantics. Your probe test is in as `test_stopping_an_automation_that_was_started_over_another_one[0.0]`, the same test as the #461 case with no gap between the two `start()` calls, extended with an assertion on the event sequence.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Also fixed, see the thread above: `stop()` no longer returns early on `is_pausing`, so `start()` over a pausing automation sets `_stop` and emits `AUTOMATION_STOPPED`; once the uninterruptible section ends, the old automation runs its cleanup instead of parking in `_can_run.wait()`. This is the one behavior change to be aware of: a `stop()` during a pending pause is now executed instead of ignored, which I think is what a user pressing stop expects. Covered by `test_starting_over_a_pausing_automation`.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

evnchn:
**Keep the implicit stop.** One reason beyond Falko's three: the permissive default composes, the strict one does not.

We agree on the premise — one automation at a time. The question is only what `start()` does when one is running.

| | implicit stop (today) | refusing `start()` |
|---|---|---|
| getting the *other* behaviour | one caller-side guard | a retry loop — `stop()` is sync, its cleanup is not |
| who pays for it | only the caller who wants strictness | every caller |
| question it forces open | none | how long to retry; how long may the old one overrun |
| how it fails | robot stops, new thing runs — visible | exception mid-script |

The guard, with nothing able to race in between:

```python
if not automator.is_stopped:
    return
automator.start(coro)
```

(Same sub-turn pending window this PR closes internally — a public `is_pending` would make it exact.)

Three reasons to keep it:

1. **It composes.** Pascal's semantics sit one guard away from today's default. Today's default is not reachable from a refusing `start()` without every caller writing the same retry loop.
2. **It fails where someone can see it.** The examples teach `start()` on a click, so the caller is often writing a ten-step script, not a framework. A refusal lands as an exception at step 3, and the common reflex is a bare `except:` — which, unlike `except Exception:`, also swallows `KeyboardInterrupt`. They cannot Ctrl-C out of their own script.
3. **The race is already handled.** Pascal's read of the race class was right; the two principles in this PR close it without charging every caller for it.

evnchn (approved):

rosys/automation/automator.py
```diff
@@ -92,16 +94,26 @@ def is_paused(self) -> bool:
...
+    def _is_pending(self) -> bool:
+        return self.automation is not None and self.automation.is_pending
+
+    async def _handle_interrupt(self, automation: Automation | None, *, stop: bool = False) -> None:
+        assert automation is not None
+        # NOTE: once a new automation has taken over, don't wait for the old one to finish its cleanup: in the plain
+        # restart case this handler runs before the new automation's first turn, so ``_on_interrupt`` (typically
+        # ``wheels.stop``) lands before its first drive command; if the new automation is already running (because the
+        # old one was still cleaning up when it started), it owns the robot and ``_on_interrupt`` is skipped
+        while automation.is_running and self.automation is automation:
```
pascalzauberzeug, line 107: **MAJOR** — Neither half of the conditional `on_interrupt` skip is pinned by a test, and the property promised in this thread does not hold in the `stopping` branch. Removing `and self.automation is automation` from this line leaves the full suite green (`386 passed, 7 skipped`), and so does replacing both the loop condition and the guard below with @evnchn 's unconditional variant — so the design point this thread settled has zero coverage. Could you add the test below and soften the claim to "skipped once the new automation has taken over and had a turn"?

<details><summary>Measurements and a ready test</summary>

The promised property, measured on this head and on `origin/main` (old automation drives 0.3, its `finally:` sleeps 1 s; `stop()`; then a new automation that sleeps 5 s before driving):

| | head 81767c36 | origin/main |
|---|---|---|
| plain restart (no `stop()` first) | 0.0 | 0.0 |
| after `stop()`, old still cleaning up | **0.3** | 0.3 |

So in the `stopping` branch the robot keeps the *stopped* automation's speed. Not a regression — but it is the branch your own test parametrizes, and the reply in this thread reads as if the conditional variant covered it.

Mutation results against `tests/test_automations.py` (baseline `28 passed`):

| mutation | file | full suite |
|---|---|---|
| delete the `if self.automation is not automation and …: return` guard | `1 failed` (`…does_not_override_the_new_drive_command[stopping]`) | — |
| delete `and self.automation is automation` from the loop | **`28 passed`** | **`386 passed, 7 skipped`** |
| evnchn's unconditional variant | **`28 passed`** | **`386 passed, 7 skipped`** |

This test passes on the current head and kills both surviving mutants with `assert 0.3 == 0.0`:

```python
async def test_a_superseded_automation_stops_the_wheels_the_new_one_leaves_alone(automator: Automator,
                                                                                wheels: Wheels):
    async def old() -> None:
        try:
            await wheels.drive(0.3, 0)
            await rosys.sleep(100)
        finally:
            await rosys.sleep(1)

    async def new() -> None:
        await rosys.sleep(100)

    automator.start(old())
    await forward(seconds=1)
    assert wheels.linear_target_speed == 0.3
    automator.start(new())
    await forward(seconds=2)
    assert wheels.linear_target_speed == 0.0, 'the superseded automation never stopped the wheels'
```

One more thing the test cannot see: the ordering guarantee in the `NOTE` holds only because `Wheels.drive` assigns `linear_target_speed` before its first `await`. With an `on_interrupt` that awaits before it commands, `[running]` fails with `0.0` — on this head and identically on `origin/main`, so the PR does not make it worse.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -200,14 +215,18 @@ def set_default_automation(self, default_automation: Callable | None) -> None:
...
+            self.log.warning('ignoring an exception from a superseded automation: %r', e)
+            return
         self.last_exception = e
         self.abort(because=f'an exception occurred in an automation{f": {e}" if str(e) else ""}')
         if rosys.is_test:
             self.log.exception('automation failed')
 
-    def _on_complete(self) -> None:
-        self.automation = None
+    def _on_complete(self, automation: Automation) -> None:
```
pascalzauberzeug, line 227: **MAJOR** — The guard protects the handle, but `AUTOMATION_COMPLETED.emit()` and `_notify('automation completed', 'positive')` two lines below still fire for a *superseded* automation while the new one is running — and the event carries no argument, so no subscriber can tell. An automation that calls `automator.start(...)` and then returns reaches this: measured `completed=[1.01]` with `ticks=4` and `is_running=True` on the new automation. Could you move both calls inside the `if`, and add the test below? It also pins the guard itself, which currently survives being removed.

<details><summary>Mechanism, measurement and a ready test</summary>

Why `on_complete` runs at all for a stopped automation — `automation.py:110-124`:

```python
if self._uninterruptible_depth == 0 and self._stop and not stop_injected:
    stop_injected = True
    send, message = iter_throw, AutomationStopped()
try:
    signal = send(message)
…
except StopIteration as err:
    self.log.info('automation is finished')
    if self.on_complete and not stop_injected:
        self.on_complete()
```

The `start()` happens *inside* `send(message)`, mid-turn. `_stop` is set synchronously, but the `not stop_injected` check runs only on the next loop iteration — which never comes, because that same `send(message)` raises `StopIteration` when the coroutine returns. So `stop_injected` is `False` and `on_complete()` fires.

The guard is load-bearing, not dead code: replacing it with an unconditional `self.automation = None` orphans the new automation — `handle=None`, `automator.is_stopped=True`, while it keeps ticking (1 → 4). The whole suite stays green either way (`386 passed, 7 skipped`), so nothing covers it today.

This test passes on the current head and fails with the guard removed (`assert False = Automator.is_running`):

```python
async def test_an_automation_that_starts_another_one_and_returns(automator: Automator):
    ticks: list[int] = []

    async def old() -> None:
        automator.start(_count(ticks))

    automator.start(old())
    await forward(seconds=3)
    assert automator.is_running
    automator.stop(because='test')
    ticks_at_stop = len(ticks)
    await forward(seconds=3)
    assert len(ticks) == ticks_at_stop, 'the automation kept running after stop()'
```

In downstream applications the unguarded emit is what hurts: a charger-parking routine, a "finished" notification and a recorded navigation event all trigger while the new automation drives.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -119,8 +131,11 @@ def start(self, coro: Coroutine | None = None, *, paused: bool = False) -> None:
             return
         self.stop(because='new automation starts')
         self.last_exception = None
-        self.automation = Automation(coro, self._handle_exception, on_complete=self._on_complete)
-        rosys.background_tasks.create(self.automation.run(), name='automation')  # type: ignore
+        automation = Automation(coro,
```
pascalzauberzeug, line 134: **MAJOR** — Both lambdas close over the local `automation`, and `Automation.__init__` stores them, so every `Automation` is now part of a reference cycle — while RoSys runs with `gc.disable()` and only collects below 300 MB of free memory. Measured over 20 sequential start-and-complete automations without an explicit collect: **20/20 still alive** on this head, **0/20** on `origin/main`. Could you let the `Automation` identify itself to its handlers instead? It is a net −2 lines, keeps the capture guarantee, and removes the subtlety the `NOTE` has to document.

<details><summary>Measurement, gc context and the tested patch</summary>

The cycle: `Automation` → lambda → closure cell → `Automation`. Before this PR the handlers were bound methods of the `Automator`, so an `Automation` was reachable only from `Automator.automation` and its task, and refcounting freed it immediately.

| | alive without `gc.collect()` | after `gc.collect()` |
|---|---|---|
| head 81767c36 | **20/20** | 0/20 |
| `origin/main` | 0/20 | 0/20 |
| patch below | 0/20 | 0/20 |

The coroutine frames are *not* retained (0/20 everywhere — the frame clears on completion). What stays is the `Automation`, its two lambdas, their shared closure cell and the `asyncio.Event`. Small per item, unbounded in count.

Why that matters here and not in an ordinary process — `rosys/rosys.py:401` and `:348-354`:

```python
gc.disable()  # NOTE disable automatic garbage collection to optimize performance

async def _garbage_collection() -> None:
    if psutil.virtual_memory().available < config.garbage_collection_mbyte_limit * 1_000_000:
        …
        gc.collect()
```

with `config.garbage_collection_mbyte_limit = 300` by default. So cyclic garbage is collected only once free memory drops below 300 MB; a robot that starts an automation per row accumulates one object graph per `start()` until then.

The patch — `_handle_exception(self, automation, e)` and `_on_complete(self, automation)` already take the automation first, so only the call sites move:

```python
# automator.py
automation = Automation(coro, self._handle_exception, on_complete=self._on_complete)

# automation.py, in __await__
self.exception_handler(self, e)
…
self.on_complete(self)
```

`Automation` is constructed in exactly two places (`automator.py:134` and `tests/test_automations.py:464`, which passes no handlers). Result: `tests/test_automations.py` 28 passed, full suite `386 passed, 7 skipped, 1 warning in 26.43s`, and the weakref probe drops to `alive_without_gc=False`.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -200,14 +215,18 @@ def set_default_automation(self, default_automation: Callable | None) -> None:
         """
         self.default_automation = default_automation
 
-    def _handle_exception(self, e: Exception) -> None:
+    def _handle_exception(self, automation: Automation, e: Exception) -> None:
+        if self.automation is not automation:
```
pascalzauberzeug, line 219: **MAJOR** — The early return is right (it is exactly evnchn's point), but it also removes the only test-visible signal for a failing superseded automation: `forward()`'s `fail_on_automation_failure` checks nothing but `automator.last_exception`. Measured with an old automation whose `finally:` raises, superseded by a new one, then a plain `await forward(seconds=5)`: this head **passes silently**, `origin/main` fails with `AssertionError: automation failed: RuntimeError(…)`. Could you name that in the body's "Behavior change to note" — and is a superseded failure meant to stay invisible to downstream suites, or should it surface some other way?

<details><summary>Measurement and why it matters downstream</summary>

Probe, with the **default** `fail_on_automation_failure`:

```python
async def old() -> None:
    try:
        await rosys.sleep(100)
    finally:
        raise RuntimeError('cleanup of the old automation failed')

automator.start(old())
await forward(seconds=1)
automator.start(_count(ticks))
await forward(seconds=5)
```

| | result |
|---|---|
| head 81767c36 | passes, `last_exception=None`, `ticks=4` |
| `origin/main` (both files swapped in) | `AssertionError: automation failed: RuntimeError('cleanup of the old automation failed')` at `rosys/testing/helpers.py:66` |

It is not silent in the log — this head still emits `[ERROR] automation.py:131: automation failed` with the traceback, plus the new warning — but `rosys/testing/helpers.py:65-66` is the only thing that turns a failure into a red test, and it reads `last_exception` alone.

That is the piece worth a sentence in the body: downstream applications restart automations routinely (a watcher, a schedule, mission chaining) and lean on `forward()`'s default to catch a crash in an automation's cleanup. After this PR that class of failure passes CI.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
pascalzauberzeug, line 1: **MAJOR** — This line still promises that `on_interrupt` "will be called when an automation pauses or stops", which `_handle_interrupt` now contradicts: with a new automation already running, the callback is skipped entirely (measured: `on_interrupt_calls=[]`). That is the intended fix for evnchn's point, but downstream this callback is a safety cleanup — `on_interrupt=<robot>.stop`, i.e. wheels *and* implement — so the silent skip deserves to be in the docstring that ends up on rosys.io, and the "Documentation is not necessary" checkbox with it. Could you make it "… unless a new automation has already taken over"?

<details><summary>Measurement and two pre-existing inaccuracies in the same line</summary>

Probe: old automation driving, `stop()`, old still in its cleanup, new automation started and running.

```
on_interrupt_calls=[]   linear_target_speed=0.5
```

The callback is not called at all for the stopped automation. In the examples shipped here that is `wheels.stop`; in an application it is typically a multi-module stop.

While you are in that line, two things that were already wrong before this PR:

- "(the cause is provided as string parameter)" — `_handle_interrupt` calls `self._on_interrupt()` with no arguments (`:113`, `:115`).
- it is also invoked on failure, via the `AUTOMATION_FAILED` subscription (`:70`), not only on pause and stop.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -154,9 +169,9 @@ def stop(self, because: str) -> None:
 
         :param because: the reason for stopping the automation
         """
-        if self.is_pausing or self.is_stopping:
+        if self.is_stopping:
```
pascalzauberzeug, line 172: **MAJOR** — Dropping `is_pausing` from this early return is the PR's headline behavior change, but rosys' own web UI cannot trigger it: `automation_controls_.py:24,31` computes `uninterruptible_is_running = automator.is_pausing or automator.is_stopping` and greys the stop button out exactly while `is_pausing`. So the user whose expectation the body cites cannot press the button — while `app_controls_.py:78` (`'disabled' if self.automator.is_stopped else 'enabled'`) leaves the Bluetooth stop enabled, i.e. the same robot behaves differently per control. Could the stop button's condition lose `is_pausing` too, and could the direct `stop()` path get the test below? The only test for this change goes through `start()`, which takes a different route afterwards.

<details><summary>The two paths, and a ready test for the direct one</summary>

`test_starting_over_a_pausing_automation` drives the change via `automator.start(new())`, i.e. `stop(because='new automation starts')`. In that path `_handle_interrupt` sees a successor and returns early. In the operator path — pause inside an `@uninterruptible` section, then press Stop — there is no successor, so the handler waits out the cleanup, calls `on_interrupt` and clears `self.automation`. Different code, currently unexercised.

This test passes on the current head and fails when the old `if self.is_pausing or self.is_stopping: return` is restored (`assert [] == ['finished']`):

```python
async def test_stopping_a_pausing_automation(automator: Automator):
    events: list[str] = []

    @rosys.automation.uninterruptible
    async def uninterruptible_section() -> None:
        await rosys.sleep(1)

    async def run() -> None:
        try:
            await uninterruptible_section()
            events.append('continued')
            await rosys.sleep(100)
        finally:
            events.append('finished')

    automator.start(run())
    await forward(seconds=0.5)
    automator.pause(because='test')
    assert automator.is_pausing
    automator.stop(because='test')
    await forward(seconds=3)
    assert events == ['finished']
    assert automator.is_stopped
```

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -167,7 +182,7 @@ def abort(self, because: str) -> None:
 
         :param because: the reason for aborting the automation
         """
-        if self.is_stopped:
+        if self.is_stopped and not self._is_pending:
```
pascalzauberzeug, line 185: **MAJOR** — Reverting this line to `if self.is_stopped: return` leaves the file at `28 passed` and the full suite at `386 passed, 7 skipped`; no test calls `automator.abort()` at all. Worth pinning, because the extension does not reach the case it was added for: a downstream emergency stop typically guards with `if automator.is_stopped or automator.is_stopping: return` before calling `abort()`, and during the pending window `is_stopped` is still `True` — measured `is_stopped=True, is_pending=True`, the guard returns, `abort()` is never reached and the automation starts and keeps ticking. Could you add the test below? And this is the second ask for a public `is_pending`, after @evnchn's "a public `is_pending` would make it exact" — a caller cannot see this window through the public properties today.

<details><summary>Measurement and a ready test</summary>

A representative downstream guard:

```python
def _handle_emergency_stop(self) -> None:
    if self.automator.is_stopped or self.automator.is_stopping:
        return
    self.automator.abort(because='an emergency stop was triggered')
```

Measured on this head, immediately after `start()` and before the automation's first turn:

```
is_stopped=True  is_stopping=False  is_pending=True  is_running=False
guard_returns_early=True
ticks_after_no_abort=2
```

`is_stopped` is deliberately unchanged by this PR, so the window this line opens is invisible to anyone reading the public properties — `_is_pending` is private, so a caller would have to reach through `automator.automation.is_pending`.

```python
async def test_aborting_an_automation_that_has_not_started_yet(automator: Automator):
    ticks: list[int] = []
    automator.start(_count(ticks))
    automator.abort(because='test')
    await forward(seconds=3)
    assert not ticks
    assert automator.is_stopped
```

It passes on the current head and fails with the line reverted (`assert not [1, 1]`).

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -119,8 +131,11 @@ def start(self, coro: Coroutine | None = None, *, paused: bool = False) -> None:
...
-        self.automation = Automation(coro, self._handle_exception, on_complete=self._on_complete)
-        rosys.background_tasks.create(self.automation.run(), name='automation')  # type: ignore
+        automation = Automation(coro,
+                                lambda e: self._handle_exception(automation, e),
+                                on_complete=lambda: self._on_complete(automation))
+        self.automation = automation
+        rosys.background_tasks.create(automation.run(), name='automation')  # type: ignore
         self.AUTOMATION_STARTED.emit()
         self._notify('automation started')
         if paused:
```
pascalzauberzeug, line 141: **MAJOR** — The PR's own principle ("every callback receives its automation explicitly instead of reading `self.automation` at call time") is broken on the line below: the local `automation` is in scope, but `self.automation.pause()` reads the field. `AUTOMATION_STARTED.emit()` and `_notify()` run their subscribers synchronously in between, so a subscriber that calls `automator.start(...)` re-entrantly makes this pause the wrong automation — measured: the re-entrantly started automation is the one paused, and the automator ends up at `is_paused=True, is_running=False` with nothing running at all. One word: `automation.pause()`.

<details><summary>Measurement</summary>

Probe with an `AUTOMATION_STARTED` subscriber that calls `automator.start(other())` once, then `automator.start(main_coro, paused=True)`:

| | main ticks | other ticks | `is_paused` | `is_running` |
|---|---|---|---|---|
| with `paused=True` | 0 | 0 | True | False |
| control, without `paused` | 0 | 5 | False | True |

The control shows the attribution: without the flag, `other()` runs; with it, `other()` — which the caller never asked to pause — is the one paused, and the automator stays paused forever.

On `origin/main` the same probe gives `main ticks=5, other ticks=0, is_paused=True`: the orphaned automation keeps ticking (#461) and `other()` is still wrongly paused.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -154,9 +169,9 @@ def stop(self, because: str) -> None:
 
         :param because: the reason for stopping the automation
         """
-        if self.is_pausing or self.is_stopping:
+        if self.is_stopping:
             return
-        if self.is_running or self.is_paused:
+        if self.is_running or self.is_paused or self.is_pausing or self._is_pending:
```
pascalzauberzeug, line 174: **Question** — Is there a state `or self.is_pausing` is meant to catch that `is_running` does not already cover? By the definitions in `automation.py`, `is_pausing` requires `_uninterruptible_depth > 0`, which makes both `is_paused` and `is_stopped` `False` and therefore `is_running` `True` — and removing the disjunct leaves the full suite at `386 passed, 7 skipped`. The behavior change for the pausing case comes from the early return two lines above, as far as I can tell.

<details><summary>The implication, from automation.py</summary>

```python
@property
def is_paused(self) -> bool:
    return self._is_waited and not self._can_run.is_set() and self._uninterruptible_depth == 0

@property
def is_pausing(self) -> bool:
    return self._is_waited and not self._can_run.is_set() and self._uninterruptible_depth > 0
```

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -92,16 +94,26 @@ def is_paused(self) -> bool:
...
-    async def _handle_interrupt(self, *, stop: bool = False) -> None:
-        assert self.automation is not None
-        while self.automation.is_running:
+    @property
+    def _is_pending(self) -> bool:
+        return self.automation is not None and self.automation.is_pending
+
+    async def _handle_interrupt(self, automation: Automation | None, *, stop: bool = False) -> None:
+        assert automation is not None
+        # NOTE: once a new automation has taken over, don't wait for the old one to finish its cleanup: in the plain
```
pascalzauberzeug, line 103: **CLEANUP** — These four lines mostly restate what the two lines below them do, and carry rationale ("which is what `main` did") that already sits verbatim in the PR description, where it stays findable and cannot age into a lie. Same for the two-line block at `:66-67`, which is background about NiceGUI's `Event.emit`. Could both come down to one line of *why* each, in line with the file's otherwise comment-free style?

<details><summary>Suggestion</summary>

```python
# a new automation owns the robot; the old one's cleanup must not override its commands
```

and, at `:66-67`:

```python
# capture the current automation: start() may replace the field before this coroutine runs
```

The rest — the scheduling order, what `main` did, who found which case — is already in the description.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

tests/test_automations.py
```diff
@@ -116,6 +116,121 @@ async def failing() -> None:
         await forward(until=lambda: True)
 
 
+async def _count(ticks: list[int]) -> None:
+    while True:
+        await rosys.sleep(1)
+        ticks.append(1)
+
+
+@pytest.mark.parametrize('gap', [2.0, 0.0])
```
pascalzauberzeug, line 125: **CLEANUP** — Four small readability points in the new block, all independent of correctness:

- this line produces the ids `[2.0]` and `[0.0]`, which say nothing — you had to explain in the thread that `[0.0]` means "same event-loop turn". `ids=['over a running automation', 'in the same loop turn']` puts it in the test name.
- `:134` `if gap:` tests a float for truthiness; `if gap > 0:` says what is meant.
- `:158` `old_state: str` where the same file already does this properly at `:469` (`method: Literal['pause', 'stop']`, with `Literal` imported at `:3`).
- `:197` `# the old automation raises after 3 s` restates the `rosys.sleep(3)` and `raise` four lines above.

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automation.py
```diff
@@ -81,13 +82,19 @@ def is_stopped(self) -> bool:
     def is_stopping(self) -> bool:
         return self._stop and not self.is_stopped
 
+    @property
+    def is_pending(self) -> bool:
```
pascalzauberzeug, line 86: **Note** — The new vocabulary covers `stop()` and `abort()` but not `pause()`/`resume()`: `start(coro, paused=True)` calls `Automation.pause()` directly, while `Automator.resume()` is gated on `is_paused`, which needs `_is_waited` — still `False` for a pending automation. So `start(paused=True)` followed by `resume()` in the same turn is a silent no-op and the automation parks forever before its first turn (measured `ticks=0, is_paused=True, is_running=False`, identical on `origin/main`, so pre-existing). Worth a follow-up issue so the next reader does not assume the pending hole is closed everywhere.

<details><summary>And one imprecision in the docstring</summary>

"scheduled but has not had its first turn yet" — `_stop` also clears the flag, so an automation stopped while pending reports `is_pending == False` although it still has not had its turn. That is deliberate (it keeps `stop()` idempotent), but the docstring reads as if only the first turn mattered.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20three%20review%20agents%20and%20three%20verification%20agents%20that%20ran%20every%20probe%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model.&a=claude-code "claude-opus-5: Findings from three review agents and three verification agents that ran every probe; triage and the decisions by the reviewer; wording by the model.")

pascalzauberzeug (changes requested):

rosys/automation/automator.py
falkoschindler: One correction to my reply above, prompted by Pascal's measurement in the thread below: the claim that the conditional variant avoids inheriting the old commanded speed holds for the plain restart case only. If `start()` is called while the old automation is still cleaning up, the new one has already had a turn when the handler gets there, `on_interrupt` is skipped, and the wheels keep the old speed until the new automation drives, as they did on `main`. The body now says so, and `test_a_superseded_automation_stops_the_wheels_the_new_one_leaves_alone` pins the restart case.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Done as proposed: `Automation` passes itself to its handlers, and the Automator hands over bound methods again. A weakref probe over 20 start-and-complete cycles reports 0/20 alive without a collect. Thanks, that also removed the closure subtlety the old comment had to explain.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Confirmed and fixed: `_on_complete` returns early for a superseded automation, so neither the event nor the notification fire. Your test is in as `test_an_automation_that_starts_another_one_and_returns`, extended with an assertion that `AUTOMATION_COMPLETED` was not emitted. Removing the guard fails it.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Named in the body under "Behavior changes to note". I would keep it that way: `test_a_new_automation_forwards_past_an_earlier_failure` documents that a failure belongs to the automation that raised it, so `last_exception` stays the current automation's. Setting it for a superseded automation would make `forward()` abort the new automation's test with the old one's error, which is what this PR fixes for `abort()`. A fixture that fails on any unhandled background-task exception would contradict that test and is a design of the testing helpers in its own right. The error stays in the log with its traceback.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Good catch. The stop button is now disabled only while stopping; pause and resume keep the uninterruptible condition. Your direct-path test is in as `test_stopping_a_pausing_automation`, and restoring the old early return fails it together with the `start()`-path test.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: `Automator.is_pending` is public now, and `test_stopping_an_automation_that_has_not_started_yet` is parametrized over `stop` and `abort`, asserting `is_pending` before and after the call. Reverting the `abort()` line fails it. `is_stopped` stays as it is, `forward()` and both control UIs build on a pending automation counting as stopped; the window is a sub-turn one, and `is_pending` makes it visible to a caller who needs it.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Your test is in as `test_a_superseded_automation_stops_the_wheels_the_new_one_leaves_alone`, and both surviving mutants die with it. The body and a follow-up in Evan's thread now say what the measurement shows: `on_interrupt` is skipped once the new automation has had a turn, and in that branch the wheels keep the old commanded speed until the new automation drives, as on `main`. Thanks for the mutation run.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Fixed, and pinned by `test_starting_paused_while_a_subscriber_starts_another_automation`, which fails with the field access restored.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Fixed all three: pauses, stops or fails; no string parameter; unless a new automation has already taken over. Checkbox updated.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Right, removed.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Done, one line each.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

tests/test_automations.py
falkoschindler: All four done.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automation.py
falkoschindler: Docstring made precise. No issue for the pause/resume case: `paused=True` is not used anywhere in this repository, and a `resume()` in the same synchronous block as `start(paused=True)` has no realistic caller. Known and deliberately not covered; if someone hits it, the fix is the same vocabulary in `resume()`.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
```diff
@@ -119,12 +128,13 @@ def start(self, coro: Coroutine | None = None, *, paused: bool = False) -> None:
             return
         self.stop(because='new automation starts')
```
pascalzauberzeug, line 133: **MAJOR** — `start()` emits `AUTOMATION_STOPPED` through `stop()` before it assigns the new automation, so a subscriber that starts an automation from that event is orphaned by the very `start()` that stopped the old one: the inner `start()`'s own `stop()` early-returns because the old automation is already `is_stopping`, and the outer `start()` overwrites the field two lines below. The subscriber's automation then keeps running with no handle and survives `automator.stop()` — the #461 shape this PR closes everywhere else, and the one case where "every `AUTOMATION_STARTED` gets its `AUTOMATION_STOPPED`" does not hold. Not a regression, `origin/main` loses the same automation, but the sibling door — re-entrancy from `AUTOMATION_STARTED` — was closed this round. Could `start()` stop again when a subscriber replaced the field, and the probe below go in?

<details><summary>Measurement, the five-event sweep and a three-line fix</summary>

Probe, failing on both sides:

```python
async def test_starting_while_a_stopped_subscriber_starts_another_automation(automator: Automator):
    ticks: list[int] = []
    stops = 0

    def start_another(_: str) -> None:
        nonlocal stops
        stops += 1
        if stops == 1:
            automator.start(_count(ticks))

    automator.AUTOMATION_STOPPED.subscribe(start_another)
    automator.start(_count([]))
    await forward(seconds=2)
    automator.start(_count([]))  # its stop() lets the subscriber start another automation in between
    await forward(seconds=3)
    automator.stop(because='test')
    ticks_at_stop = len(ticks)
    await forward(seconds=5)
    assert len(ticks) == ticks_at_stop, 'an automation kept running after stop()'
```

| | `AUTOMATION_STARTED` | `AUTOMATION_STOPPED` | ticks after `stop()` |
|---|---|---|---|
| head `9945d505` | 3 | 2 | 2 → 7 |
| `origin/main` (both files swapped in) | 3 | 1 | 2 → 7 |

Swept over all five events — one re-entrant `start()` from a subscriber, then `automator.stop()` and five more seconds:

```
STARTED:   fired=2 ticks=4 (at stop 4) handle=None
PAUSED:    fired=1 ticks=2 (at stop 2) handle=None
FAILED:    fired=1 ticks=3 (at stop 3) handle=None
COMPLETED: fired=1 ticks=3 (at stop 3) handle=None
STOPPED:   fired=2 ticks=7 (at stop 2) handle=None
```

Only `AUTOMATION_STOPPED` leaks; the other four retire the previous automation correctly.

One level of re-entrancy closes with three lines, and the probe then passes with `tests/test_automations.py` at `33 passed`:

```python
previous = self.automation
self.stop(because='new automation starts')
if self.automation is not previous:
    self.stop(because='new automation starts')  # a subscriber started another automation re-entrantly
```

A subscriber that starts an automation on *every* stop cannot be closed by a local guard — a `while` loop would spin — so if you would rather not chase it, the body's invariant could take the qualifier instead.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20one%20review%20agent%27s%20mutation%20matrix%20and%20probes%20plus%20the%20model%27s%20own%20reads%20and%20runs%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model. "claude-opus-5: Findings from one review agent's mutation matrix and probes plus the model's own reads and runs; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -92,16 +94,23 @@ def is_paused(self) -> bool:
     def is_pausing(self) -> bool:
         return self.automation is not None and self.automation.is_pausing
 
-    async def _handle_interrupt(self, *, stop: bool = False) -> None:
-        assert self.automation is not None
-        while self.automation.is_running:
+    @property
+    def is_pending(self) -> bool:
```
pascalzauberzeug, line 98: **CLEANUP** — The precision from the round-2 thread landed in `Automation.is_pending` but not here, and this is the copy that reaches rosys.io: `Automation` is not in `rosys/automation/__init__.py`'s `__all__`, which is what `docs/generate_reference_nav.py` iterates over, so the corrected wording is never rendered. This docstring still says "has been started but not had its first turn yet", while `_stop` clears the flag too — the PR's own test at `tests/test_automations.py:152-157` calls `automator.stop()` and then asserts `not automator.is_pending` for an automation that never had its turn. Could it mirror the class below?

```python
@property
def is_pending(self) -> bool:
    """whether an automation is scheduled but has neither had its first turn nor been stopped yet

    A pending automation counts as stopped.
    """
```

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20one%20review%20agent%27s%20mutation%20matrix%20and%20probes%20plus%20the%20model%27s%20own%20reads%20and%20runs%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model. "claude-opus-5: Findings from one review agent's mutation matrix and probes plus the model's own reads and runs; triage and the decisions by the reviewer; wording by the model.")

tests/test_automations.py
```diff
@@ -116,6 +116,211 @@ async def failing() -> None:
...
+    assert automator.is_stopped
+    assert events == ['started', 'stopped', 'started', 'stopped']
+
+
+@pytest.mark.parametrize('method', ['stop', 'abort'])
+async def test_stopping_an_automation_that_has_not_started_yet(automator: Automator, method: Literal['stop', 'abort']):
+    ticks: list[int] = []
+    automator.start(_count(ticks))
+    assert automator.is_pending
+    getattr(automator, method)(because='test')
```
pascalzauberzeug: **CLEANUP** — Two readability leftovers in the new block, both independent of correctness. `:152` uses `getattr(automator, method)(because='test')` where the same file spells the same shape out at `:580-583` (`if method == 'pause': … else: …`) — the very test the round-2 thread pointed at for the `Literal` annotation; `getattr` on a public API also takes the call out of mypy's reach. And `:161` is 122 columns against the 120 in `AGENTS.md`; no gate catches it, since `E501` is in ruff's ignore list (`pyproject.toml:206`), `C0301` is disabled for pylint (`:160`), and autopep8 does not split the continuation line. Worth folding into another commit rather than a round of its own.

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20one%20review%20agent%27s%20mutation%20matrix%20and%20probes%20plus%20the%20model%27s%20own%20reads%20and%20runs%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model. "claude-opus-5: Findings from one review agent's mutation matrix and probes plus the model's own reads and runs; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -154,9 +164,9 @@ def stop(self, because: str) -> None:
 
         :param because: the reason for stopping the automation
         """
-        if self.is_pausing or self.is_stopping:
+        if self.is_stopping:
```
pascalzauberzeug, line 173: **Note** — The early return this PR rewrote survives its own deletion: dropping `if self.is_stopping: return` leaves the full suite at `391 passed`. It is load-bearing — during the cleanup `_is_waited` is still `True`, so `is_running` reports `True` and a second `stop()` falls through, emitting `AUTOMATION_STOPPED` again and running `on_interrupt` a second time. Recording it so the line is not mistaken for covered ground; a test would be eight lines (`stop()` twice while the `finally` runs, assert the event fired once), verified to pass here and to fail with the line removed.

<details><summary>The test I ran</summary>

```python
async def test_stopping_twice_while_the_automation_cleans_up(automator: Automator):
    stops: list[str] = []
    automator.AUTOMATION_STOPPED.subscribe(stops.append)

    async def run() -> None:
        try:
            await rosys.sleep(100)
        finally:
            await rosys.sleep(2)

    automator.start(run())
    await forward(seconds=1)
    automator.stop(because='first press')
    await forward(seconds=1)
    assert automator.is_stopping
    automator.stop(because='second press')
    await forward(seconds=3)
    assert stops == ['first press']
```

```
$ # on 9945d505
1 passed in 0.02s
$ # with `if self.is_stopping: return` deleted from stop()
FAILED …::test_stopping_twice_while_the_automation_cleans_up
```

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20one%20review%20agent%27s%20mutation%20matrix%20and%20probes%20plus%20the%20model%27s%20own%20reads%20and%20runs%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model. "claude-opus-5: Findings from one review agent's mutation matrix and probes plus the model's own reads and runs; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
```diff
@@ -92,16 +94,23 @@ def is_paused(self) -> bool:
...
-        assert self.automation is not None
-        while self.automation.is_running:
+    @property
+    def is_pending(self) -> bool:
+        """whether an automation has been started but not had its first turn yet (it counts as stopped meanwhile)"""
+        return self.automation is not None and self.automation.is_pending
+
+    async def _handle_interrupt(self, automation: Automation | None, *, stop: bool = False) -> None:
+        assert automation is not None
+        while automation.is_running and self.automation is automation:
```
pascalzauberzeug, line 107: **Note** — The body's "`on_interrupt` (typically `wheels.stop`) runs before the new automation's first turn, as on `main`" reads as a safety guarantee, and it is one only for automations whose cleanup issues no further commands. The wait loop now ends as soon as a new automation is in the field, so the superseded automation is usually still running its `finally` when `on_interrupt` fires, and anything it commands afterwards wins — with nothing owning the robot, since that automation is dead and the new one has not driven yet. Measured `linear_target_speed == 0.2` on this head and identically on `origin/main`, so nothing this PR broke.

<details><summary>Measurement</summary>

```python
async def old() -> None:
    try:
        await wheels.drive(0.3, 0)
        await rosys.sleep(100)
    finally:
        await rosys.sleep(1)
        await wheels.drive(0.2, 0)   # a cleanup that repositions before it gives up

async def new() -> None:
    await rosys.sleep(100)

automator.start(old())
await forward(seconds=1)
automator.start(new())
await forward(seconds=3)
```

| | `linear_target_speed` after the restart |
|---|---|
| head `9945d505` | 0.2 |
| `origin/main` (both files swapped in) | 0.2 |

`on_interrupt` does run — the wheels go to 0.0 first — it just runs too early to be the last word.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20one%20review%20agent%27s%20mutation%20matrix%20and%20probes%20plus%20the%20model%27s%20own%20reads%20and%20runs%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model. "claude-opus-5: Findings from one review agent's mutation matrix and probes plus the model's own reads and runs; triage and the decisions by the reviewer; wording by the model.")

pascalzauberzeug (changes requested):
Round 3. I re-ran the twelve points from the previous round against this head instead of reading the replies: every named test, the mutation it is supposed to kill, the weakref probe for the reference cycle over 20 start-and-complete cycles with `gc.isenabled() == False`, and a re-entrancy sweep over all five automation events. All twelve hold. Thanks for the thoroughness — the handler signature change in particular came out cleaner than the closures it replaced.

That sweep is also where the one open point comes from, and it is the reason I am not approving yet rather than a judgement on this round's work.

Gates on `9945d505`: `391 passed, 7 skipped`, mypy clean, pylint 10.00, autopep8 clean on the touched files.

Not about the PR: `tests/test_cameras.py::test_rtsp_camera` and `tests/test_mjpeg_camera.py::test_mjpeg_camera` skip on one run and fail on the next on this machine — they want a local stream source. I deselected them for the mutation runs.

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Findings%20from%20one%20review%20agent%27s%20mutation%20matrix%20and%20probes%20plus%20the%20model%27s%20own%20reads%20and%20runs%3B%20triage%20and%20the%20decisions%20by%20the%20reviewer%3B%20wording%20by%20the%20model. "claude-opus-5: Findings from one review agent's mutation matrix and probes plus the model's own reads and runs; triage and the decisions by the reviewer; wording by the model.")

rosys/automation/automator.py
falkoschindler: Confirmed, and done as proposed: `start()` remembers the automation it is about to retire and stops again if a subscriber replaced it in between. Your probe is in as `test_starting_while_a_stopped_subscriber_starts_another_automation`, extended with an assertion on the event sequence (three `started`, three `stopped`), and it fails with the guard removed. The body names the one case that stays open, a subscriber that starts an automation on every stop, since a loop would spin there. Thanks for the sweep, that was the last door.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Mirrored, thanks for catching that this is the copy that gets rendered.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

tests/test_automations.py
falkoschindler: Both done: the call is spelled out as in the pause/stop test, and the signature is wrapped.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Your test is in as `test_stopping_twice_while_the_automation_cleans_up`, and the line now fails it when removed.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

rosys/automation/automator.py
falkoschindler: Agreed, and the body now says so: `on_interrupt` runs before the new automation's first turn, but the old automation is usually still in its cleanup at that point, and whatever the cleanup commands afterwards is not overridden, as on `main`. Waiting for the cleanup to end before calling `on_interrupt` would skip it in practically every restart, because the new automation has had its first turn by then, so I left the semantics as they are.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Analysis%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20author.&a=claude-code "claude-fable-5-1: Analysis and wording by the model, decisions by the author.")

pascalzauberzeug (approved):

PR #468, opened 2026-09-01
Add a resolution parameter to UsbCamera and SimulatedCamera

pascalzauberzeug:
### Motivation

Each camera type exposes its resolution differently: `MjpegCamera` has a single `resolution` parameter, `UsbCamera` two separate `width`/`height` parameters, and `SimulatedCamera` only a plain attribute. Generic tooling (camera UIs, config layers) has to special-case every type to answer "what resolution does this camera stream?".

### Implementation

- `UsbCamera`: replace the `width`/`height` parameters with one `resolution` parameter (v4l2 still sets both values on the device).
- `SimulatedCamera`: replace the constructor arguments and the `resolution` attribute with a `resolution` parameter; `SimulatedDevice.size` is now public and mutable so the parameter can be changed while connected — covered by a new test.
- `RtspCamera` is untouched: RTSP has no resolution control — the encoder profile (`substream`) determines it.
- Update tests and the camera docs examples; the `docs/examples/cameras/control.py` UI already handles a generic `resolution` parameter.

### Progress

---
⬛ claude-fable-5

### DIFF
--- a/rosys/vision/usb_camera/usb_camera.py
+++ b/rosys/vision/usb_camera/usb_camera.py
@@ -4,6 +4,7 @@ from typing import Any
 import numpy as np
 
 from ... import rosys
+from ...helpers.deprecation import deprecated_param
 from ..camera.configurable_camera import ConfigurableCamera
 from ..camera.transformable_camera import TransformableCamera
 from ..image import Image
@@ -14,6 +15,8 @@ from .usb_device import UsbDevice
 
 class UsbCamera(ConfigurableCamera, TransformableCamera):
 
+    @deprecated_param('width')
+    @deprecated_param('height', stacklevel=3)
     def __init__(self,
                  *,
                  id: str,  # pylint: disable=redefined-builtin
@@ -22,6 +25,8 @@ class UsbCamera(ConfigurableCamera, TransformableCamera):
                  auto_exposure: bool = True,
                  exposure: float = 0.01,
                  resolution: tuple[int, int] = (800, 600),
+                 width: int | None = None,
+                 height: int | None = None,
                  fps: int = 10,
                  **kwargs) -> None:
         super().__init__(id=id,
@@ -33,6 +38,8 @@ class UsbCamera(ConfigurableCamera, TransformableCamera):
         self.detect: bool = False
         self.color: str | None = None
 
+        if width is not None or height is not None:
+            resolution = (width or resolution[0], height or resolution[1])
         self._register_parameter('auto_exposure', self.get_exposure, self.set_exposure, auto_exposure)
         self._register_parameter('exposure', self.get_exposure, self.set_exposure, exposure)
         self._register_parameter('resolution', self.get_resolution, self.set_resolution, resolution)
@@ -42,6 +49,13 @@ class UsbCamera(ConfigurableCamera, TransformableCamera):
         return super().to_dict() | {
             name: param.value for name, param in self._parameters.items()
         }
+
+    @classmethod
+    def args_from_dict(cls, data: dict[str, Any]) -> dict:
+        data = super().args_from_dict(data)
+        if 'width' in data and 'height' in data:
+            data['resolution'] = (data.pop('width'), data.pop('height'))
+        return data
 
     @property
     def is_connected(self) -> bool:
--- a/rosys/vision/usb_camera/usb_camera.py
+++ b/rosys/vision/usb_camera/usb_camera.py
@@ -21,8 +21,7 @@ class UsbCamera(ConfigurableCamera, TransformableCamera):
                  connect_after_init: bool = True,
                  auto_exposure: bool = True,
                  exposure: float = 0.01,
-                 width: int = 800,
-                 height: int = 600,
+                 resolution: tuple[int, int] = (800, 600),
                  fps: int = 10,
                  **kwargs) -> None:
         super().__init__(id=id,
@@ -36,8 +35,7 @@ class UsbCamera(ConfigurableCamera, TransformableCamera):
 
         self._register_parameter('auto_exposure', self.get_exposure, self.set_exposure, auto_exposure)
         self._register_parameter('exposure', self.get_exposure, self.set_exposure, exposure)
-        self._register_parameter('width', self.get_width, self.set_width, width)
-        self._register_parameter('height', self.get_height, self.set_height, height)
+        self._register_parameter('resolution', self.get_resolution, self.set_resolution, resolution)
         self._register_parameter('fps', self.get_fps, self.set_fps, fps)
 
     def to_dict(self) -> dict[str, Any]:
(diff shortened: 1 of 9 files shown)


evnchn:
**Restoring an existing `cameras.json` raises `TypeError` on this branch.** The migration went in as 2f5dd1c and back out as 8920c7d, so persisted `width`/`height` dicts have no path forward.

Took a backup dict generated on `main` (400b7c5) and fed it to `CameraProvider.restore_from_dict` here:

```
SimulatedCamera: TypeError: object.__init__() takes exactly one argument
UsbCamera:       TypeError: object.__init__() takes exactly one argument
```

The same dict restores fine on `main`. The legacy keys land in `**kwargs` and travel up the MRO to `object.__init__`.

Deliberate for 0.x? If so, ignore — otherwise the reverted commit was already the fix.

evnchn:
Two things I measured on this branch, independent of the API-shape question:

- **`resolution` is the only parameter here that JSON can't round-trip.** Backup goes through `json.dumps` ([persistable.py:114](https://github.com/zauberzeug/rosys/blob/400b7c5ed184ae0b5a911393b9424cfc1e91a243/rosys/persistence/persistable.py#L114)), so `(800, 600)` comes back as `[800, 600]`, while `fps`, `exposure`, `auto_exposure` and `color` all restore as their own type. The declared `tuple[int, int]` is therefore a `list` until a getter refreshes the cache. The `==` short-circuit at [configurable_camera.py:59](https://github.com/zauberzeug/rosys/blob/400b7c5ed184ae0b5a911393b9424cfc1e91a243/rosys/vision/camera/configurable_camera.py#L59) is shape-sensitive, so re-applying the JSON-shaped `[800, 600]` misses the guard and re-fires the device write, where the tuple form is correctly skipped.
- **Stream dimensions now have two representations and the named one is going away.** `ImageSize(width, height)` and `(width, height)` coexist today, with `.tuple` bridging them, while `Detector.resolution` ([detector.py:55](https://github.com/zauberzeug/rosys/blob/400b7c5ed184ae0b5a911393b9424cfc1e91a243/rosys/vision/detector.py#L55)) is a scalar model input size. This PR drops `self.resolution = ImageSize(width=…, height=…)` and rebuilds it from `value[0]`/`value[1]` inside the new setter.

<details><summary>measured</summary>

```
pre-connect cache : [800, 600] list
post-connect cache: (800, 600) tuple
re-set same TUPLE -> setter calls: 0
re-set same LIST  -> setter calls: 1
```
Instrumented by wrapping the registered setter and counting invocations; `SimulatedCamera`, PR head.
</details>

evnchn:
Separate from the two defects above, on the shape itself: **could the other cameras move onto `width=`/`height=`, rather than these two moving onto a tuple?** Two named ints answer *"what resolution does this camera stream?"* just as generically, and they don't ask the codebase to agree on things cameras don't actually agree about.

A positional shape needs a fleet-wide convention on axis **order**, **count** and **unit**:

- **Count isn't bounded at three.** Once a third slot is admissible there's no principled stop — a device carrying dimensions *and* channels *and* a time window is a 4- or 5-tuple, and every reader has to memorise the order. `width=`/`height=`/`depth=` never renumbers.
- **The third slot has competing claimants, and they're mutually exclusive.** Depth for a stereo camera, channels for a sensor that isn't RGB, a time axis for anything that buffers frames ahead. Whichever one wins, it wins for every camera in the codebase — including cameras nobody has designed for yet.
- **The unit isn't fixed either.** A flatbed's resolution is DPI, not pixels: `resolution=(600, 600)` is type-identical to a 600×600 sensor and means something entirely different. A line-scan or barcode sensor has one spatial axis, so `(w, 1)` is a fiction you can't distinguish from a real 1px sensor.
- **The obvious counter, and why I don't think it closes.** Each of those would get its own field or class, so the tuple never has to grow. But then a depth camera reads `resolution=(w, h)` *plus* `depth=n` — dimensions split across two parameters of two different shapes, which is the per-camera special-casing this PR set out to remove.

None of that arrived as a tidy design argument — it started as a handful of hot takes about future sensors, and the four bullets above are what survived trying to make them checkable. The originals, lightly cleaned:

<details><summary>here be dragons</summary>

> I'd like to make an argument for width and height kwargs over the resolution tuple. It has to do with a mysterious new camera with 4 colours instead of RGB, and — while a stretch — a magic ball that allows someone to see N seconds into the future. Core argument being: IF we ever have a 3rd dimension, the resolution tuple, or n-ple, is going to be a mess and makes zero sense?

> There can be so many ways the 3rd dimension could be like. Imagine a depth camera. What, of more colour, more depth, or more time information, do we entertain as the 3rd dimension? Chaos is sure to ensue. How about a 1D scanner, from a barcode or a flatbed scanner, also? Oh the shenanigans!

The magic ball is deliberately silly, and that's the load-bearing part: the slot gets claimed by a device nobody designed for, and the tuple has already committed. Same for the barcode scanner in the other direction — it doesn't have two axes to give.
</details>

Would moving mjpeg/rtsp toward named width/height (or the existing `ImageSize`) be preferable to adding a third tuple surface?

---
⬛ claude-opus-5 — hot takes evnchn's, rationalization and measurements mine

rosys/vision/usb_camera/usb_camera.py
```diff
@@ -21,8 +21,7 @@ def __init__(self,
                  connect_after_init: bool = True,
                  auto_exposure: bool = True,
                  exposure: float = 0.01,
-                 width: int = 800,
-                 height: int = 600,
+                 resolution: tuple[int, int] = (800, 600),
```
ftilde, line 27: I like the overall idea of grouping, but would prefer ImageSize. I think we can make it serialize properly. (This also adresses @evnchn's comment about tuple/list serialization roundtrips.

ftilde (changes requested):
Not necessarily a blocker, but I think we need an explicit decision what we do about breaking deserialization for existing camera persistence. Are we fine with that? Do we want migration code? Or smaller fix: Accept losing config, but warn instead of crashing (potentially? Not sure where the TypeError ends up).

rosys/vision/usb_camera/usb_camera.py
pascalzauberzeug: I like the idea of using ImageSize. I just used the same tuple the [MJPEG-camera](https://github.com/zauberzeug/rosys/blob/400b7c5ed184ae0b5a911393b9424cfc1e91a243/rosys/vision/mjpeg_camera/mjpeg_camera.py#L25) does

NiklasNeugebauer: In the past I usually kept the old parameters and marked them as deprecated with the decorators we have in rosys. Then for saving and loading I added migration code that updates old persistence to the new convention.

pascalzauberzeug: I brought them back in.
@ftilde I think we should discuss ImageSize for a future PR

ftilde (approved):
I can live with postponing using `ImageSize` since we apply the deprecation mechanism here. 👍

PR #469, opened 2026-09-03
Keep a long MCAP recording readable: duration rotation, run names, metadata and the log

pascalzauberzeug:
### Motivation

A recording that runs for hours was only half usable: it rotated into files whose
only connection was the clock, nothing in a file said what it was recorded during,
and the log lived somewhere else entirely. This makes a long recording something
you can still read a day later.

### Implementation

- Rotate by duration as well as size, cutting inside the writer so no message falls into the gap
- `start(name=..., metadata=...)`: the files of one recording share a name and number themselves as they rotate, and every one of them carries the caller's context as a JSON metadata record
- Distinguish what the recorder wrote itself from what exists deliberately: the disk budget rolls away its own numbered files first and keeps everything else
- `merge_recordings` joins recordings into one and carries their metadata over, adding a record describing the merge
- The recordings page groups the files of one run into a single entry and can merge a run in the background
- Record the log on a `/log` topic in the `foxglove.Log` schema, so the Logs panel shows the lines next to the data

<details>
<summary>Details</summary>

- `split()` is gone again: it was written for filing a recording away without interrupting it, and no caller ever wanted that.
- The merge writes beside the recordings under a name the listing ignores and only takes the target name once complete, so a merge in progress never appears as a half-written recording, and the sources are deleted only after the finished file is in place.
- The log handler drops records from the recorder's own logger. The recorder logs its "queue full" warning while holding the queue lock, so recording that line would re-enter the writer on the same thread and deadlock — under exactly the load where it matters.
- Measured on a Jetson in a downstream application: 500 MB rotation every eleven minutes at ~47 MB/min, a segment boundary indistinguishable from a normal sample interval, no dropped messages over a quarter of an hour, and a 540 MB merge in 39 s.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Research%2C%20implementation%20and%20tests%20by%20the%20agent%3B%20scope%20and%20design%20decisions%20by%20the%20author.&p=agent&a=claude-code "claude-opus-5: Research, implementation and tests by the agent; scope and design decisions by the author.")

### DIFF
--- a/rosys/analysis/recording/mcap_recorder.py
+++ b/rosys/analysis/recording/mcap_recorder.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import os
+import re
 import threading
 from collections.abc import Callable, Collection
 from datetime import UTC, datetime
@@ -43,6 +44,21 @@ _DEFAULT_PAYLOAD_BYTES = 1024
 """Assumed size of a payload whose footprint cannot be measured (a small sensor value)."""
 
 _DROP_WARNING_INTERVAL = 10.0  # seconds between 'queue full' warnings, so drops do not spam the log
+
+_AUTO_NAME = re.compile(r'\d{8}_\d{6}_\d{6}\.mcap')  # the timestamp names _open_new_file generates
+
+
+def is_auto_named(path: Path | str) -> bool:
+    """Whether a recording still carries its generated timestamp name.
+
+    Renaming a recording (see :meth:`McapRecorder.rename_recording`) marks it as worth
+    keeping: the disk budget deletes auto-named files first and renamed ones only when
+    those alone exceed the budget.
+
+    :param path: the recording file to test.
+    :return: ``True`` if the file name is a generated timestamp name.
+    """
+    return _AUTO_NAME.fullmatch(Path(path).name) is not None
 
 
 class _QueuedMessage(NamedTuple):
@@ -98,9 +114,11 @@ class RecordingSource(Protocol):
 class McapRecorder:
     """Records sensor data to MCAP files for replay and analysis in Foxglove Studio.
 
-    Supports automatic file rotation by size and disk budget enforcement. Peak disk usage is
-    ``max_total_size_mb + max_file_size_mb``: the budget is enforced only before a file is
-    opened, so the currently growing file can exceed it by up to one file's worth.
+    Supports automatic file rotation by size and duration, and disk budget enforcement.
+    Peak disk usage is ``max_total_size_mb + max_file_size_mb``: the budget is enforced only
+    before a file is opened, so the currently growing file can exceed it by up to one file's
+    worth. The budget deletes auto-named files (oldest first) before touching renamed ones,
+    so a recording renamed to be kept survives until kept files alone exceed the budget.
 
     Messages are enqueued from the event loop (cheap, non-blocking) and written to disk by a
     single background consumer via ``rosys.run.io_bound`` so that encoding, ZSTD compression
@@ -130,6 +148,7 @@ class McapRecorder:
         *,
         output_dir: Path | str = '~/.rosys/mcap',
         max_file_size_mb: float = 100,
+        max_file_duration: float | None = None,
         max_total_size_mb: float = 1000,
         chunk_size: int = 1_048_576,
         flush_interval: float = 1.0,
@@ -143,10 +162,15 @@ class McapRecorder:
 
         :param output_dir: directory recordings are written to (created if missing).
         :param max_file_size_mb: on-disk size at which the active file is rotated to a new one.
+        :param max_file_duration: seconds after which the active file is rotated to a new one
+            (default: no duration-based rotation). Checked as messages are written, so an idle
+            recording only rotates once data flows again.
         :param max_total_size_mb: disk budget for the directory; the oldest recordings are
-            deleted before a new file is opened to stay under it. Peak disk usage is therefore
-            ``max_total_size_mb + max_file_size_mb`` (the budget is enforced only before a file
-            is opened, so the growing file can exceed it by up to one file's worth).
+            deleted before a new file is opened to stay under it. Auto-named (timestamped)
+            files go first; renamed recordings are deleted only when they alone exceed the
+            budget. Peak disk usage is therefore ``max_total_size_mb + max_file_size_mb``
+            (the budget is enforced only before a file is opened, so the growing file can
+            exceed it by up to one file's worth).
         :param chunk_size: MCAP chunk size in bytes (larger chunks compress better and flush
             less often).
         :param flush_interval: seconds between background flushes of the queue to disk.
@@ -162,6 +186,7 @@ class McapRecorder:
         self.output_dir = Path(output_dir).expanduser()
         self.output_dir.mkdir(parents=True, exist_ok=True)
         self.max_file_size = int(max_file_size_mb * 1_048_576)
+        self.max_file_duration = max_file_duration
         self.max_total_size = int(max_total_size_mb * 1_048_576)
         self.chunk_size = chunk_size
         self.profile = profile
@@ -189,6 +214,8 @@ class McapRecorder:
         self._sources: list[RecordingSource] = []
         self._loop: asyncio.AbstractEventLoop | None = None  # captured at start for loop-safe emits
         self._message_count: int = 0
+        self._file_message_count: int = 0
+        self._file_started_at: float = 0.0
         self._dropped_message_count: int = 0
         self._last_drop_warning: float = float('-inf')
         self._is_recording: bool = False
@@ -311,7 +338,9 @@ class McapRecorder:
         The currently-recording file cannot be renamed (the writer holds it open).
         ``new_name`` is reduced to a bare filename and given a ``.mcap`` suffix;
         empty, whitespace-only or dots-only names are rejected (they would escape
-        the output directory) by returning ``None``.
+        the output directory) by returning ``None``. A renamed recording is deleted
+        by the disk budget only when renamed files alone exceed it (see
+        :func:`is_auto_named`).
 
         :param path: the recording to rename.
         :param new_name: the desired name (reduced to a bare ``.mcap`` filename).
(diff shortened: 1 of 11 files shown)


rosys/analysis/recording/naming.py
```diff
@@ -0,0 +1,43 @@
...
+    :param path: the recording file to read.
+    :return: the run's name and the part's number, or ``None`` for a file that is no numbered part.
+    """
+    match = _OWN_NAME.fullmatch(Path(path).name)
+    if match is None or match.group('run') is None:
+        return None
+    return match.group('run'), int(match.group('part'))
+
+
+def check_file_name(name: str) -> None:
```
ftilde: `ensure_plain_file_name`?

rosys/analysis/recording/naming.py
```diff
@@ -0,0 +1,43 @@
+"""How recordings are named, and how a name is read back."""
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+TIMESTAMP_FORMAT = r'%Y%m%d_%H%M%S_%f'  # microseconds -> unique per run
+
+# part <part> of the run <timestamp> or <timestamp>_<name>, or a single unnumbered <timestamp>.mcap
+_OWN_NAME = re.compile(r'(?P<run>\d{8}_\d{6}_\d{6}(?:_.+)?)_(?P<part>\d{2,})\.mcap|\d{8}_\d{6}_\d{6}\.mcap')
```
ftilde: I find the "own"/"kept"/... naming quite confusion. Can we make it uniform? Suggestion:
"kept" -> "finished"? Or is there still a semantic difference between "kept" and actually finished files?
"own/auto" -> "part" or "segment" or something?

ftilde: Maybe we want to store the segments (which are not yet finished for the user to consume, I guess? Except for debugging purposes on aborted recordings) in a subfolder for easier distinction?

rosys/analysis/recording/mcap_recorder.py
```diff
@@ -130,7 +128,9 @@ def __init__(
         *,
         output_dir: Path | str = '~/.rosys/mcap',
         max_file_size_mb: float = 100,
+        max_file_duration: float | None = None,
```
ftilde: i guess these relate to the segments during recording. should we name them as such? max_segment_duration/size_mb?

ftilde: or "max_segment_file_..."

rosys/analysis/recording/merging.py
```diff
@@ -0,0 +1,146 @@
...
+                                       message.publish_time)
+                    count += 1
+        writer.finish()
+    return count
+
+
+def _collected_metadata(sources: list[Path]) -> dict[str, dict[str, str]]:
+    """The distinct context and merge records of ``sources``, named for the merged file.
+
+    A file carries its records ahead of its first chunk, so the scan stops there without
```
ftilde: Implementation detail. I don't think a caller cares.

rosys/analysis/recording/mcap_recorder.py
```diff
@@ -353,6 +379,45 @@ async def reindex_unindexed(self) -> None:
...
+        if is_auto_named(target):
+            raise ValueError(f'{target.name} reads as a file the recorder wrote itself, which its budget deletes first')
+        if not sources or any(source.parent != self.output_dir or source == self.current_recording
+                              for source in sources):
+            raise ValueError('only finished recordings in the output directory can be merged')
+        if name in self._merging or target.exists():
+            raise FileExistsError(f'{target.name} exists or is being merged already')
+        self._merging.add(name)
+        try:
+            result = await rosys.run.io_bound(merge_into_place, sources, target, start_time_ns=start_time_ns)
```
ftilde: **MAJOR** — The disk budget deletes the recorder's own files oldest first whenever a part is opened, and a finished run's parts are exactly those files.
A merge that runs while a new recording rotates near the budget can therefore lose a part it has not opened yet: the merge fails and that part is gone.
Could `merge_into_place` open all sources (after the reindex step) before it starts writing?
An unlinked file stays readable through an open handle, so a budget deletion mid-merge no longer breaks it, and the merged file still holds the data.

<details><summary>Failure path, disk peak, alternative</summary>

Read from the code, not reproduced by a run:

1. A run has finished; its parts are the oldest own files in `output_dir`.
2. A new recording is running close to `max_total_size_mb`; the page's merge button starts `recorder.merge(...)` via `io_bound`.
3. `merge_recordings` opens the sources one after another. Meanwhile the writer thread rotates, `_enforce_disk_budget` runs and deletes the oldest own files — the parts not yet opened.
4. `open(source)` raises `FileNotFoundError`, the staging file is removed, the page shows "Could not merge", and the deleted parts are lost.

`_merging` guards only the target name, not the sources.

With the sources opened up front, the budget believes it freed the space, but it only comes back once the merge closes the handles. Together with the staging file `*.mcap.merge-*`, which the `*.mcap` glob does not count, the peak during a merge exceeds the documented `max_total_size_mb + max_file_size_mb` by roughly the merge's size — worth a line in the docstring. The file-descriptor count equals the number of parts, which is small.

Considered alternative: let the budget skip the sources of a running merge. That couples the budget to the merge and can leave the directory over budget with nothing it may delete.

</details>

rosys/analysis/recording/merging.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    found: dict[str, list[dict[str, str]]] = {METADATA_NAME: [], MERGE_METADATA_NAME: []}
+    for source in sources:
+        with open(source, 'rb') as stream:
+            for record in StreamReader(stream, emit_chunks=True).records:
+                if isinstance(record, (Chunk, DataEnd)):
+                    break
+                if not isinstance(record, Metadata):
+                    continue
+                records = found.get(_base_name(record.name))
```
ftilde: **Question** — `_collected_metadata` carries over only `recording*` and `merge*` records, so any other metadata record in a source is dropped, and the sources are deleted after the merge.
Is that intended?
`reindex` now copies all metadata records, and the `merge_recordings` docstring says the metadata of the sources is carried over.

<details><summary>Probe</summary>

A source written with the `mcap` writer holding the metadata records `recording` and `calibration`, merged with `merge_recordings([a], out)`:

```python
with open(out, 'rb') as f:
    print([m.name for m in make_reader(f).iter_metadata()])
# ['recording', 'merge']
```

`found.get(_base_name(record.name))` returns `None` for `calibration`, so the record is skipped.
If carrying everything over is wanted: deduplicate identical `(name, payload)` pairs and number clashing names, as is already done for `recording`.

</details>

rosys/analysis/recording/recordings_page_.py
```diff
@@ -66,19 +72,73 @@ def _in_range(info: RecordingInfo) -> bool:
...
+        def _run_row(run: str, parts: list[RecordingInfo]) -> None:
+            """Render the files of one run as one entry that expands into its parts.
+
+            The entry keeps the shape of a single recording's row, so a run and a lone
+            recording sit on the same edges; only the parts inside are indented.
+
+            :param run: the name the run's files share.
+            :param parts: the run's files, newest first.
+            """
+            first = datetime.fromtimestamp(min(part.mtime for part in parts), tz=local_tz)
```
ftilde: **CLEANUP** — `first` is the smallest mtime of the parts, i.e. when part 1 was last written, so the span starts one rotation interval after the run began (about 11 min at the measured 500 MB and 47 MB/min).
Could the start come from the run name's timestamp (`TIMESTAMP_FORMAT`, UTC, converted to `local_tz`)?

<details><summary>Why it shows</summary>

The run name carries the true start as a UTC timestamp, while the row renders local time from mtimes.
Next to each other, the name and the span disagree by the rotation interval plus the UTC offset.

</details>

ftilde (changes requested):
Some manual comments, some done with review skill.

Mostly fine, I think some semantics could use cleanup; and a potential problem with delete on small budgets.

rosys/analysis/recording/naming.py
pascalzauberzeug: Agreed, three words for two kinds of files was one too many. The branch now uses two, and they are folders: a **part** is a file the recorder writes itself, `parts/<run>_<part>.mcap`; a **kept** recording is every `.mcap` at the top level of the output directory (renamed, merged), which the budget spares. `is_auto_named` is gone, the folder decides. Your subfolder idea in the second comment is what settled it.

"Finished" is a different axis: a rotated part is finished but not kept.

Commits: 9ef9ffef.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20reviewer%2C%20verdicts%20by%20the%20author%2C%20changes%20and%20wording%20by%20the%20agent.&a=claude-code "claude-fable-5-1: Findings from the reviewer, verdicts by the author, changes and wording by the agent.")

rosys/analysis/recording/mcap_recorder.py
pascalzauberzeug: Renamed to `max_part_size_mb` and `max_part_duration` (attributes `max_part_size`, `max_part_duration`). `max_file_size_mb` existed before this PR, so the body carries a breaking note and downstream applications adapt on their next pin.

Commit: 9ef9ffef.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20reviewer%2C%20verdicts%20by%20the%20author%2C%20changes%20and%20wording%20by%20the%20agent.&a=claude-code "claude-fable-5-1: Findings from the reviewer, verdicts by the author, changes and wording by the agent.")

rosys/analysis/recording/naming.py
pascalzauberzeug: Renamed to `ensure_plain_file_name`.

Commit: 9ef9ffef.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20reviewer%2C%20verdicts%20by%20the%20author%2C%20changes%20and%20wording%20by%20the%20agent.&a=claude-code "claude-fable-5-1: Findings from the reviewer, verdicts by the author, changes and wording by the agent.")

rosys/analysis/recording/mcap_recorder.py
pascalzauberzeug: Good catch, the sources were guarded by nothing. `merge_recordings` now opens all of them before it writes a byte, so a part the budget deletes meanwhile is still read through its handle and ends up in the merged file. The recorder's docstring names the higher peak during a merge. No deterministic test for the race; opening up front is the fix a reader can check.

Commit: b17c6958.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20reviewer%2C%20verdicts%20by%20the%20author%2C%20changes%20and%20wording%20by%20the%20agent.&a=claude-code "claude-fable-5-1: Findings from the reviewer, verdicts by the author, changes and wording by the agent.")

rosys/analysis/recording/merging.py
pascalzauberzeug: Not intended, the docstring promised more than the code did. Every metadata record of the sources is now carried over, deduplicated and numbered per name the way `recording` already was; `merge` keeps its offset so the merge at hand owns the plain name. Test added with a foreign record.

Commit: b17c6958.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20reviewer%2C%20verdicts%20by%20the%20author%2C%20changes%20and%20wording%20by%20the%20agent.&a=claude-code "claude-fable-5-1: Findings from the reviewer, verdicts by the author, changes and wording by the agent.")

rosys/analysis/recording/merging.py
pascalzauberzeug: Moved out of the docstring; the reason for stopping at the first chunk is a one-line comment at the `break` now.

Commit: b17c6958.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20reviewer%2C%20verdicts%20by%20the%20author%2C%20changes%20and%20wording%20by%20the%20agent.&a=claude-code "claude-fable-5-1: Findings from the reviewer, verdicts by the author, changes and wording by the agent.")

rosys/analysis/recording/recordings_page_.py
pascalzauberzeug: Right, the smallest mtime is the end of part 1. The span now starts at the run's own timestamp, converted to the viewer's zone, so name and span agree.

Commit: fa63e3cb.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20reviewer%2C%20verdicts%20by%20the%20author%2C%20changes%20and%20wording%20by%20the%20agent.&a=claude-code "claude-fable-5-1: Findings from the reviewer, verdicts by the author, changes and wording by the agent.")

ftilde (approved):

PR #470, opened 2026-09-03
Share one SSL context, add request timeouts and guard udev scans in the vision helpers

NiklasNeugebauer:
> 1. #470 — vision I/O helpers **(this PR)**
> 2. #471 — RTSP and MJPEG reconnection
> 3. #472 — USB and simulated reconnection

### Motivation

Groundwork for device-level camera reconnection (see the stack above).
This PR unifies all HTTP clients to simplify its usage in the later PRs and help share code in the various camera `Device` objects.
In addition to code duplication this caused two minor issues:

- Every `httpx.AsyncClient` built its own SSL context, which costs tens of milliseconds of event loop time per request, and most requests ran without a timeout, so a silent camera could hold a request forever.
- libudev keeps a non-atomic reference count on its context, so two scans sharing it unguarded can corrupt it; creating a fresh context per scan instead costs several milliseconds.

Neither change alters behavior on its own, which is why they are split out of the reconnection PRs.

### Implementation

- New `rosys/vision/http.py` with `new_async_client()`: one cached SSL context (`functools.cache`) and an explicit default timeout of 5 s.
  The Arkvision REST client, the MJPEG camera provider, the OpenIPC settings interface and the Jovision RTSP interface use it instead of constructing `httpx.AsyncClient` directly.
- `usb_camera_scanner.py` reuses a single cached `pyudev.Context` and runs every scan under a `threading.Lock`.
  Only plain `(uid, device node)` strings leave the lock, because `pyudev.Device` objects hold a reference to the context and drop it on garbage collection.
- `.gitignore` learns about `.mypy_cache` (a stale in-tree cache crashed mypy on this branch).

### Progress

---
⬛ claude-fable-5-1

### DIFF
--- a/rosys/vision/http.py
+++ b/rosys/vision/http.py
@@ -1,0 +1,21 @@
+import functools
+import ssl
+
+import httpx
+
+DEFAULT_TIMEOUT = 5.0
+"""Seconds without data before a camera connection counts as lost."""
+
+
+@functools.cache
+def _ssl_context() -> ssl.SSLContext:
+    """Build the SSL context once: constructing one costs tens of milliseconds of event loop time."""
+    return httpx.create_ssl_context()
+
+
+def new_async_client(*, timeout: float | None = DEFAULT_TIMEOUT, **kwargs) -> httpx.AsyncClient:
+    """Create an `httpx.AsyncClient` with a shared SSL context and an explicit timeout.
+
+    The timeout is explicit because a silent stream must end the request rather than wait forever.
+    """
+    return httpx.AsyncClient(verify=_ssl_context(), timeout=timeout, **kwargs)
--- a/rosys/vision/http.py
+++ b/rosys/vision/http.py
@@ -18,4 +18,5 @@ def new_async_client(*, timeout: float | None = DEFAULT_TIMEOUT, **kwargs) -> ht
 
     The timeout is explicit because a silent stream must end the request rather than wait forever.
     """
-    return httpx.AsyncClient(verify=_ssl_context(), timeout=timeout, **kwargs)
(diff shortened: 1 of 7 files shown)


evnchn:
**Note** — `httpx.AsyncClient()` has defaulted to `Timeout(5.0)` since well below the `httpx>=0.25.0` floor, so no call site here ran untimed and `DEFAULT_TIMEOUT = 5.0` restates the existing default. `create_ssl_context()` measures ~4 ms for me (M4, httpx 0.28.1) rather than tens, worth re-checking on the robot. The deduplication is still worth having; it is the "hold a request forever" line that would send someone hunting a timeout bug that was never there.

rosys/vision/http.py
```diff
@@ -0,0 +1,21 @@
...
+    """Build the SSL context once: constructing one costs tens of milliseconds of event loop time."""
+    return httpx.create_ssl_context()
+
+
+def new_async_client(*, timeout: float | None = DEFAULT_TIMEOUT, **kwargs) -> httpx.AsyncClient:
+    """Create an `httpx.AsyncClient` with a shared SSL context and an explicit timeout.
+
+    The timeout is explicit because a silent stream must end the request rather than wait forever.
+    """
+    return httpx.AsyncClient(verify=_ssl_context(), timeout=timeout, **kwargs)
```
NiklasNeugebauer:
```suggestion
def new_async_client(*, timeout: float | None = DEFAULT_TIMEOUT, **kwargs) -> httpx.AsyncClient:
    """Create an `httpx.AsyncClient` with a shared SSL context and an explicit timeout. """
```

evnchn: This diff looks malformed...

evnchn:
image

This can't be right?

ftilde (approved):
Fixed a few small issues, otherwise good.

PR #471, opened 2026-09-03
Reconnect RTSP and MJPEG cameras automatically at the device level

NiklasNeugebauer:
> 1. #470 — vision I/O helpers
> 2. #471 — RTSP and MJPEG reconnection **(this PR)**
> 3. #472 — USB and simulated reconnection

### Motivation

Cameras lose their connection due to network glitches, a bad cable or a power hiccup.
Today a dropped connection ends the device's capture task for good, so the only way back is a full `CameraProvider` scan, which is heavy and, with `auto_scan=False`, may also surface cameras we deliberately don't want connected (common in test and development setups).

This supersedes #423: reconnection lives **inside each device**, where the connection already lives, and providers are no longer involved in reconnection at all.
The review feedback on #423 is folded in; the branch was split so the RTSP/MJPEG core can be reviewed on its own, with USB and simulated cameras following in the next PR of the stack.

We also introduce a base class for devices (see implementation section) to share the new funcitonality more directly.

### Implementation

**One self-healing loop for all devices.**
The new `CaptureDevice` base (`rosys/vision/capture_device.py`) runs one capture session at a time and reopens the stream `reconnect_interval` seconds (default 3.0) after a session ends or fails, until `shutdown()` tears it down.
Subclasses only supply `_run_session()` and say how the camera is reached.
The loop is a small state machine (`CaptureState`: `CONNECTING` / `STREAMING` / `REFUSED` / `STOPPED`).

**Backoff on explicit refusals by the camera**
When the camera answers without a stream (MJPEG: any non-200 response after the auth handshake; RTSP: `Unauthorized` on gstreamer's stderr), the device enters `REFUSED` and backs off to one attempt every 30 s (`MAX_RECONNECT_INTERVAL`) instead of giving up.
It stays `is_active` and recovers on its own once the answer changes. The most likely cause is that someone edits credentials on the camera or that its ip is corrected.

**Uniform connection-state contract** across devices and cameras:

- `is_connected`: the camera is actually delivering frames right now.
- `is_active`: a connection is wanted, i.e. the self-healing loop is alive (streaming or waiting to reconnect).
- `is_reconnecting`: `is_active and not is_connected`.

`connect()` and `disconnect()` keep their semantics: `connect()` creates the self-healing device (replacing one whose loop died), `disconnect()` stops the retries and tears it down.
`connect()` now always creates a device, even when no address is known yet; such a camera reports `is_active` but not `is_connected` and starts streaming as soon as the address appears.

**Providers only discover and rebind.**
`MjpegCameraProvider` and `RtspCameraProvider` no longer call `connect()`/`disconnect()`.
Their scan adds newly found cameras (which connect themselves via `connect_after_init`) and pushes a changed address into a running device through new `ip` setters; a deliberately disconnected camera is left alone even with `auto_scan` enabled.

**Supporting changes:**

- Back-off policy (`MIN_RECONNECT_INTERVAL`, `MAX_RECONNECT_INTERVAL`, `clamp_reconnect_interval`) lives in `rosys/vision/reconnect.py`, so devices do not depend on the camera base module.
  Every camera takes a `reconnect_interval`, clamps it to at least 0.1 s (zero would starve the event loop) and persists it in `to_dict()`.
- `url` becomes a derived, nullable property; `AxisMjpegDevice` derives its query string from `axis_settings` instead of patching a stored URL, and its setters no longer restart the session from inside the capture task (the frame loop notices the changed URL).
- `MjpegDeviceFactory` warns about an unknown vendor instead of raising; a camera whose settings have no address raises a domain error.
- `on_connect` fires once a session is actually connected, so parameters are really reapplied after every reconnect.
- Permanent misconfigurations (unknown vendor, missing address, refused stream) are logged once at `WARNING`/`ERROR` without a traceback, while the expected "waiting for an address" stays at `DEBUG`.

### Tests

`tests/test_camera_reconnect.py` covers the behavior without hardware:

- `MjpegDevice` against a local flaky HTTP server: reconnects after the stream drops, backs off after a `401` but retries at a new address, keeps one capture loop across an address change, stops streaming when shut down during a callback.
- `RtspDevice` with the gstreamer session stubbed: reconnects until `shutdown()`, backs off after a rejected login, keeps one loop across concurrent restarts, re-raises a cancellation of its caller.
- Camera/provider contract: `is_active` without a known address and while a login is rejected, `connect()` replaces a device whose loop died, providers add and connect new cameras, rebind moved cameras and leave disconnected ones alone.
- Policy: `reconnect_interval` is persisted, clamped and passed to a replaced device; no wait between attempts exceeds the cap.

### Progress

---
⬛ claude-fable-5-1

### DIFF
--- a/tests/test_camera_reconnect.py
+++ b/tests/test_camera_reconnect.py
@@ -1,0 +1,879 @@
+import asyncio
+import logging
+from contextlib import asynccontextmanager, suppress
+from unittest.mock import AsyncMock, patch
+
+import pytest
+from nicegui import background_tasks
+from rosys.vision.reconnect import MAX_RECONNECT_INTERVAL, MIN_RECONNECT_INTERVAL
+
+import rosys
+from rosys.testing import forward
+from rosys.vision import MjpegCamera, MjpegCameraProvider, RtspCamera, RtspCameraProvider
+from rosys.vision.mjpeg_camera.arkvision_mjpeg_device import ArkVisionMjpegDevice
+from rosys.vision.mjpeg_camera.axis_mjpeg_device import AxisMjpegDevice
+from rosys.vision.mjpeg_camera.mjpeg_device import CameraAddressUnknown, CaptureState, MjpegDevice
+from rosys.vision.mjpeg_camera.mjpeg_device_factory import MjpegDeviceFactory
+from rosys.vision.mjpeg_camera.motec_mjpeg_device import MotecMjpegDevice
+from rosys.vision.mjpeg_camera.openipc_zauberzeug_mjpeg_device import OpenIpcZauberzeugMjpegDevice
+from rosys.vision.rtsp_camera.rtsp_device import RtspDevice
+
+# A MAC that maps to a "GOODCAM" URL in both the RTSP and MJPEG vendor tables.
+# GOODCAM needs no settings interface, so the device is constructed without any network access.
+GOODCAM_MAC = '2c:6f:51:00:00:01'
+
+# A MAC that maps to AXIS, whose stream settings are part of the URL rather than a settings interface.
+AXIS_MAC = '00:40:8c:00:00:01'
+
+MOTEC_MAC = '2c:26:5f:00:00:01'
+ARKVISION_MAC = '18:fd:cb:00:00:01'
+OPENIPC_ZAUBERZEUG_MAC = '7a:7a:21:00:00:01'
+
+# A MAC no vendor table knows, so no stream URL can be built for it.
+UNKNOWN_VENDOR_MAC = '02:00:00:00:00:01'
+
+
+async def _no_stream(self) -> None:
+    """Stand-in for a capture session that never delivers a frame and never ends on its own."""
+    await rosys.sleep(60.0)
+
+
+def stalled_mjpeg_stream():
+    """Keep an `MjpegDevice` in a single capture session without touching the network."""
+    return patch.object(MjpegDevice, '_run_session', _no_stream)
+
+
+def stalled_rtsp_stream():
+    """Keep an `RtspDevice` in a single capture session without spawning gstreamer."""
+    return patch.object(RtspDevice, '_run_session', _no_stream)
+
+
+class FakeGstreamerProcess:
+    """Stand-in for the gstreamer process, so a device reports `is_connected` without spawning one."""
+
+    def __init__(self) -> None:
+        self.returncode: int | None = None
+
+    def terminate(self) -> None:
+        self.returncode = -15
+
+    async def wait(self) -> int | None:
+        return self.returncode
+
+
+async def _connected_rtsp_stream(self) -> None:
+    """Stand-in for a gstreamer session that comes up and reapplies parameters, but spawns nothing."""
+    self._capture_process = FakeGstreamerProcess()  # pylint: disable=protected-access
+    try:
+        await self._invoke_on_connect()  # pylint: disable=protected-access
+        await rosys.sleep(60.0)
+    finally:
+        self._capture_process = None  # pylint: disable=protected-access
+
+
+def connected_rtsp_stream():
+    """Keep an `RtspDevice` in a single *connected* capture session without spawning gstreamer."""
+    return patch.object(RtspDevice, '_run_session', _connected_rtsp_stream)
+
+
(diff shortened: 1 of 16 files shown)


evnchn (commented):
**MAJOR** — CI is red on `ruff check` (I001), reproduced with the pinned `v0.15.11`: this is the only error, and #472 already carries the corrected order.

tests/test_camera_reconnect.py
```diff
@@ -0,0 +1,879 @@
...
+import rosys
+from rosys.testing import forward
+from rosys.vision import MjpegCamera, MjpegCameraProvider, RtspCamera, RtspCameraProvider
+from rosys.vision.mjpeg_camera.arkvision_mjpeg_device import ArkVisionMjpegDevice
+from rosys.vision.mjpeg_camera.axis_mjpeg_device import AxisMjpegDevice
+from rosys.vision.mjpeg_camera.mjpeg_device import CameraAddressUnknown, CaptureState, MjpegDevice
+from rosys.vision.mjpeg_camera.mjpeg_device_factory import MjpegDeviceFactory
+from rosys.vision.mjpeg_camera.motec_mjpeg_device import MotecMjpegDevice
+from rosys.vision.mjpeg_camera.openipc_zauberzeug_mjpeg_device import OpenIpcZauberzeugMjpegDevice
+from rosys.vision.rtsp_camera.rtsp_device import RtspDevice
```
evnchn, line 19:
```suggestion
import pytest
from nicegui import background_tasks

import rosys
from rosys.testing import forward
from rosys.vision import MjpegCamera, MjpegCameraProvider, RtspCamera, RtspCameraProvider
from rosys.vision.mjpeg_camera.arkvision_mjpeg_device import ArkVisionMjpegDevice
from rosys.vision.mjpeg_camera.axis_mjpeg_device import AxisMjpegDevice
from rosys.vision.mjpeg_camera.mjpeg_device import CameraAddressUnknown, CaptureState, MjpegDevice
from rosys.vision.mjpeg_camera.mjpeg_device_factory import MjpegDeviceFactory
from rosys.vision.mjpeg_camera.motec_mjpeg_device import MotecMjpegDevice
from rosys.vision.mjpeg_camera.openipc_zauberzeug_mjpeg_device import OpenIpcZauberzeugMjpegDevice
from rosys.vision.reconnect import MAX_RECONNECT_INTERVAL, MIN_RECONNECT_INTERVAL
from rosys.vision.rtsp_camera.rtsp_device import RtspDevice
```

evnchn:
**Question** — `CaptureState.REFUSED` is set in `mjpeg_device.py` and nowhere else, so `RtspDevice` stays outside the state machine and the uniform contract holds for MJPEG only: `is_refused` is always `False` there, the 30 s cap never comes from `_retry_interval` (`_wait_before_retry` re-implements it against `_authorized`), and a UI reading the documented `is_refused` misses a rejected RTSP login. `test_rtsp_device_backs_off_after_rejected_login` patches `_run_session` to set `_authorized`, so it passes either way. Was the parallel flag deliberate, the stderr sniff being a guess where an HTTP status is not?

Smaller: the retry line in `_run_capture_task` is `log.info` for every reason, so a camera awaiting an address logs at INFO every `reconnect_interval`, not the `DEBUG` the body promises.

NiklasNeugebauer:
> **Question** — `CaptureState.REFUSED` is set in `mjpeg_device.py` and nowhere else, so `RtspDevice` stays outside the state machine and the uniform contract holds for MJPEG only: `is_refused` is always `False` there, the 30 s cap never comes from `_retry_interval` (`_wait_before_retry` re-implements it against `_authorized`), and a UI reading the documented `is_refused` misses a rejected RTSP login. `test_rtsp_device_backs_off_after_rejected_login` patches `_run_session` to set `_authorized`, so it passes either way. Was the parallel flag deliberate, the stderr sniff being a guess where an HTTP status is not?
> 
> Smaller: the retry line in `_run_capture_task` is `log.info` for every reason, so a camera awaiting an address logs at INFO every `reconnect_interval`, not the `DEBUG` the body promises.

Hm both sound sensible

ftilde (approved):
I've found and then fixed a few issues directly to speed up merging this PR.

I'm not super happy about the general shape of the changes (error prone with callbacks flying between parent/child classes), but we can clean that up later.

PR #472, opened 2026-09-03
Reconnect USB and simulated cameras at the device level too

NiklasNeugebauer:
> 1. #470 — vision I/O helpers
> 2. #471 — RTSP and MJPEG reconnection
> 3. #472 — USB and simulated reconnection **(this PR)**

### Motivation

Completes device-level reconnection for the remaining two camera families, so every camera in RoSys follows the same contract introduced for RTSP and MJPEG in the PR below this one.

### Implementation

**`UsbDevice` becomes a `CaptureDevice`.**
A capture session reads frames until the capture is lost (failed reads past `MAX_READ_FAILURES` or a closed handle, e.g. an unplugged cable).
The loop then re-resolves the `/dev/video*` node from the camera uid, which can change on re-plug (`find_device_node` picks the lowest-numbered node), reopens it and reapplies the camera's parameters through `on_connect`, including auto-exposure and exposure.
This replaces the `on_repeat` polling and lets reads run at the camera's native rate instead of being capped by the poll interval.
A node that exists but cannot be opened (another process holds it) is logged once at `WARNING`.

**`SimulatedDevice` fails like a real one.**
It gains `disconnect()` to simulate a cable drop and a `simulate_failing` flag that drops the connection randomly over time, so the failure simulation moves from `SimulatedCameraProvider` into the device, matching where recovery lives.
The provider's `simulate_device_failure` attribute stays as a deprecated alias (emits `DeprecationWarning`) for one release.

**Providers only discover.**
`UsbCameraProvider` and `SimulatedCameraProvider` stop connecting and disconnecting cameras; a `UsbCamera` created before its video device appears is `is_active` and connects on its own once the node shows up.

**Lifecycle.**
The `Camera` base registers `rosys.on_shutdown(self.disconnect)` once per camera instead of once per device, so devices recreated on every reconnect cycle no longer leak shutdown handlers.

**Docs.**
`docs/examples/cameras/README.md` gets an "Automatic Reconnection" section explaining `is_connected` vs `is_active`, and the control example gains a connect button and binds its controls to `is_active`.

### Tests

Added to `tests/test_camera_reconnect.py`:

- `UsbDevice` against a fake capture: reopens after a simulated cable drop, releases a capture that only fails to read, reapplies parameters after reconnect, keeps one capture loop across concurrent reconnects, is active without a video device and connects once one appears.
- `SimulatedCamera`: resumes images after `disconnect()` once `reconnect_interval` elapsed, reapplies parameters, passes its parameters to the device, and `camera.disconnect()` stops the loop.
- Providers leave disconnected cameras alone; the USB provider adds a camera for a new device.
- The deprecated alias warns, devices do not leak shutdown handlers, `is_active` tracks `connect()`/`disconnect()`.

### Progress

---
⬛ claude-fable-5-1

### DIFF
--- a/docs/examples/cameras/README.md
+++ b/docs/examples/cameras/README.md
@@ -47,6 +47,30 @@ It automatically updates every 0.1 seconds to detect and display new cameras, an
 {! examples/cameras/control.py !}
 ```
 
+## Automatic Reconnection
+
+Cameras can lose their connection due to network glitches, a bad cable or a power hiccup.
+Every camera reconnects on its own: once connected, the underlying device keeps trying to restore its stream every `reconnect_interval` seconds (default 3.0) for as long as the camera stays connected.
+The wait is owned by the camera, so it also applies to a device created later, and it is clamped to at least 0.1 s so a misconfigured `0` cannot starve the event loop.
+`RtspCamera` and `MjpegCamera` re-open their stream, `UsbCamera` re-opens the video device (even if its `/dev/video*` node changed), and `SimulatedCamera` resumes after a simulated drop.
+Reconnection runs until the camera is disconnected, so `disconnect()` both stops the retries and tears down the device.
+`is_connected` tells whether a camera is streaming right now, while `is_active` tells whether a connection is wanted at all, i.e. whether the camera keeps trying.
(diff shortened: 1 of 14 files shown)


ftilde (approved):
Rebased on top of main, applied some fixes directly.

PR #474, opened 2026-09-04
Upgrade to NiceGUI 3.16 and its new scene object system

pascalzauberzeug:
### Motivation

NiceGUI 3.16 rebuilt the scene objects into a modular system ([#6085](https://github.com/zauberzeug/nicegui/pull/6085)) and deprecated the `nicegui.elements.scene.scene_objects` module along with subclassing `Object3D` without a `component` parameter. Both warn on every start and will raise in NiceGUI 4.0.

### Implementation

- Import scene objects from `nicegui.elements.scene.objects` in the driving, path planning and vision objects ([#6085](https://github.com/zauberzeug/nicegui/pull/6085))
- Let `PathObject` inherit from `Group` instead of passing the `'group'` type string to `Object3D` ([#6085](https://github.com/zauberzeug/nicegui/pull/6085))
- Raise the NiceGUI minimum to 3.16.0, the first release that ships the new module
- Scan the log and video directories inside the worker thread rather than on the event loop, and absorb the `None` that NiceGUI has typed for `run.io_bound` since 3.14 ([#6056](https://github.com/zauberzeug/nicegui/pull/6056))
- Drop the shutdown guards from `rosys.run.io_bound` and `rosys.run.cpu_bound`, which duplicate what `run._run` has been doing all along ([#6056](https://github.com/zauberzeug/nicegui/pull/6056))
- Scan the log directory entirely in the worker thread and skip files that log rotation removes mid-scan
- Stat each log file once, in the worker, so rendering the list no longer touches the filesystem at all
- Drop the direct `lxml` pin, which only existed to lift a transitive NiceGUI dependency past Dependabot alert 115

No new tests: there are none for scene objects to extend, and what covers the import paths is that `import rosys` loads every one of these modules, so a wrong path fails the suite at collection time.

Measured on `import rosys` under NiceGUI 3.16: two `warn_once` messages before this change — the deprecated `scene_objects` module, and the one about subclassing `Object3D` without a `component` parameter — and none after.

`nicegui/elements/scene/objects/` first appears in v3.16.0; v3.14.0 and v3.15.0 do not carry it, so 3.16.0 is the lowest floor that works.

The page changes are more than `None` handling. `Path.glob` is a generator function, so `run.io_bound(logs_dir.glob, '*.log')` only built the generator inside the thread and left the actual directory scan to the event loop. `_find_log_files` now hands `run.io_bound` a helper that returns a list, which also collapses the two calls into one; `get_all_videos` got the same treatment.

The `rosys/run.py` simplification is cleanup that rides along rather than a consequence of the bump, and the commit message overstates it: `run._run` already carried the whole shutdown guard in v3.13.0 (`nicegui/run.py:61-72`) — the `app.is_stopping` pre-check, the filter on `'cannot schedule new futures after shutdown'`, the silent `CancelledError` catch — but declared `-> R` and returned a bare `return  # type: ignore`. [#6056](https://github.com/zauberzeug/nicegui/pull/6056) only made that return type honest, which is what let mypy surface the duplication here. `cpu_bound` lost its `is_stopping()` pre-check too. NiceGUI raises `'Process pool not set up.'` only when `run.process_pool is None`, and after `set_up()` the sole thing that clears it is `run.reset()`, a testing hook — `tear_down()` shuts the pool down but keeps the reference. So a call during shutdown reaches `_run` and gets the same `None` as `io_bound`, and the pre-check was as redundant as the guards next to it. `io_bound` stays a wrapper rather than becoming an alias, with a `NOTE` saying why: rosys documents `R | None` as its own contract, and NiceGUI 4.0 will raise `CancelledError` instead.

Left for follow-ups: `run.process_pool_start_method` ([#6117](https://github.com/zauberzeug/nicegui/pull/6117), v3.15.0) still resolves to an implicit `fork` on Linux, which warns on every `run.cpu_bound` and will switch to `spawn` in NiceGUI 4.0; and the scene groups in `path_object_.py` and `obstacle_object_.py` parent their children to the scene root instead of to themselves.

</details>

### Progress

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Migration%20written%20by%20an%20earlier%20model%20from%20the%20author%27s%20pointer%20at%20the%20deprecation%20warnings%3B%20follow-up%20changes%20and%20this%20body%20by%20the%20model%20after%20a%20review%20pass%2C%20triage%20by%20the%20author.&a=claude-code "claude-opus-5: Migration written by an earlier model from the author's pointer at the deprecation warnings; follow-up changes and this body by the model after a review pass, triage by the author.")

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ readme = "README.md"
 requires-python = ">=3.11, <3.14"
 keywords = ["robot", "framework", "automation", "control", "steer"]
 dependencies = [
-    "nicegui>=3.13.0,<4",
+    "nicegui>=3.16.0,<4",
     "pyserial>=3.5,<4",
     "numpy<3.0.0",
     "scipy>=1.7.2,<2",
--- a/rosys/analysis/logs_page_.py
+++ b/rosys/analysis/logs_page_.py
@@ -50,9 +50,10 @@ class LogsPage:
 
 
 async def _find_log_files(logs_dir: Path) -> list[Path]:
-    logs = await run.io_bound(logs_dir.glob, '*.log')
-    rotated = await run.io_bound(logs_dir.glob, '*.log.*')
-    paths = list({p.resolve(): p for p in [*logs, *rotated]}.values())
+    def glob_logs() -> list[Path]:
+        return [*logs_dir.glob('*.log'), *logs_dir.glob('*.log.*')]
+    logs = await run.io_bound(glob_logs) or []
+    paths = list({p.resolve(): p for p in logs}.values())
     return sorted(paths, key=lambda p: p.stat().st_mtime, reverse=True)
 
 
--- a/rosys/analysis/logs_page_.py
+++ b/rosys/analysis/logs_page_.py
@@ -50,11 +50,16 @@ class LogsPage:
 
 
 async def _find_log_files(logs_dir: Path) -> list[Path]:
-    def glob_logs() -> list[Path]:
-        return [*logs_dir.glob('*.log'), *logs_dir.glob('*.log.*')]
-    logs = await run.io_bound(glob_logs) or []
-    paths = list({p.resolve(): p for p in logs}.values())
-    return sorted(paths, key=lambda p: p.stat().st_mtime, reverse=True)
+    def scan() -> list[Path]:
+        paths = {p.resolve(): p for p in [*logs_dir.glob('*.log'), *logs_dir.glob('*.log.*')]}
+        modified_at: dict[Path, float] = {}
+        for path in paths.values():
+            try:
+                modified_at[path] = path.stat().st_mtime
+            except FileNotFoundError:
+                continue
+        return sorted(modified_at, key=modified_at.__getitem__, reverse=True)
+    return await run.io_bound(scan) or []
 
 
 def _file_info(path: Path) -> str:
--- a/rosys/analysis/videos_page_.py
+++ b/rosys/analysis/videos_page_.py
@@ -68,7 +68,7 @@ def format_video_name(video: Path) -> tuple[str, str]:
 
 
 async def get_all_videos() -> dict[str, list[Path]]:
-    all_videos = sorted(await run.io_bound(VIDEO_FILES.glob, '*.mp4'), reverse=True)
+    all_videos = sorted(await run.io_bound(VIDEO_FILES.glob, '*.mp4') or [], reverse=True)
     video_date_dict: dict[str, list[Path]] = defaultdict(list)
     for video in all_videos:
         video_date = video.stem.split('_')[0]
--- a/rosys/analysis/videos_page_.py
+++ b/rosys/analysis/videos_page_.py
@@ -68,7 +68,7 @@ def format_video_name(video: Path) -> tuple[str, str]:
 
 
 async def get_all_videos() -> dict[str, list[Path]]:
-    all_videos = sorted(await run.io_bound(VIDEO_FILES.glob, '*.mp4') or [], reverse=True)
+    all_videos = await run.io_bound(lambda: sorted(VIDEO_FILES.glob('*.mp4'), reverse=True)) or []
     video_date_dict: dict[str, list[Path]] = defaultdict(list)
     for video in all_videos:
         video_date = video.stem.split('_')[0]
--- a/rosys/driving/driver_object.py
+++ b/rosys/driving/driver_object.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import numpy as np
 from nicegui import ui
-from nicegui.elements.scene.scene_objects import Group, Sphere
+from nicegui.elements.scene.objects import Group, Sphere
 
 from ..rosys import config
 from .driver import Driver
--- a/rosys/driving/robot_object_.py
+++ b/rosys/driving/robot_object_.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from nicegui import ui
-from nicegui.elements.scene.scene_objects import Extrusion, Group, Stl
+from nicegui.elements.scene.objects import Extrusion, Group, Stl
 
 from ..geometry import Prism
 from ..rosys import config
--- a/rosys/pathplanning/area_object_.py
+++ b/rosys/pathplanning/area_object_.py
@@ -1,4 +1,4 @@
-from nicegui.elements.scene.scene_objects import Group, Object3D
+from nicegui.elements.scene.objects import Group, Object3D
 
 from .area import Area
 from .area_manipulation import AreaManipulation, AreaManipulationMode
--- a/rosys/pathplanning/obstacle_object_.py
+++ b/rosys/pathplanning/obstacle_object_.py
@@ -1,6 +1,6 @@
 from contextlib import nullcontext
 
-from nicegui.elements.scene.scene_objects import Extrusion, Group
+from nicegui.elements.scene.objects import Extrusion, Group
 
 from .obstacle import Obstacle
 from .path_planner import PathPlanner
--- a/rosys/pathplanning/path_object_.py
+++ b/rosys/pathplanning/path_object_.py
@@ -1,15 +1,11 @@
 from contextlib import nullcontext
 
-from nicegui.elements.scene.scene_object3d import Object3D
-from nicegui.elements.scene.scene_objects import Curve
+from nicegui.elements.scene.objects import Curve, Group
 
 from ..driving import PathSegment
 
 
-class PathObject(Object3D):
-
-    def __init__(self) -> None:
-        super().__init__('group')
+class PathObject(Group):
 
     def update(self, path: list[PathSegment], height: float = 0) -> None:
         for obj in list(self.scene.objects.values()):
--- a/rosys/run.py
+++ b/rosys/run.py
@@ -40,7 +40,7 @@ def awaitable(func: Callable) -> Callable:
 
 async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R | None:
     if is_stopping():
-        return None
+        return None  # NiceGUI raises 'Process pool not set up.' while stopping, before its own shutdown guard runs
     with cpu():
         return await run.cpu_bound(callback, *args, **kwargs)
 
--- a/rosys/run.py
+++ b/rosys/run.py
@@ -27,16 +27,7 @@ log = logging.getLogger('rosys.run')
 
 
 async def io_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R | None:
-    if is_stopping():
-        return None
-    try:
-        return await run.io_bound(callback, *args, **kwargs)
-    except RuntimeError as e:
-        if 'cannot schedule new futures after shutdown' not in str(e):
-            raise
-    except asyncio.exceptions.CancelledError:
-        pass
-    return None
+    return await run.io_bound(callback, *args, **kwargs)
 
 
 def awaitable(func: Callable) -> Callable:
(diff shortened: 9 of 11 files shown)


evnchn:
**Good** — the central claim holds: 3.16 genuinely subsumes the shutdown handling `20e2a230` removed. Verified rather than read off the commit message.

<details><summary><b>Verified, with citations</b> — so nobody re-derives this; the diff is large and most of it is not cheap to re-check</summary>

- **`_run` carries all three deleted guards** — `nicegui/run.py::_run` @ `v3.16.0`: `core.app.is_stopping` early return, the `'cannot schedule new futures after shutdown'` `RuntimeError` filter, `except asyncio.CancelledError: pass`. Both `io_bound` and `cpu_bound` delegate to it. Same predicate as rosys's: `is_stopping()` reads `app.is_stopping`, and `nicegui.app is nicegui.core.app`.
- **The `3.16.0` floor is exactly right** — `nicegui/elements/scene/objects/` is **absent** at `v3.14.0` and `v3.15.0`, **present at `v3.16.0`** (35 files). Not merely "new-ish".
- **Deprecations actually cleared: 2 to 0** — base code against `nicegui==3.16.0` prints the `scene_objects` module warning *and* the `Object3D`-subclassing warning; this branch prints neither. :warning: Both arrive via nicegui's `warn_once` **logging**, not `DeprecationWarning` — a `warnings.catch_warnings` count reads `0` on both trees and will fool you. Capture stderr instead.
- **Python floor safe** — nicegui 3.16 is `Requires-Python >=3.10,<4`; rosys's 3.11-3.13 sits inside it.
- **Gates, re-run on this branch** — `pytest` 400 passed / 7 skipped (skip reasons read: 1 Linux-only `UsbCamera`, 6 `arp-scan` — none cover this change), `mypy` clean over 217 source files, `pylint` 10.00/10.
- **The page fix is real, not cosmetic** — base `_find_log_files` raises `TypeError: Value after * must be an iterable, not NoneType` while stopping; the branch returns `[]`.
- **Ruled out** — `or []` binds as `(await ...) or []`; no `scene_objects`/`scene_object3d` references remain in `rosys/`, `docs/` or `tests/`; dropping nicegui's `lxml`/`lxml-html-clean` is inert because rosys pins `lxml` directly; `BrokenProcessPool` (a `RuntimeError` subclass) propagated before and still does; `mcap_recorder.py:428`'s comment about `io_bound` refusing work while stopping is still accurate.

</details>

**MAJOR** — `rosys/analysis/logs_page_.py:57`: moving enumeration into the worker widens an unguarded `stat()` race into a page-crashing one. The race pre-dates this PR; the restructure is what makes it reachable.

- base passed `logs_dir.glob` (a *lazy* generator) to `io_bound`, so readdir actually ran on the loop, in the same uninterrupted block as `resolve()` and `stat()`
- `glob_logs()` materializes the list in the thread, so an await boundary now sits between readdir and `p.stat().st_mtime` — the paths are already fixed when the loop resumes
- the glob deliberately matches `*.log.*`, i.e. rotated files, and rotation prunes exactly those; `_file_info` five lines below already guards `FileNotFoundError`, so vanishing files are already treated as live here

```
this branch (20e2a230):  20/20 renders raised FileNotFoundError  (FileNotFoundError on robot.log.113)
main        (16a69da3):   0/20 renders raised FileNotFoundError  (clean)
positive control: raised as expected     <- on both trees, so main's 0/20 is not blindness
```

<details><summary>Repro — <code>nicegui 3.16.0</code>, Python 3.13.11, macOS 26.5.2 arm64</summary>

It calls `_find_log_files` directly; that is the same call `LogsPage._content()` makes one frame up with nothing in between, so this is the real render path rather than a bypass. Exits non-zero when the race fires, and aborts if its own positive control cannot see a raise.

```python
import asyncio
import sys
import tempfile
from pathlib import Path

from nicegui import core, run

from rosys.analysis.logs_page_ import _find_log_files

TOTAL_LOGS = 20_000   # volume: makes the directory scan take milliseconds
PRUNED = 200          # how many rotated files rotation removes mid-render
TRIALS = 20


def populate(logs_dir: Path) -> list[Path]:
    (logs_dir / 'robot.log').write_text('x')
    rotated = []
    for i in range(1, TOTAL_LOGS):
        path = logs_dir / f'robot.log.{i}'
        path.write_text('x')
        rotated.append(path)
    return rotated


async def rotate(victims: list[Path]) -> None:
    """Stand in for logrotate pruning old files while the page is being rendered."""
    await asyncio.sleep(0.005)
    for path in victims:
        path.unlink(missing_ok=True)


async def trial() -> str:
    with tempfile.TemporaryDirectory(prefix='mre474-') as tmp:
        logs_dir = Path(tmp)
        rotated = populate(logs_dir)
        victims = rotated[:PRUNED]
        rotator = asyncio.create_task(rotate(victims))
        try:
            await _find_log_files(logs_dir)
            return 'clean'
        except FileNotFoundError as e:
            return f'FileNotFoundError on {Path(e.filename).name}'
        finally:
            await rotator


async def main() -> None:
    core.loop = asyncio.get_running_loop()
    results = []
    for index in range(TRIALS):
        outcome = await trial()
        results.append(outcome)
        print(f'  trial {index + 1:2d}: {outcome}')
    raised = sum(1 for r in results if r != 'clean')
    print(f'\n{raised}/{TRIALS} renders raised FileNotFoundError')

    # positive control: the harness can see a raise on this path
    with tempfile.TemporaryDirectory(prefix='mre474-ctl-') as tmp:
        logs_dir = Path(tmp)
        (logs_dir / 'gone.log').write_text('x')
        real_glob = Path.glob

        def glob_then_delete(self, pattern):
            found = list(real_glob(self, pattern))
            for path in found:
                path.unlink(missing_ok=True)
            return iter(found)
        Path.glob = glob_then_delete  # type: ignore[method-assign]
        try:
            await _find_log_files(logs_dir)
            print('positive control: NO raise  <-- harness is blind, ignore results above')
            sys.exit(2)
        except FileNotFoundError:
            print('positive control: raised as expected (harness can see it)')
        finally:
            Path.glob = real_glob  # type: ignore[method-assign]

    run.tear_down()
    sys.exit(1 if raised else 0)


asyncio.run(main())
```

</details>

Moving `resolve()`/`stat()` into `glob_logs()` closes it — the follow-up your body already names.

**Note** — `rosys/run.py:42`: the retained `is_stopping()` in `cpu_bound` is load-bearing but uncommented, two functions below where visually identical lines were just deleted. `nicegui.run.cpu_bound` raises `RuntimeError: Process pool not set up.` *before* reaching `_run`, so `_run`'s guard never fires and the message does not match the filter. Not a present bug, and your body explains it — but the code does not, and this diff is what makes those lines look like leftovers.

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Review%2C%20verification%20and%20MRE%20from%20the%20model%2C%20run%20on%20both%20trees%3B%20verdict%20and%20severity%20by%20Evan.&a=claude-code "claude-opus-5: Review, verification and MRE from the model, run on both trees; verdict and severity by Evan.")

pascalzauberzeug:
Thanks for the repro, it is what made the fix cheap to verify.

Moving `resolve()`/`stat()` into the worker would have narrowed the gap rather than closed it: the rotation runs on the logging thread, which in rosys is normally the event loop, and that thread can still prune a file while the worker is halfway through its `stat()` calls. So a4ec224e scans the directory entirely in the worker *and* drops a file that is gone by the time it is stat'ed, the tolerance `_file_info` already has five lines below. Your script on that commit: 0/20 raised, and each render returns 19800 paths, i.e. the 200 pruned files are skipped rather than crashing the page.

The kept `is_stopping()` in `cpu_bound` now carries a one-line reason.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Fix%2C%20verification%20against%20the%20reviewer%27s%20repro%20and%20this%20reply%20by%20the%20model%3B%20the%20go%20for%20the%20thread-race%20extension%20and%20for%20posting%20by%20the%20author.&a=claude-code "claude-fable-5-1: Fix, verification against the reviewer's repro and this reply by the model; the go for the thread-race extension and for posting by the author.")

evnchn:
**Verified on `a4ec224e` — closed, not narrowed.** With rotation on its own thread deleting throughout the worker-side scan, so pruning lands inside the `stat()` loop, 5/5 runs came back clean at 19800 paths each.

One heads-up: my repro's own positive control now exits 2 with `harness is blind` — the fix absorbs exactly the case it used. Patching `Path.stat` to raise `PermissionError` instead makes it discriminate again.

falkoschindler (approved):
Approving. The migration is sound, and Evan's verification pass covers the parts that were expensive to re-check, so I will not re-derive them here. Three follow-ups are already on the branch; CI is green on `60c6275c` across 3.11, 3.12 and 3.13.

**`708218ac` — dropped the `lxml` pin.** Evan noted that NiceGUI dropping `lxml` and `lxml-html-clean` was inert because rosys pins `lxml` directly. That is what made the pin worth a second look: it only ever existed to lift NiceGUI's transitive dependency past Dependabot alert 115, and nothing in rosys imports `lxml`. After the bump the pin was the sole reason it was installed at all, so every install, robot images included, was building a C extension for no importer. Re-locking removes it from `uv.lock` entirely.

**`1d298ade` — dropped the `cpu_bound` pre-check.** This reverses the call we all made, so here is the evidence rather than the conclusion. `nicegui.run.cpu_bound` raises `Process pool not set up.` only when `run.process_pool is None`. After `set_up()` the single thing that sets it back to `None` is `run.reset()`, a testing hook; `tear_down()` shuts the pool down but leaves the reference in place. A call during shutdown therefore reaches `_run`, hits the `core.app.is_stopping` guard and returns `None`, exactly like `io_bound`, and after `tear_down()` the executor's `cannot schedule new futures after shutdown` is swallowed by the same filter. The pre-check was as redundant as the guards this PR already removed. Without it the `cpu()` bookkeeping records an id around an await that resolves immediately, which the only reader of that list, `rosys/testing/helpers.py`, does not notice.

`io_bound` is a pure pass-through now, and it keeps a `NOTE` on why it stays a wrapper rather than becoming an alias: rosys documents `R | None` as its own contract in AGENTS.md across 24 call sites, NiceGUI 4.0 will raise `CancelledError` instead of returning `None`, and the wrapper is the one place where rosys gets to decide what to do about that. As an alias it would inherit the new behaviour silently, with neither mypy nor a test noticing.

**`60c6275c` — one `stat()` per log file, in the worker.** `a4ec224e` moved the glob and the sort off the loop, but `_file_info` still stat'ed every file twice more while the list rendered, so most of the per-file I/O stayed on the loop. The scan now returns path, mtime and size from a single `stat()` per file and `_file_info` only formats. That also closes the window between scan and render in which a rotated file rendered as `n/a`, and the scan skips any `OSError` rather than only `FileNotFoundError`, so a log that cannot be stat'ed no longer takes the page down with it.

Evan, one consequence for your repro: with `OSError` absorbed, a `PermissionError` positive control gets swallowed along with everything else. It needs to raise something outside `OSError` to stay discriminating.

Two things deliberately left as they are. `get_all_videos` would 500 on an `.mp4` whose stem does not parse as a date, but only `timelapse_recorder.py` writes into that directory and its names always start with `%Y%m%d_`, so reaching it takes copying files into `~/.rosys` by hand. And the log scan has no test: it is a sort by mtime plus an `except OSError: continue` on a debug page, and pinning the interesting half would mean patching `Path.stat` globally, which is heavier than the code it guards. [※](https://zauberzeug.github.io/colophon/#m=claude&t=Findings%20from%20an%20automated%20review%20pass%2C%20verified%20against%20the%20NiceGUI%203.16%20sources%20by%20the%20model%2C%20which%20also%20wrote%20the%20three%20follow-up%20commits%20and%20this%20review%3B%20severity%20calls%2C%20what%20to%20fix%20and%20what%20to%20leave%2C%20and%20the%20go%20to%20approve%20by%20Falko.&a=claude-code "claude: Findings from an automated review pass, verified against the NiceGUI 3.16 sources by the model, which also wrote the three follow-up commits and this review; severity calls, what to fix and what to leave, and the go to approve by Falko.")

PR #478, opened 2026-09-09
Let intrinsics choose how far to undistort

pascalzauberzeug:
### Motivation

`cv2.undistortPoints` without termination criteria runs exactly five fixed-point iterations and never checks the residual.
That is fine for a mild radial model and badly wrong for a rational one: on a 2560 x 1920 camera with a 14-coefficient rational model the border of the image ends up tens of pixels off.
`Calibration` does this in both of the places where a distorted pixel becomes something else — `_points_to_rays`, which feeds `project_from_image`, the ground projection and any triangulation, and `undistort_points`, through which `SpatialResection.pnp_with_lsa` pushes every observation before it solves for a camera pose.
So the error lands both in what we project and in what we fit: 98 of 400 pixels of a full-image grid fail to reach the ground plane at all, and a noise-free synthetic resection comes out 4.41 mm off.
A downstream application currently subclasses `Calibration` just to override the first of the two; with this change the override becomes unnecessary.

Every camera converges now. Every rational-model calibration out there was fitted and checked against the five-iteration numbers, so this moves projections by pixels to hundreds of pixels at the image border — towards the pixel they came from. `Intrinsics.undistortion_iterations` is there for whoever needs the old numbers back bit-for-bit.

### Implementation

- Add `Intrinsics.undistortion_iterations` as the cap of the fixed-point iteration; the default of 200 converges, 5 reproduces `cv2.undistortPoints` to 5e-14. `scale` and `crop` carry it along, backups without the key load with the default.
- Undistort with `cv2.undistortPointsIter` and that cap in `_points_to_rays` and in `undistort_points`, the latter keeping the undistorted camera matrix it already passes as `P`. Fisheye keeps OpenCV's own criteria, which already converge; `cv2.omnidir.undistortPoints` has no such parameter.
- Build `K`, `D`, `R` and `xi` as `float64` instead of mixing a `float32` camera matrix with `float64` distortion coefficients; with the iteration converged, the `float32` rounding of `K` was the largest remaining residual.
- Add `test_rational_projection_round_trip_across_the_whole_image`, driving a grid over the full image of a rational-model camera, corners included, and pinning the image -> ground -> image round trip.
- Add `test_spatial_resection_with_a_rational_distortion_model`, resecting a camera pose from a noise-free planar target seen through the same model.
- Build the omnidirectional rotation in `undistort_points` from its matrix; `np.array(Rotation)` raised a `TypeError`, so that branch never worked and nothing exercised it. Add `test_undistort_points_omnidirectional`.
- Tighten `test_distort_points_pinhole` to 1e-3 px and the resection test to 1e-6, and drop the docstring note that warned about the pinhole redistortion residual.

#### The residual

Measured over a 40 x 40 grid on a full 2560 x 1920 image, `K = [[1450, 0, 1290], [0, 1450, 960], [0, 0, 1]]`, against a converged reference (eps 1e-10, at most 200 iterations).
The error is read the way a user sees it: undistort, re-project the resulting ray, compare to the pixel we started from.

| distortion vector | median | 95th percentile | max |
|---|---|---|---|
| `[0.982, 1.568, 0, 0, 0.107, 1.328, 1.939, 0.578, 0 …]` | 0.228 px | 30.37 px | 90.68 px |
| `[-0.134, 0.741, 0, 0, 0.065, 0.222, 0.672, 0.319, 0 …]` | 0.219 px | 28.74 px | 81.11 px |

The centre of the image is fine; the error lives where a ground projection needs the rays most.
Run through `project_from_image` with its default `reprojection_tolerance`, 98 of 400 grid pixels fail the tolerance check and come back as `None`.

#### What it costs a resection

Noise-free planar target on `z = 0`, camera looking straight down at 1 m, 63 correspondences spread out to the image border, image points from the exact forward model, straight into `pnp_with_lsa`.

| | five fixed iterations | converged |
|---|---|---|
| average reprojection error | 0.4093 px | 0.000000 px |
| position error | 4.41 mm | 0.000000 mm |
| rotation error | 0.2240 deg | 0.000000 deg |

On perfect input, so the whole of it was the undistortion residual.

#### Runtime

The extra cost scales with the number of points: 1 point 0.8 µs -> 3.4 µs, 192 points 5 µs -> 65 µs, 10 000 points 0.27 ms -> 3.0 ms.
Real 14-coefficient calibrations from downstream applications land at 0.1 to 0.5 ms per 500 points; the strongest of them need the full 200 iterations only at a handful of border pixels where the model is not invertible anyway.

#### Criteria

`TERM_CRITERIA_EPS | TERM_CRITERIA_MAX_ITER`, eps 1e-10, cap from `undistortion_iterations`.
With the eps in place the cap is only a safety net: a mild model converges after a handful of iterations whichever cap it is given, and only the border pixels of a strong rational model ever reach 200.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude&t=Defect%20and%20its%20diagnosis%20named%20by%20the%20author%3B%20reproduction%2C%20fix%2C%20tests%2C%20self-review%20and%20wording%20from%20the%20model.&a=claude-code "claude: Defect and its diagnosis named by the author; reproduction, fix, tests, self-review and wording from the model.")

### DIFF
--- a/rosys/vision/calibration.py
+++ b/rosys/vision/calibration.py
@@ -16,11 +16,7 @@ from .image import Image, ImageSize
 FloatArray: TypeAlias = NDArray[np.float32] | NDArray[np.float64]
 
 UNDISTORTION_TERMINATION_CRITERIA = (cv2.TERM_CRITERIA_EPS | cv2.TERM_CRITERIA_MAX_ITER, 200, 1e-10)
-"""Termination criteria for the iterative undistortion of image points.
-
-OpenCV's default is five fixed iterations without a convergence check,
-which leaves rational distortion models tens of pixels off at the image border.
-"""
+"""OpenCV's default of five fixed iterations leaves rational distortion models off at the image border."""
 
 
 class CameraModel(StrEnum):
@@ -485,7 +481,6 @@ class Calibration:
     @overload
     def distort_points(self, image_points: list[Point], *, crop: bool = False) -> list[Point]:
         """Distort a list of image points.
-        Note: For pinhole models the redistortion can be off by more than 1px for large distortions.
 
         :param image_points: The list of image points to distort.
         :param crop: Whether cropping is applied to the image during distortion.
@@ -495,7 +490,6 @@ class Calibration:
     @overload
     def distort_points(self, image_points: FloatArray, *, crop: bool = False) -> FloatArray:
         """Apply lens distortion to image points using the camera calibration.
-        Note: For pinhole models the redistortion can be off by more than 1px for large distortions.
 
         :param image_points: (Nx2) The image points to distort.
         :param crop: Whether cropping is applied to the image during distortion.
--- a/rosys/vision/calibration.py
+++ b/rosys/vision/calibration.py
@@ -15,6 +15,13 @@ from .image import Image, ImageSize
 
 FloatArray: TypeAlias = NDArray[np.float32] | NDArray[np.float64]
 
+UNDISTORTION_TERMINATION_CRITERIA = (cv2.TERM_CRITERIA_EPS | cv2.TERM_CRITERIA_MAX_ITER, 200, 1e-10)
+"""Termination criteria for the iterative undistortion of image points.
+
+OpenCV's default is five fixed iterations without a convergence check,
+which leaves rational distortion models tens of pixels off at the image border.
+"""
+
 
 class CameraModel(StrEnum):
     PINHOLE = 'pinhole'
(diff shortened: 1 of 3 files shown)


evnchn:
**Raise the default: seven of this PR's own tests fail at `undistortion_iterations = 5`.**

Matching OpenCV's internal count keeps every existing caller bit-identical, but bit-identical to a wrong output is a weak bar to hold — and the PR's own suite says so. Changing the explicit `= 200` back to `= 5` in the three tests that set it:

```
FAILED tests/test_camera_calibration.py::test_rational_projection_round_trip_across_the_whole_image
FAILED tests/test_camera_calibration.py::test_distort_points_pinhole[distortion0 … distortion4]
FAILED tests/test_spatial_resections.py::test_spatial_resection_with_a_rational_distortion_model
7 failed, 44 passed
```

Five of those seven are `test_distort_points_pinhole`, the mild 5-parameter model — so this is not only about exotic rational calibrations.

Measured on the rational calibration from the new test, 2560×1920, 60×60 grid:

| `undistortion_iterations` | max round-trip error | frame over 1 px | pixels that never reach the ground | cost per 1600 pts |
|---|---|---|---|---|
| 5 | 90.68 px | 36 % | 21 % | 0.075 ms |
| 20 | 1.52 px | 2 % | 0 % | 0.240 ms |
| 200 | 0.0000 px | 0 % | 0 % | 0.455 ms |

The 21 % column is the one that would worry me: at the default, a fifth of the image plane comes back as NaN from `project_from_image` rather than as a wrong number, which is exactly what the new `assert not np.isnan(ground_points).any()` catches.

Where the error sits at the default — max round-trip error per cell, image left to right, top to bottom:

```
  90.7   40.2   13.4    4.1    1.4    0.6    0.4    0.6    1.3    3.7   11.9   36.4   85.2
  49.5   13.9    2.9    0.5    0.1    0.0    0.0    0.0    0.1    0.4    2.5   12.0   44.5
  25.8    5.0    0.6    0.1    0.0    0.0    0.0    0.0    0.0    0.0    0.5    4.1   22.4
  15.9    2.3    0.2    0.0    0.0    0.0    0.0    0.0    0.0    0.0    0.1    1.9   13.5
  13.3    1.8    0.1    0.0    0.0    0.0    0.0    0.0    0.0    0.0    0.1    1.4   11.1
  15.9    2.3    0.2    0.0    0.0    0.0    0.0    0.0    0.0    0.0    0.1    1.9   13.4
  25.8    4.9    0.6    0.1    0.0    0.0    0.0    0.0    0.0    0.0    0.5    4.1   22.3
  49.4   13.8    2.9    0.5    0.1    0.0    0.0    0.0    0.1    0.4    2.4   12.0   44.3
  90.5   40.1   13.3    4.1    1.4    0.6    0.4    0.6    1.2    3.6   11.9   36.3   85.0
```

The middle third is exact; it falls apart towards the left and right edges and worst at the corners.

Since `TERM_CRITERIA_EPS` at 1e-10 makes the count a ceiling and the whole range stays under half a millisecond per 1600 points, 20 already buys the frame back and 200 is exact. Would you take one of those as the default?

rosys/vision/calibration.py
```diff
@@ -43,12 +43,15 @@ class Intrinsics:
     :param xi: The omnidirectional camera parameter xi (only for ``CameraModel.OMNIDIRECTIONAL``).
     :param rotation: An inner rotation matrix, useful for visual-inertial calibration or omnidirectional projection.
     :param size: The size of the image.
+    :param undistortion_iterations: Upper bound for the fixed-point iteration that inverts a pinhole distortion model;
+        OpenCV's 5 leave rational models off by pixels at the image border, 200 converge.
```
NiklasNeugebauer:
```suggestion
```
Unneeded detail in docstring

NiklasNeugebauer:
Interesting. I was surprised to learn that point projection uses an iterative solution.
@evnchn @pascalzauberzeug I would imagine that it strongly depends on the lens and application. In my experience, using a simpler model often yields a better solution for the central 90% of the image anyway, but for most applications which would use <1000 points the time increase is likely easily in the budget. I would support raising the default number as I don't think anyone would complain about more accurate point projection.

NiklasNeugebauer:
Ill do a full review later today

NiklasNeugebauer (approved):

PR #479, opened 2026-09-09
Choose the plane orientation IPPE can solve

pascalzauberzeug:
### Motivation

`SpatialResection.pnp_with_lsa` picks `cv2.SOLVEPNP_IPPE` for coplanar object points. OpenCV centres those points and rotates them onto z = 0 itself, but it only returns a correct pose when that plane's z axis points *away* from the camera. With the axis pointing at the camera the pose comes back hundreds of millimetres wrong in an otherwise perfect scene, and for an exactly fronto-parallel plane OpenCV's `rot2vec` divides by `sin(π)` and returns `rvec = [nan nan nan]`. Both cases report `success=True`. Which way the axis points is whatever the plane fit's SVD produced, so whether a resection was right depended on the target's shape and where it sat, and a downstream application had no way to notice.

### Implementation

- Fit the plane once, run IPPE with its z axis pointing both ways, drop a non-finite pose, and keep the candidate with the lower reprojection error; the LM refinement afterwards is unchanged.
- Replace the nested `is_planar` predicate with `_fit_plane_frame`; the automatic algorithm choice reads "planar" off the same SVD.
- Compute the reprojection error in one place, used for the candidate choice and for the returned result.
- Update the docstring: the extrinsic guess applies to `ITERATIVE` only, the sole algorithm OpenCV uses it for.
- Add `test_spatial_resection_with_planar_points`: square and oblong grids, centred and shifted off the camera axis, flat and tilted, on and off z = 0.
- Remove the comments in `pnp_with_lsa` that restated the code.

Measured on noise-free synthetic projections, so the only error is the algorithm's: default intrinsics at 1920x1080 without distortion, camera looking straight down from 0.6 m above the target. 96 scenes: grids of 6x6, 6x5 and 5x7 points spanning ±0.2 m, tilted 0° and 30°, centred and shifted 0.15 m off the camera axis, at plane offsets of 0, 0.4, 0.8 and 1.6 m, solved with `'IPPE'` and with the automatic choice. A scene counts as wrong when the camera position is more than 1 mm off or the pose is not finite.

| | wrong of 96 |
|---|---|
| before | 64 |
| now | 0 |

Rotating the points into their own plane frame before the IPPE call is not enough on its own. It repairs a centred square grid and changes nothing for an oblong or off-axis one, because the sign of the plane normal is still the SVD's. For a square grid the two in-plane singular values tie to 1e-16, so that sign is not even stable against the rounding an offset carries into the centroid: same physical scene, two object frames, opposite outcomes.

`cv2.solvePnPGeneric` is not a way out either. Both solutions it returns come from the same orientation, so in the wrong one both are wrong: 20 px and 194 px mean reprojection error, where the right orientation reaches 1e-13 px.

Deliberately left alone: `success` is hardcoded to `True` whenever OpenCV returns a pose, which is what makes a bad resection silent rather than loud. That is worth its own change.

</details>

### Progress

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Defect%20analysis%2C%20fix%2C%20test%20and%20this%20description%20by%20agents%20over%20several%20sessions%3B%20direction%2C%20triage%20and%20severity%20by%20the%20author.&p=agent&a=claude-code "claude-fable-5-1: Defect analysis, fix, test and this description by agents over several sessions; direction, triage and severity by the author.")

### DIFF
--- a/rosys/vision/spatial_resection.py
+++ b/rosys/vision/spatial_resection.py
@@ -47,7 +47,6 @@ class SpatialResection:
         if np.isnan(world_points).any() or np.isnan(image_points).any():
             raise ValueError('Points contain NaNs')
 
-        # Prepare calibration and undistort image observations for a unified pinhole model
         calibration = Calibration(intrinsics=self.intrinsics)
         K_undist = calibration.get_undistorted_camera_matrix()
 
@@ -57,9 +56,7 @@ class SpatialResection:
 
         plane_frame = _fit_plane_frame(object_points)
 
-        # Decide on algorithm
         if algorithm is None:
-            # Automatic selection
             num_points = object_points.shape[0]
             if num_points >= 4 and plane_frame is not None:
                 method_flag = cv2.SOLVEPNP_IPPE
@@ -90,7 +87,6 @@ class SpatialResection:
         else:
             raise TypeError('Algorithm must be int, str or None')
 
-        # Run PnP on undistorted points with zero distortion and K_undist
         use_guess = p0 is not None and r0 is not None
         if use_guess:
             assert p0 is not None and r0 is not None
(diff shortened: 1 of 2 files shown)


NiklasNeugebauer (dismissed):
approved. i did some refactoring, moved the new code into its own function and ended up going for a `| None` return instead of mirroring OpenCVs non-pythonic flag return pattern.

pascalzauberzeug:
There was a simple conflict with https://github.com/zauberzeug/rosys/pull/478. I will merge without a new review

PR #480, opened 2026-09-11
Keep the GNSS read buffer bounded while the receiver has no time

pascalzauberzeug:
### Motivation

`GnssHardware` cleared its read buffer only after a valid measurement. A receiver that has not had GNSS time since power-up (a robot standing indoors) sends NMEA sentences without a timestamp, which never match `NMEA_PATTERN`. The buffer then grows for as long as the receiver stays without time, and the pattern runs over all of it every 10 ms. After a few hours the main thread sat at 100 % CPU, the UI took minutes to load and socket.io connections to the detectors timed out.

### Implementation

- Split the incoming data into lines and keep only the unfinished tail, so the buffer never holds more than one partial sentence.
- Emit the newest complete GGA/GST/HRP set of a read, as before.
- Sentences split across two reads are no longer lost (the old cut before each match dropped them).

- Profiling a robot in this state with py-spy put 96.8 % of the main thread's samples on the `finditer` line. The receiver was sending `$GPGGA,,,,,,0,00,,,M,,M,,*66`, `$PSSN,HRP,,,,,,,,,00,0,,*48` and `$GPGST,,,,,,,,*57`.
- A second robot, also indoors with 0 satellites, was not affected. Its receiver had got time earlier and kept stamping its sentences, so they matched and cut the buffer back.
- The tail is only unbounded if the receiver never sends a newline, which an NMEA stream does not do.
- The new tests drive the parser through a pseudo-terminal. The split-read test fails on the old parser. The growth itself is not asserted, because it is only visible in the internal buffer.

</details>

### Progress

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Diagnosed%20on%20the%20robot%20together%20with%20the%20author%3B%20code%2C%20test%20and%20wording%20by%20the%20model.&p=agent "claude-opus-5: Diagnosed on the robot together with the author; code, test and wording by the model.")

### DIFF
--- a/rosys/hardware/gnss/gnss_hardware.py
+++ b/rosys/hardware/gnss/gnss_hardware.py
@@ -14,7 +14,7 @@ from .nmea import Gga, GpsQuality, Gst, Pssn
 class GnssHardware(Gnss):
     """This hardware module connects to a Septentrio SimpleRTK3b (Mosaic-H) GNSS receiver."""
     NMEA_TYPES: ClassVar[set[str]] = {'GPGGA', 'GPGST', 'PSSN,HRP'}
-    NMEA_PATTERN = re.compile(r'\$(?P<type>[A-Z,]+),(?P<timestamp>\d{6}(?:\.\d+)?)[^*]*\*[0-9A-Fa-f]{2}\r\n')
+    NMEA_PATTERN = re.compile(r'\$(?P<type>[A-Z,]+),(?P<timestamp>\d{6}(?:\.\d+)?)[^*]*\*[0-9A-Fa-f]{2}')
 
     def __init__(self, *, antenna_pose: Pose | None, reconnect_interval: float = 3.0, max_measurement_age: float = 0.05) -> None:
         """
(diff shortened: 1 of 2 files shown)


evnchn:
**`test_measurement_after_a_receiver_without_time` is green on `main` too, so it guards nothing.**

A regression test earns its place by failing without the fix. Green on both sides is only meaningful when the test pins behaviour that was already correct — and this one is named for the behaviour the PR changes.

| | `…without_time` | `…split_across_reads` | `buffer` after the test's own 100 rounds |
|---|---|---|---|
| `main` `2e548fcc` | passes | **fails** | 7800 bytes, still growing |
| this branch `1422f9aa` | passes | passes | 0 bytes |

The three sentences in `SENTENCES_WITHOUT_TIME` carry no 6-digit timestamp, so `NMEA_PATTERN` never matches them and on `main` nothing trims `buffer`. The test asserts only `last_measurement is None`, then that a good set still parses — both hold on either side, which is why it cannot see the growth. (Buffer sizes read out of the running `_run` coroutine's frame.)

https://github.com/zauberzeug/rosys/blob/1422f9aae657c747d526568736ae5f5ab2bafbbc/tests/test_gnss_hardware.py#L49-L57

The split-read half is genuinely covered. The growth is not, and it is the one in the title. An assertion that fails on `main` needs the buffer reachable from the test — as an attribute, or however you prefer it.

falkoschindler (approved):
Looks good, and the line-based parsing is a clear simplification over the old cut-before-each-match loop. I traced the newest-set selection within a read, split reads, the tail handling and `read_all` on the pty fixture; all of that holds up.

One thing the rewrite dropped: `latest_messages` was no longer cleared after a set had been emitted, so a later sentence repeating an already emitted timestamp completed the stale set again and `NEW_MEASUREMENT` fired twice for the same fix. In practice the age check would have caught the repeat, but the old invariant is worth keeping. I pushed 2f334209 which restores the `clear()`, spells the completeness check as a subset test, and adds a parametrized test for the set selection (newest set in a read wins, mismatched timestamps do not emit, repeated sentence does not re-emit, consecutive sets emit once each). The re-emit case fails on the previous commit.

Not addressed on purpose: the tail is still unbounded if the port never sends a `\n`, which your details section already names and which no configured Septentrio does. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%2C%20follow-up%20commit%20and%20wording%20from%20the%20model%3B%20which%20findings%20to%20address%20was%20decided%20by%20the%20reviewer.&a=claude-code "claude-fable-5-1: Findings, follow-up commit and wording from the model; which findings to address was decided by the reviewer.")

PR #481, opened 2026-09-16
Read only the two BNO055 blocks the IMU module parses

Johannes-Thiel:
### Motivation

`ImuHardware` declared two things that contradicted each other: it parses `cal_gyr` and the quaternion, but left Lizard's `data_select` at its `0xffff` default, so the firmware read all nine BNO055 blocks on every step. Seven of them reach no consumer — neither in rosys nor in anything built on it, since `ImuMeasurement` only carries rotation and angular velocity.

Those reads are synchronous in Lizard's main loop, so the waste is loop time. Measured on an ESP32 Robot Brain core at 100 kHz I²C, `data_select` switched at runtime, cycle time taken from Lizard's own timestamp in the `core` line, two minutes per window:

| `data_select` | blocks read | n | median | p5 | p95 |
|---|---|---|---|---|---|
| 65535 (default) | 9 | 2003 | 60 ms | 55 | 65 |
| 33 (this PR) | 2 | 2929 | **41 ms** | 36 | 46 |
| 0 (no I²C) | 0 | 3486 | 34 ms | 31 | 39 |

So this buys **19 ms per cycle** on every robot with a core IMU, without giving up a single value anyone uses. The remaining 7 ms and the options to remove them are [zauberzeug/lizard#296](https://github.com/zauberzeug/lizard/issues/296).

### Implementation

- Add `ImuHardware.DATA_SELECT = 0x0021` and set it from the generated startup code, next to the `core_message_fields` it is derived from.

Deliberately not configurable. The mask is a consequence of the field list, not a choice, and a wrong value fails silently: a gated block keeps its last value, the status line keeps its length, and rosys keeps parsing valid numbers while the quaternion is frozen. With the IMU carrying 90 % of the angular-rate estimate in a Kalman-based locator, that reads as a robot that is not turning, with no error anywhere. If a consumer ever needs more blocks, the selection belongs one level up — naming measurements, so that `core_message_fields` and the mask are generated together and cannot diverge.

Setting a property in the startup code follows the existing pattern in `estop.py` and `bumper.py`.

No test: the only assertable thing is the generated string, which pins the literal rather than the invariant and would need editing whenever the field list legitimately changes. The behaviour is verified on hardware instead (numbers above).

One honest limit on that verification: the 41 ms were measured by setting the property at runtime on a running core. The generated startup code carrying the assignment has not been deployed to a robot yet.

### Progress

---
⬛ claude-opus-5 (extended thinking)


Johannes-Thiel:
Closing unmerged — wrong layer. Recorded in zauberzeug/lizard#296 as a rejected approach.

Two reasons, both of which apply to any consumer-side `data_select`, not just to this patch:

`core.output` is assembled in `robot.py` from the `core_message_fields` of *all* modules, so `ImuHardware` only owns part of it. A hardcoded mask here silently freezes any IMU property another module or a project's own startup snippet adds to the output — no error, just a value that stops changing. That is the exact failure mode I argued against when rejecting a `data_select` parameter, and this constant reintroduces it. It does not bite today (nothing in rosys reads more than `cal_gyr` and the quaternion), but it is a trap that arms itself the moment someone outputs `imu.yaw`.

More importantly, the firmware-side fixes are strictly better and need no knowledge of what the consumer parses. A single contiguous read over `0x08`–`0x35` turns 18 transactions into 2 and keeps all nine blocks; moving the read into its own task takes it out of the loop entirely (34 ms, the measured floor). Either one makes the mask pointless: at ~5 ms of read time, reading nine blocks stops mattering.

The measurements that motivated this patch stand and are in zauberzeug/lizard#296.

PR #482, opened 2026-09-22
Reduce CPU cost of RTSP camera capture by about a third

agent-NiklasNeugebauer:
### Motivation

RTSP capture spawns a gstreamer pipeline per camera and reads decoded frames over a pipe.
On a Jetson Orin Nano this cost about **0.78 of a core per 720p30 camera**, which a robot running several cameras pays over and over, on the same event loop as its control logic.

### Implementation

Four changes, each measured independently against this PR's base:

- **Carry I420 over the pipe instead of RGB**, converting with `cv2.cvtColor` in Python.
  Halves pipe traffic and drops gstreamer's scalar `videoconvert`.
  The frames also get *more* faithful: against an ffmpeg reference on SMPTE bars the mean per-channel error drops from 6.03 to 1.00, because `videoconvert` was using a less accurate YUV matrix.
- **Convert single-threaded.** OpenCV spread a single 1280x720 `cvtColor` across all six cores, where the dispatch cost several times the conversion itself. Python-side CPU roughly halved.
- **Size the decoder pipe to the workload** — 512 KiB `StreamReader` limit, 256 KiB kernel pipe, up from 64 KiB defaults.
  Deliberately *smaller* than one 1080p frame: anything larger lets frames pile up behind the pipeline's `leaky=downstream` queue, which hands the robot stale images (see below).
- **Handle gstreamer's four-byte row stride.** Widths that are not a multiple of four (854x480, 1366x768) arrive padded; the common unpadded case keeps a zero-copy path.

| workload | before | after | change |
|---|---|---|---|
| 1280x720 H.264 | 23.0 | 15.2 | **-33%** |
| 1920x1080 H.264 | 43.6 | 28.6 | **-34%** |
| 3 concurrent cameras | 25.2 | 17.2 | **-32%** |
| 1280x720 H.265 | 47.7 | 38.4 | **-19%** |

CPU-seconds per 1000 delivered frames, median of paired rounds on a Jetson Orin Nano (gstreamer subprocess + Python process combined).

#### Two bugs this found along the way

Both were introduced by the optimisation work itself and were invisible to the CPU metric, so they are worth a reviewer's attention:

- **Odd widths crashed.** The first I420 version asserted `len(payload) == w*h*3//2`, which only holds when the width is a multiple of four. Any 854x480 or 1366x768 camera would have raised `AssertionError` on every frame and retried forever. Fixed and covered by unit tests, which were verified to fail against the broken version.
- **Frames went stale.** Larger read buffers benchmarked faster but sit *downstream* of the `leaky=downstream` queue, which can only drop what it still holds. After a 2 s consumer stall the robot received 11 buffered frames in a burst instead of live ones. The current buffer sizes are the freshness/CPU compromise; 4 MiB reintroduces the problem at 720p.
  The buffers are a fixed *byte* size, so smaller frames fit in greater number — and RoSys defaults to `substream=1`, the low-resolution stream, which is exactly the case the first fix did not cover: at 854x480 it still replayed 9 frames. The `StreamReader` limit (not the kernel pipe) turned out to be the dominant store, so it is now 512 KiB: 3 replayed frames at 480p, 2 at 720p, 0 at 1080p, at no measurable CPU cost (+0.7 ± 0.8, within noise). The residual ~100 ms is honest — a fixed byte-size buffer cannot be frame-sized for every resolution, and the limit must be chosen before caps are negotiated.

Also verified: no RSS growth over 1800 frames, no descriptor growth over 12 reconnect cycles, colour channel order correct against a deliberately swapped control.

#### Worth a conscious decision

`enlarge_pipe_buffer` reads `stream._transport` to reach the pipe object — private asyncio API.
It is guarded, degrades silently if the attribute disappears, and was verified not to leak descriptors or change reconnect behaviour.
Still a dependency on an implementation detail, so please accept or reject it explicitly.

#### Hardware decode: correction

An earlier revision of this description claimed NVDEC was impossible here because `/dev/v4l2-nvdec` is character device 1:3 (`/dev/null`). That was wrong.
The node is a sentinel the NVIDIA shim `readlink`s to route into `libtegrav4l2.so`; the engine is reached via `/dev/dri/renderD128`, `/dev/nvmap` and `/dev/host1x-fence`.

The actual blocker was that `libgstnvvideo4linux2.so` probes for an integrated GPU by running `lsmod | grep nvgpu`, and `lsmod` is missing from the image — so it took a discrete-GPU path that prerolls and never delivers a frame.
Installing `kmod` fixes it, cutting decode CPU by roughly 10x.

That is a separate change and is not part of this PR; it is recorded here only to retract the earlier claim.
This PR's wins are re-measured against hardware decode separately.

### Progress

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-4-1&t=Autonomous%20optimisation%20session%20by%20the%20agent%3A%20benchmark%20harness%2C%20measurements%20and%20code%20changes%3B%20the%20human%20set%20the%20goal%20and%20the%20device.&p=agent&a=claude-code "claude-opus-4-1: Autonomous optimisation session by the agent: benchmark harness, measurements and code changes; the human set the goal and the device.")


NiklasNeugebauer:
I will look at this further. Seems a bit convoluted. Maybe we can drop the buffer growing code and slim down the rest a little.

If we can get nvdec based decoding that should be a priority, but 30% would be a nice efficientcy gain.

PR #484, opened 2026-09-22
Show the Lizard loop period in the RobotBrain developer UI

Johannes-Thiel:
### Motivation

An overloaded Lizard main loop has repeatedly been the root cause of puzzling problems, most recently lost UART communication at higher baud rates, and nothing in the UI showed the state of the loop. One of our field robots currently runs its loop at about 60–70 ms instead of the nominal 10 ms without anyone noticing. This makes the loop period visible in the RobotBrain developer UI, so a developer can tell at a glance whether Lizard keeps its cadence. A warning on top of it is a separate step once we know typical values.

### Implementation

- Lizard prints one `core` message per iteration of its main loop, so the spacing of their hardware `millis` timestamps is the loop period. No firmware change is needed.
- `RobotBrain` keeps the core timestamps of the last `LOOP_PERIOD_WINDOW` (10 s) in seconds, like the rest of the module, and exposes `get_mean_loop_period()` and `get_max_loop_period()`. Both return `None` until two core messages arrived and again once core messages cease for `CORE_MESSAGE_TIMEOUT` (1 s), e.g. because the ESP is disabled, restarted or hangs, so the UI does not keep showing a healthy-looking period next to "Ready: False".
- A backwards jump of `millis`, i.e. a microcontroller restart, clears the window. So does a gap in the stream longer than `CORE_MESSAGE_TIMEOUT`: with the serial monitor paused or the host stalled, the old timestamps would otherwise make a loop running at a clean 10 ms report a mean of 731 ms and a maximum of 5 s for the next ten seconds.
- Two labels in `developer_ui` next to "Ready" show both values in milliseconds, refreshed by a `ui.timer` once per second. The statistics are methods rather than bound properties on purpose: the maximum walks over the whole window, and NiceGUI's binding loop would have evaluated a bound property every 0.1 s per open page.
- Mean and maximum are shown separately on purpose: a lost line, whether dropped on the wire or by a stalled host, looks like a doubled period, which stands out in the maximum but barely moves the mean.
- Verified on a robot with the first commit: the UI showed 71 ms / 84 ms, and 403 core lines from the serial log over the same 12 s gave a mean of 72.0 ms and a maximum of 86 ms. The later commits change when the statistics are computed and reset, not how.
- One sequential test in `test_robot_brain.py` drives the existing `CommunicationSimulation`: statistics from a run of core messages, the window forgetting old timestamps, the reset on a microcontroller restart, a fresh window after a gap in the stream, and `None` once the messages cease.

### Progress

---
⬛ claude-fable-5-1


falkoschindler (approved):
Nice, this closes exactly the blind spot from the field robot with the 60–70 ms loop. Two rounds of review turned up a handful of things I fixed directly on the branch rather than bouncing them back, so please have a look at the three follow-up commits (d78515c9, 167b45a4, 8d6c384b), the PR description is updated to match:

- The statistics stayed on screen forever once core messages stopped, showing a healthy-looking period next to "Ready: False" after a disable, restart or hang. They now turn to "-" one second after the last core message.
- A gap in the stream, e.g. the serial monitor paused or a host stall, left the old timestamps in the window, so a loop running at a clean 10 ms reported a mean of 731 ms and a max of 5 s for the next ten seconds. Such a gap now starts a fresh window like a restart does.
- The two values were bound properties, and NiceGUI's binding loop evaluated the O(n) walk over the window every 0.1 s per open page. They are now methods refreshed by a ui.timer once per second.
- Smaller things: timestamps stored in seconds like the rest of the module, the window bounded by a maxlen, and the three tests merged into one sequential test that also covers the gap and the stale case.

Left alone on purpose: restart() has its own "messages ceased" check on the hardware time, which predates this PR; unifying it with the new one would be a separate small change if we want it. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20two%20automated%20code%20reviews%2C%20decisions%20from%20the%20conversation%20with%20the%20reviewer%2C%20wording%20from%20the%20model.&a=claude-code "claude-fable-5-1: Findings from two automated code reviews, decisions from the conversation with the reviewer, wording from the model.")

Johannes-Thiel:
I have looked at your changes and have not found any problems with them. We should be good to do the merge.

PR #485, opened 2026-09-22
Decode RTSP streams on the Jetson hardware decoder

agent-NiklasNeugebauer:
### Motivation

Decoding RTSP streams in software is the largest single CPU cost of camera capture on a Jetson.
The hardware decoder was previously believed unusable on this platform — [#482](https://github.com/zauberzeug/rosys/pull/482) recorded it as blocked below the container, because `/dev/v4l2-nvdec` is character device 1:3, i.e. `/dev/null`.

That conclusion was wrong. The node *is* `/dev/null`, but deliberately so: it is a sentinel the NVIDIA v4l2 shim `readlink`s off `/proc/self/fd/N` to recognise its own device and route the call into `libtegrav4l2.so`, which reaches the engine through `/dev/dri/renderD128`, `/dev/nvmap` and `/dev/host1x-fence`. Opening a `HOST1X_CLASS_NVDEC` channel with `DRM_TEGRA_OPEN_CHANNEL` succeeds, so the engine was reachable the whole time.

The real blocker is in the plugin: `libgstnvvideo4linux2.so` distinguishes an integrated from a discrete GPU by shelling out to `lsmod | grep 'nvgpu'`. `lsmod` is not in the image, the probe fails, the plugin assumes a discrete GPU, and the resulting CUVID path fails with `S_EXT_CTRLS for CUDA_GPU_ID failed`, prerolls, and then delivers zero frames forever.

### Implementation

- Install `kmod` in the `Dockerfile` so the plugin's GPU probe can succeed — this alone is what unblocks hardware decode
- Select `nvv4l2decoder ! nvvidconv` over `avdec_* ! videoconvert` in `RtspDevice._run_session` when the element is present, keeping the I420 handoff the reader already expects
- Add `nvdec_is_available()`, an async, cached probe via `gst-inspect-1.0`, and `disable_nvdec()` to retire the hardware path at runtime
- Bound the first frame of a hardware pipeline by `NVDEC_FIRST_FRAME_TIMEOUT`; on timeout, log, retire the hardware path, terminate the process and let the existing reconnect loop come back with software decoding
- Deliberately use **no** parser between `rtp*depay` and `nvv4l2decoder`: an `h265parse` there negotiates a format the decoder accepts and then silently never emits a frame
- Add tests for decoder selection on both kinds of host and for the fallback, both verified to fail against the pre-change code

#### Results

CPU-seconds per 1000 delivered frames on a Jetson Orin Nano, paired A/B (`.auto/ab.sh`), medians of 6 pairs:

| workload | before | after | change |
|---|---|---|---|
| 1280x720 H.264 vs this PR's base | 29.1 | 12.5 | **-56.9%** (23.1x stderr) |
| 1280x720 H.264 vs #482 alone | 22.4 | 14.1 | **-36.4%** (15.2x stderr) |
| 1280x720 H.265 | 52.4 | 13.4 | **-74%** |

Also verified: colour matches ffmpeg as before (per-channel error 1.00, and 56.6 against a deliberately swapped reference), no RSS growth over 1200 frames, no descriptor growth over 12 reconnect cycles, 3 concurrent cameras share the single engine at 84.7 fps combined, and 1080p and 854x480 (the padded-stride path) both decode correctly.

#### Notes for the reviewer

- **Which of #482's wins survive.** Single-threaded `cvtColor` matters *more* now, not less: removing it costs +52.6% (25.8x stderr), because Python conversion is the remaining bottleneck once decode is cheap. The I420 choice is now load-bearing rather than an optimisation — `nvvidconv` offers no plain RGB output at all. The tuned buffer sizes are still required: restoring the old 8 MiB reader reintroduces the staleness defect under NVDEC too (6 replayed frames after a 2 s stall).
- **NV12 was tried and rejected.** It is NVDEC's native format and avoids a plane shuffle, but OpenCV's NV12 conversion is slower than its I420 one: +4.0%, resolved at 3.3x stderr.
- **Freshness at substream resolutions is marginally worse** than software: 5 replayed frames after a 2 s stall at 854x480 versus 4 before (720p and 1080p are unaffected, at 0). The frames now queue inside the hardware decoder, upstream of anything the reader controls, so shrinking the read buffer no longer helps. `disable-dpb=true` cuts it to 3 but breaks B-frame streams, so it is not used.
- **Hardware access needs no `--privileged`** — `/dev/dri/renderD128`, `/dev/nvmap`, `/dev/host1x-fence` and `/dev/v4l2-nvdec` suffice. `docker-compose.yml` already runs privileged, so nothing there changes.

### Progress

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Root%20cause%2C%20fix%20and%20measurements%20by%20the%20agent%3B%20the%20human%20set%20the%20goal%20and%20the%20device.&p=agent)


NiklasNeugebauer:
No priority right now. Will look at this once I am back in office. Seems promising though.

